PR #34, opened 2022-02-13
fix a bug of using article as table name in the reminders query

sywhb:


packages/api/src/routers/svc/reminders.ts
```diff
@@ -56,7 +57,7 @@ export function remindersServiceRouter() {
       const user = await models.user.get(userId)
       if (!user || !user.email) {
         console.log('user not found', userId)
-        res.status(404).send('Not Found')
+        res.status(404).send('User Not Found')
```
jacksonh, line 60: I guess its worth retrying until the message expires here.

packages/api/src/routers/svc/reminders.ts
```diff
@@ -67,7 +68,7 @@ export function remindersServiceRouter() {
 
       if (!reminders) {
         console.log('reminders not found', userId, scheduleTime)
-        res.status(404).send('Not Found')
+        res.status(404).send('Reminders Not Found')
```
jacksonh, line 71: If there are no reminders, should we just return `200` so the task isn't retried?

jacksonh (approved):

packages/api/src/routers/svc/reminders.ts
sywhb: Good catch!

PR #64, opened 2022-02-15
add pdf attachment to library

sywhb:


packages/api/src/routers/svc/pdf_attachments.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    const newsletterEmail = await getNewsletterEmail(email)
+    if (!newsletterEmail || !newsletterEmail.user) {
+      return res.status(401).send('UNAUTHORIZED')
+    }
+
+    const user = newsletterEmail.user
+
+    analytics.track({
```
jacksonh, line 40: Nice. Thank you.

jacksonh (approved):

PR #71, opened 2022-02-16
iPhone NavBar Visibility

satindar:
Hides/shows nav bar based on the scroll position and velocity.

Mostly works but still needs some polish for smoother animations. Also need to fix the popover behavior for the font stepper.


satindar:
https://user-images.githubusercontent.com/836745/154406884-9fc17fb5-8a0c-40ad-b125-6496bdfc9d57.mp4

jacksonh (approved):

PR #79, opened 2022-02-17
fix: Should not get fake uploading url in demo

sywhb:


packages/inbound-email-handler/src/pdf.ts
```diff
@@ -42,11 +42,11 @@ const getUploadIdAndSignedUrl = async (
     email,
   }
 
-  if (process.env.REST_BACKEND_ENDPOINT === undefined) {
+  if (process.env.INTERNAL_SVC_ENDPOINT === undefined) {
```
sywhb, line 45: @jacksonh Could you also help me to add this value to the secrets?

jacksonh (approved):
Ah my mistake too. I saw this and was thinking that value was node.env which would be prod for both prod and demo.

PR #82, opened 2022-02-17
Add support for Bloomberg newsletter

sywhb:
* filter and handle Bloomberg newsletter
* add parser

### DIFF
--- a/packages/inbound-email-handler/src/newsletter.ts
+++ b/packages/inbound-email-handler/src/newsletter.ts
@@ -7,10 +7,11 @@ const EMAIL_FORWARDING_SENDER_ADDRESSES = [
   'Gmail Team <forwarding-noreply@google.com>',
 ]
 const NEWSLETTER_SENDER_REGEX =
-  '<.+@((axios.com)|(mail.bloombergbusiness.com))>'
-const CONFIRMATION_CODE_PATTERN = '^\\(#\\d+\\)'
-const AXIOS_URL_PATTERN = 'View in browser at <.+>'
-const BLOOMBERG_URL_PATTERN = '<a class="view-in-browser__url".+>'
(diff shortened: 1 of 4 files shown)


packages/inbound-email-handler/src/newsletter.ts
```diff
@@ -6,9 +6,12 @@ const EMAIL_CONFIRMATION_CODE_RECEIVED_TOPIC = 'emailConfirmationCodeReceived'
 const EMAIL_FORWARDING_SENDER_ADDRESSES = [
   'Gmail Team <forwarding-noreply@google.com>',
 ]
-const NEWSLETTER_SENDER_REGEX = '<.+@axios.com>'
-const CONFIRMATION_CODE_PATTERN = '^\\(#\\d+\\)'
-const AXIOS_URL_PATTERN = 'View in browser at <.+>'
+const NEWSLETTER_SENDER_REGEX =
```
jacksonh, line 9: Eventually we will need a more scalable solution here. I wonder if there are some more headers we can look at.

packages/inbound-email-handler/test/newsletter.test.ts
```diff
@@ -28,6 +28,11 @@ describe('Newsletter email test', () => {
 
       expect(isNewsletter('', from)).to.be.true
     })
+
+    it('should return true when email is from bloomberg', () => {
+      const from = 'From: Bloomberg <noreply@mail.bloombergbusiness.com>'
```
jacksonh, line 33: 👍 for tests

packages/inbound-email-handler/src/newsletter.ts
sywhb: Yeah I agree. Unfortunately, Bloomberg and Axios newsletters are unlike Substack's which provides very informative headers. So we can only either check the email sender address or parse the DOM

Issue #41, opened 2022-02-14
Improve typography on reader view


jacksonh:
Based on the updated designs we can improve the typography on the reader view:

<img width="698" alt="Screen Shot 2022-02-13 at 20 33 30" src="https://user-images.githubusercontent.com/75189/153800591-c5e2462a-ce3d-4689-bc69-8ca7fe333b86.png">

PR #87, opened 2022-02-18
Tweak the line height for reader text and titles

jacksonh:
We still aren't using the degular font, so I kept the font sizes
the same as they are now. We will need to update if we change
the font.


jacksonh:
Implements https://github.com/omnivore-app/omnivore/issues/41

PR #90, opened 2022-02-18
dev: Frontend development via GitHub Codespaces

kyleconroy:
I've been an Omnivore user for a bit and was really excited to see the open source announcement a few days ago. I've had some success using Codespaces to standardize development environments and make it easier for first-time contributors to get up and running without having to futz around with a local development environment. 

Since you already use Docker Compose, it should be straight forward.


jacksonh:
Sorry about all the approval blocks here. I'll try to figure out a better system that also prevents us from becoming a bitcoin mine.

kyleconroy:
Ended up not being able to get the web container to boot. Going to have to debug locally. Maybe the networking setup is a bit different on Codespaces?

jacksonh:
Oh just saw this comment. It could be environment variables or some of the mapping we do of environment to host names.

I can try to look into it, never used Codespaces before.

kyleconroy:
The API container isn't coming up cleanly, not sure what the problem is. Will investigate further.

jkeech:
👋 I'm on the Codespaces team in GitHub, and someone pointed me to this PR. I wanted to drop by to offer some help getting this repo working well in Codespaces.

I took a quick look, and the first issue I hit was loading the `web` service on port 3000 using the normal next.js dev local config since the [production docker-compose maps 3000:8080](https://github.com/omnivore-app/omnivore/blob/main/docker-compose.yml#L77). This means that when you run `yarn workspace @omnivore/web dev` to launch the server on port 3000 within the `web` service and you forward 3000 locally, docker ends up routing that to port 8080 in the `web` container, so there is no way to directly hit port 3000 in the `web` container since it's always remapped to 8080 (localhost:3000 on the client side -> codespace docker-compose 3000:8080 -> web container:8080)

One fix for this, while still using the production docker-compose.yml config, is to run the dev server on port 8080 (or really any other port that is not remapped in the docker-compose config). If you run `yarn workspace @omnivore/web dev -p 8080`, you can forward either 8080 or 3000 on the client and both will work since they will both end up hitting 8080 in the container.

image

image

----

An alternative setup which might be interesting is to not use docker-compose directly for the codespace. If you launch this repo in the default Codespaces image by removing the `dockerComposeFile` and `service` properties from `devcontainer.json`, you'll be dropped into a container that has docker-in-docker support. You can then follow the existing directions in the readme and run `docker-compose up` in the codespace terminal to run all of the services as child docker containers inside of the main codespace container.

The nested docker-compose setup does not have the same port forwarding issue mentioned above since the `web` service is already running on port 8080 in the production compose config. Running in the nested docker-compose mode also gives easier access to all of the compose container logs in the codespace terminal.

image

jacksonh:
Thanks @jkeech, regarding 8080 we could use that port directly without remapping now. It was changed to 3000 when we had a mix of services running inside and outside of docker.

kyleconroy:
@jkeech Thanks for the tips! I didn't realize that the default container supported running Docker. I've updated this PR with a very (very!) minimal devcontainer.json.

When running `docker-compose up`, the web container still fails to start. I can successfully get it to run if I follow the frontend development guide:

```
docker-compose up api content-fetch
cd packages/web
cp .env.local .env
yarn install
yarn dev
```

That almost works. The issue is that the email registration talks directly to the API, which is hard-coded to localhost:4000. I can change that in the .env file, but is there anyway to get that domain programatically?

jacksonh:
Does `.env.local` get loaded automatically? We can move that file elsewhere, its really just a helper for people to setup locally.

jacksonh:
I'll try this in the morning, but based on that last screenshot I'm going to guess an issue connecting to postgres

jkeech:
> I can change that in the .env file, but is there anyway to get that domain programatically?

It should be possible to use an `onCreateCommand` in devcontainer.json (or a similar lifecycle hook) to programmatically write the `.env` file. There's a env var injected into each codespace called `$CODESPACE_NAME`, and the githubpreview url format is `https://$CODESPACENAME-$PORT.githubpreview.dev`.

One thing to be aware of when using the githubpreview urls in Codespaces is that the default visibility for forwarded ports is `Private`. This ensures that only you can access your forwarded ports, even though it's a publicly routable DNS name. Those urls are authenticated using a cookie tied to your GitHub account, so you might need to manually navigate to that url in the browser for port 4000 before automated requests would work (otherwise the auth cookie wouldn't be set).

Alternatively, you can make the port visibility `Public` in the ports tab, which drops the auth check and allows anyone to access that url. Or, if you connect from VS Code desktop instead of connecting in the browser, VS Code can bind to local ports on your machine and you'll be able to use `localhost:3000` and `localhost:4000` instead of the githubpreview urls. The reason you can't use localhost urls when connecting from the browser client is that we can't bind to local ports on the client side when running in the browser.

jacksonh:
Will try this one again in the future

Issue #88, opened 2022-02-18
Some images are not being displayed in the Axios newsletter


jacksonh:
Here is an example image proxy URL that fails: https://proxy-demo.omnivore-image-cache.app/100%x0,sCh5EyC35VQVAZdNY4byf0b_pccwDyG5LeHsgI10LSLc/https://datawrapper.dwcdn.net/E8vUS/fallback.png 

this came from my saved copy of the newsletter here: https://demo.omnivore.app/jacksonh/axios-pm-gen-z-surprise-17f0988918c

with an original URL here: https://link.axios.com/view/61eef5e5492d41078a32440cfxaw4.a1v4/df289925

it looks like the issue is the image sizing is set to "100%x0" so it does not get resized properly.

PR #95, opened 2022-02-18
Handle non-number size attributes in images

jacksonh:
The previous code assumed we would always have number size values
for width and height, but attributes like 100% are valid. In
cases where we don't have a numeric value we can just fallback
and let the item be sized by the reader CSS.


jacksonh:
This fixes https://github.com/omnivore-app/omnivore/issues/88

PR #98, opened 2022-02-21
Labels API and testing page

sywhb:


jacksonh:
So in the end we will have these methods, right?

- `labels()` -- return all the labels for the user
- `createLabel(name, color, description (optional))` -- create a new label for the user
- `deleteLabel(name)` 
- `setLabels(linkId, labels[])`

sywhb:
> 

Yes. A tiny difference is I use `id` as the parameter in `deleteLabel(id)`

I was writing some descriptions in the PR when I saw your comment.

packages/api/src/resolvers/function_resolvers.ts
```diff
@@ -423,6 +428,12 @@ export const functionResolvers = {
         ctx.models
       )
     },
+    async labels(article: { linkId: string }): Promise<Label[] | undefined> {
```
jacksonh, line 431: One thing about this is it will do a DB query per an article in the response, right now we limit the total items to 100, maybe we should lower it. iOS will fetch 10 at a time and web fetches 20 at a time. 

Ideally we could try to join the labels table when we load, but I know thats tricky given how these queries are setup. I'm not sure if apollo gives us the information to get all the returned articles at once so we can do something like `get labels where in (list of link ids)`

sywhb: @jacksonh I agree with your point. 
My plan was having a separate API for fetching labels for articles like a micro-service architecture to make it more flexible and reduce loading time by caching or reduce refetching.

However, in here, since we don't need change labels on the home page, maybe a better way is to have a single API to fetch everything by joining the tables

Issue #99, opened 2022-02-21
Resaving the same page does not seem to add it to the library list


jacksonh:
Steps to reproduce:

1. Save a URL from iOS
2. Go into the web app and remove the Link
3. Save the same URL again from iOS 

Note that the link is not in the library.

jacksonh:
Note this only happens when saving from iOS, likely because it is using the `SavePage` API rather than the `SaveLink` API.

PR #100, opened 2022-02-21
Use null instead of undefined when unarchiving links

jacksonh:
This fixes unarchiving content by re-saving it via the Save*
API methods.

Setting to undefined is not the same as null in this case, and
will result in archived_at being non-null and the link then
being excluded from the GetArticles query.


jacksonh:
This fixes https://github.com/omnivore-app/omnivore/issues/99

PR #112, opened 2022-02-23
Update the help text for newsletters

jacksonh:


sywhb (approved):
Nice

jacksonh:
> Nice

Thanks, hopefully a little better. Always a place we can improve a lot.

Issue #94, opened 2022-02-18
sitename is not indexed in search


jacksonh:

jacksonh:
Looking at this for an updating indexing function. Need to test how long this migration will take:


```

CREATE OR REPLACE FUNCTION update_page_tsv() RETURNS trigger AS $$
begin
  new.tsv :=
    setweight(to_tsvector('pg_catalog.english', coalesce(new.title, '')), 'A') ||
    setweight(to_tsvector('pg_catalog.english', coalesce(new.author, '')), 'A') ||
    setweight(to_tsvector('pg_catalog.english', coalesce(new.description,'')), 'A') ||
    setweight(to_tsvector('pg_catalog.english', coalesce(regexp_replace(new.url, '^((http[s]?):\/)?\/?([^:\/\s]+)((\/\w+)*\/)([\w\-\.]+[^#?\s]+)(.*)?(#[\w\-]+)?$', '\3'), '')), 'A') ||
    setweight(to_tsvector('pg_catalog.english', coalesce(regexp_replace(new.url, '^((http[s]?):\/)?\/?(.*\.)?([^:\/\s]+)(\..*)((\/+)*\/)?([\w\-\.]+[^#?\s]+)(.*)?(#[\w\-]+)?$', '\4'), '')), 'A') ||
    setweight(to_tsvector('pg_catalog.english', coalesce(new.content,'')), 'B');
  return new;
end
$$ LANGUAGE plpgsql;

CREATE OR REPLACE TRIGGER article_tsv_update BEFORE INSERT OR UPDATE
    ON omnivore.pages FOR EACH ROW EXECUTE PROCEDURE update_page_tsv();
```

jacksonh:
Deployed on prod now.

PR #126, opened 2022-02-25
Add the link domain to the search index

jacksonh:
This adds the domain of the saved link to the search index in
two formats, the full domain like www.omnivore.app and just
the secondary domain like `omnivore`.

This also adds some ranking to search, making title, description,
and domain more highly ranked than the content but search results
wont change as we still order by saved_at.


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/94

PR #129, opened 2022-02-25
Remove snackbar offset

satindar:


jacksonh:
Interesting, glad you tested on that weirdly curved device.

jacksonh (approved):

satindar:
@jacksonh Set the bottom padding to 8. Looks better on the SE.

PR #138, opened 2022-02-25
Newsletter Email for mac/iOS

satindar:


jacksonh:
Did you move the `Create` button to the top? It's showing up above the list for me.

satindar:
> Did you move the `Create` button to the top? It's showing up above the list for me.

Yeah switched that so you see it appear in the list below. Also if the list of emails is long it's still on screen.

Easy to switch around though.

jacksonh:
oh Yeah, that is smart about the long list.

jacksonh (approved):

Issue #137, opened 2022-02-25
Favicon is not being setting for /settings/* pages


jacksonh:
example: https://omnivore.app/settings/emails

PR #147, opened 2022-02-27
Import highlightjs types

jacksonh:
We still need these dev packages because of the way our docker
builds are setup.


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/137

PR #152, opened 2022-02-28
Add a flag in readability to retain table elements in newsletter emails

sywhb:


packages/api/src/utils/parser.ts
```diff
@@ -166,6 +169,7 @@ const getReadabilityResult = (
       const article = new Readability(document, {
         debug: DEBUG_MODE,
         createImageProxyUrl,
+        keepTables: isNewsletter,
```
jacksonh, line 172: 👍

packages/api/src/utils/parser.ts
```diff
@@ -205,7 +209,8 @@ const applyHandlers = async (url: string, window: DOMWindow): Promise<void> => {
 
 export const parsePreparedContent = async (
   url: string,
-  preparedDocument: PreparedDocumentInput
+  preparedDocument: PreparedDocumentInput,
+  isNewsletter?: boolean
```
jacksonh, line 213: I think in the future we might want to create an `options` var that we pass in, but single param is better for now since we only have one.

jacksonh (approved):

packages/api/src/utils/parser.ts
sywhb: Yes. We can update this if we need more.
Actually, it's this line `this._cleanConditionally(articleContent, "table");` which removes some tables.

It cleans an element of all tags of type "tag" if they look fishy.
"Fishy" is an algorithm based on content length, classnames, link density, number of images & embeds, etc.

So by disabling this function for `tables` in newsletters, some ads will also be kept.

I'm thinking about updating the algorithm to still remove ads but keep tables with declarative text

@jacksonh

sywhb: I spent a few hours on updating the algorithm to remove ads but keep certain Golang Weekly newsletters table elements. Unfortunately, I haven't found a good way to do that.

Tests work fine with Bloomberg and Substack newsletters, except for Axios, I have made some changes to accommodate it. I think we might just go on with this solution first

jacksonh:
Deployed to prod, i guess we can test in four days 😆

PR #167, opened 2022-03-01
SwiftUI State Initialization

satindar:
Refactor to make sure we never initialize state in a view's initializer. 

https://swiftcraft.io/blog/how-to-initialize-state-inside-the-views-init-


jacksonh (approved):

jacksonh:
Optimistically approved this one but going to go back and read that article so I understand whats going on better.

Issue #130, opened 2022-02-25
Font size changes on web are not persisted


jacksonh:
Steps to reproduce:

1. Navigate to the reader
2. Adjust the font size
3. Navigate back to the library
4. Go back to the reader -- note that the font size has not been changed

These should at least be saved in the personalization cookie so they are maintained per-device.

PR #170, opened 2022-03-01
Ensure font size changes are persisted

jacksonh:


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/130

PR #178, opened 2022-03-03
New function to determine if an HTML blob is probably a newsletter based on its content

jacksonh:


jacksonh:
Will deploy this to demo for testing

PR #185, opened 2022-03-04
iPad Refresh

satindar:
The grid view doesn't support `refreshable` so added a refresh button on the toolbar.

Also added searchable to macOS (12.0 and above only).


jacksonh:
Huh, i noticed iPad Apple News had a refresh button too (in a popover)

jacksonh (approved):

satindar:
> Huh, i noticed iPad Apple News had a refresh button too (in a popover)

Sort of have this style working but has some visual bugs. Will mess with it some more and try to get it right.

PR #188, opened 2022-03-06
Fix readerview scrollbars in iOS dark mode

jacksonh:
- Set root color scheme CSS for iOS to determine system colour
- After load set webview to opaque


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/184

jacksonh:
@satindar wanted to flag you on this one. It seems like its probably safe to keep setting those properties every time `didLoad` is called, I assume webview does nothing.

satindar (approved):
Looks good! Nice find on the bug. Thanks for fixing 👍🏽

Issue #179, opened 2022-03-03
validateUrl is not used correctly in pageSaveRequest


jacksonh:
The service has its own URL validation, but we should use the utility method.

https://github.com/omnivore-app/omnivore/blob/f7e3a1657eb9c7af1a67bf25faf8265f75dc3d4a/packages/api/src/services/create_page_save_request.ts#L48

PR #200, opened 2022-03-07
Use the validateUrl method to validate URLs

jacksonh:


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/179

PR #203, opened 2022-03-08
Reader delete and archive actions [iOS]

satindar:
Adds a toolbar button that opens a menu where the user can delete the item or archive/unarchive it.


jacksonh:
Nice, i think you did the hard part here. I am wondering if we should do a half modal with these actions as buttons. But thats probably an easy change?

jacksonh (approved):

satindar:
> Nice, i think you did the hard part here. I am wondering if we should do a half modal with these actions as buttons. But thats probably an easy change? 

Yeah an action sheet should be pretty easy. I'll merge this one, wrap up the progress updater branch, them create another for an action sheet.

PR #237, opened 2022-03-15
Apple async image errors

satindar:
If the card image fails to download then the allocated space is set to zero.


jacksonh (approved):
nice one.

PR #243, opened 2022-03-16
Update to Swift 5.6

satindar:
This will require upgrading to Xcode 13.3


jacksonh:
Does this require upgrading iOS supported version?

jacksonh (approved):

satindar:
> Does this require upgrading iOS supported version?

No this will still support iOS14. Just have to update to Xcode 13.3 for compiling.

PR #246, opened 2022-03-16
Use a single process pool for all webviews

jacksonh:
This is a first step to improving our caching on iOS.

### DIFF
--- a/apple/OmnivoreKit/Sources/Views/Article/WebAppView.swift
+++ b/apple/OmnivoreKit/Sources/Views/Article/WebAppView.swift
@@ -5,6 +5,27 @@ import WebKit
 
 public let readerViewNavBarHeight = 50.0
 
+enum WebViewConfigurationManager {
+  private static let processPool = WKProcessPool()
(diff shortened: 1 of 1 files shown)


apple/OmnivoreKit/Sources/Views/Article/WebAppView.swift
```diff
@@ -28,7 +49,7 @@ public let readerViewNavBarHeight = 50.0
     }
 
     func makeUIView(context: Context) -> WKWebView {
-      let webView = WebView(frame: CGRect.zero)
+      let webView = WebViewManager.create()
```
satindar, line 52: @jacksonh Should this be calling `.shared` instead?

jacksonh: @satindar I left shared in there because I'm using it in testing, but I wanted to have it create a new one every time for now, just to be the same as our previous behavior. 

Once we start using shared we need to remove the old content on dismiss or it will show briefly.

Issue #176, opened 2022-03-02
Titles and descriptions are HTML encoded


jacksonh:
![image](https://user-images.githubusercontent.com/75189/156394534-43384994-d2e1-4821-9791-88dcd0c94d1a.jpeg)

Original URL https://www.theatlantic.com/ideas/archive/2022/03/putins-war-dispelled-the-worlds-illusions/623335/

jacksonh:
Fixed this for descriptions but need to also for titles

jacksonh:
PR that fixes descriptions: https://github.com/omnivore-app/omnivore/pull/230

PR #251, opened 2022-03-16
Use html decoding when getting values from fetched oembed

jacksonh:
If we fetch oembed data from an external source, instead of
handling it in readabilityjs we need to html decode it.


jacksonh:
Closes out https://github.com/omnivore-app/omnivore/issues/176

PR #252, opened 2022-03-16
Drop ios14 support

satindar:
make way for swift async 🙌🏽


jacksonh:
Hmmmmmm, think we should do a few smaller releases first? Not sure what the stats are at now.

satindar:
> Hmmmmmm, think we should do a few smaller releases first? Not sure what the stats are at now.

@jacksonh Mixpanel is showing iOS14 around 16% of devices. Declining about 0.5% a week. Checked Intercom and we have 13 iOS14 users in the last 30 days. 3 with more than 10 sessions. One of them is in Seattle :)

satindar:
this one is stale so going to reopen a new one

PR #253, opened 2022-03-17
make readingProgress required in the elastic page data

sywhb:


jacksonh:
Great, with this can we delete from the `function_resolvers`?

sywhb:
> Great, with this can we delete from the `function_resolvers`?

@jacksonh Yes, we can delete them from `function_resolvers` since they are stored in elastic

sywhb:
@jacksonh Oh I will run an update query in elastic to update those null readingProgress to 0 first.

So we won't have trouble with the new pages added yesterday

jacksonh (approved):

sywhb:
@jacksonh I have added `readingProgressPercent=0` and `readingProgressAnchorIndex=0` to the pages in elastic where `readingProgressPercent` and `readingProgressAnchorIndex` don't exist.

PR #254, opened 2022-03-17
default sort by savedAt to avoid recent read page to appear on top of…

sywhb:
… the list


jacksonh (approved):
I wonder why so much code was regenerated.

sywhb:
> I wonder why so much code was regenerated.

Because I did a save on the `graphql` ts file and my IDE automatically performs a lint on it.
I will regenerate it to avoid these irrelevant changes @jacksonh

sywhb:
Now it is clearer @jacksonh

PR #258, opened 2022-03-17
fix reading progress not updated after reading by updating cache

sywhb:


sywhb:
@jacksonh

packages/web/lib/networking/queries/useGetLibraryItemsQuery.tsx
```diff
@@ -259,6 +259,8 @@ export function useGetLibraryItemsQuery({
           readingProgressAnchorIndex: 0,
         })
         break
+      case 'refresh':
```
jacksonh, line 262: woah, this is a nice trick. Should be really handy.

jacksonh (approved):

PR #263, opened 2022-03-17
Remove extra theme objects in stitches config

satindar:
Since we dropped the lighter and darker themes we no longer need the stitches config for them.

### DIFF
--- a/packages/web/components/tokens/stitches.config.ts
+++ b/packages/web/components/tokens/stitches.config.ts
@@ -166,10 +166,10 @@ export const { styled, css, theme, getCssText, globalCss, keyframes, config } =
       md: '(min-width: 768px)',
       lg: '(min-width: 992px)',
       xl: '(min-width: 1200px)',
-    }
+    },
   })
 
-export const darkTheme = createTheme(ThemeId.Dark, {
+const darkThemeSpec = {
   colors: {
     ...yellowDark, // Brand
     ...orangeDark, //Accent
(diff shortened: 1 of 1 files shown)


jacksonh:
Ah thanks so much for this one. Makes it way easier to read

jacksonh (approved):

packages/web/components/tokens/stitches.config.ts
```diff
@@ -203,90 +203,21 @@ export const darkTheme = createTheme(ThemeId.Dark, {
...
-    highlightBackground: 'rgb(255,234,159)',
-    highlight: '#FFD234',
-    error: '#FA5E4A',
+}
 
-    // Brand Colors
-    loginBg: '#FFD492',
+// Dark and Darker theme now match each other.
+// Use the darkThemeSpec object to make updates.
+export const darkTheme = createTheme(ThemeId.Dark, darkThemeSpec)
```
satindar, line 210: We might be able to remove these as well. As long as nobody has preferences for lighter and darker in the backend (or we convert on the client...which we might do. will check)

satindar: (meant to refer to darker and lighter themes)

jacksonh: Yeah hopefully we convert, I'd be worried about removing because it could be some of the oldest users that would have them.

jacksonh:
Going to deploy this one to demo to check

packages/web/components/tokens/stitches.config.ts
satindar: Probably better to leave these in for now. I can't tell if we're converting it. Could do a follow up PR to clean it up but at least this one gets rid of the extra noise and makes it easier to see where updates are needed when we add to the theme objects.

satindar:
> Going to deploy this one to demo to check

@jacksonh did this look ok on demo?

jacksonh:
> > Going to deploy this one to demo to check
> 
> 
> 
> @jacksonh did this look ok on demo?

I think so, but mind deploying it again? I just deployed a bunch of other stuff to demo.

PR #264, opened 2022-03-18
Fix/elastic update conflict

sywhb:


jacksonh:
This looks good I'll merge it in the morning. I want to be careful about front end changes we merge right now while production backend is frozen.

PR #266, opened 2022-03-18
Fix/city journal parsing

sywhb:


packages/readabilityjs/Readability.js
```diff
@@ -168,8 +168,8 @@ Readability.prototype = {
     // NOTE: These two regular expressions are duplicated in
     // Readability-readerable.js. Please keep both copies in sync.
     articleNegativeLookBehindCandidates: /breadcrumbs|breadcrumb|utils/i,
-    articleNegativeLookAheadCandidates: /outstream(.?)_|sub(.?)_/i,
-    unlikelyCandidates: /-ad-|ai2html|banner|breadcrumbs|breadcrumb|combx|comment|community|cover-wrap|disqus|extra|footer|gdpr|header|legends|menu|related|remark|replies|rss|shoutbox|sidebar|skyscraper|social|sponsor|supplemental|ad-break|agegate|pagination|pager(?!ow)|popup|yom-remote|copyright|keywords|outline|infinite-list|beta|recirculation|site-index|hide-for-print|post-end-share-cta|post-end-cta-full|post-footer|main-navigation|programtic-ads|outstream_article|hfeed|comment-holder/i,
+    articleNegativeLookAheadCandidates: /outstream(.?)_|sub(.?)_|m_/i,
```
jacksonh, line 171: Just wondering how this one shows up?

jacksonh:
Can we also add this page as a test case? To do that in readabilityjs we run a command like this:

`yarn generate-testcase city-journal "https://www.city-journal.org/will-western-sanctions-accelerate-russian-realignment?wallit_nosession=1"`

then look at the expected page to ensure it looks the way we want and check everything in:

`open test/test-pages/city-journal/expected.html`

packages/readabilityjs/Readability.js
sywhb: The top and right arrow are images with class name `m_article_blahblahblah go_to_top/right` and it matches the `articleNegativeLookAheadCandidates` regex so it would not be removed as a `okMaybeItsACandidate` element.

That's why I added `m_` in the `articleNegativeLookAheadCandidates` and `back-to-top|show-up-next` to the `unlikelyCandidates` to remove these arrow images
@jacksonh

jacksonh (approved):

PR #268, opened 2022-03-18
Remove next dependency from <ArticleContainer component

jacksonh:


jacksonh:
I think I'd rather pass a callback down to the router so we can use `router?.push('/home'),` which will be faster than `window.location.href`

jacksonh:
I'm going to close this one out and try to incorporate in another patch

PR #285, opened 2022-03-22
Fix postgres error when merging highlights

sywhb:
fix article_id type mismatch (uuid => varchar) when saving highlight in postgres by setting `article_id` to be `undefined`
@jacksonh

### DIFF
--- a/packages/api/test/resolvers/highlight.test.ts
+++ b/packages/api/test/resolvers/highlight.test.ts
@@ -52,6 +52,43 @@ const createHighlightQuery = (
   `
 }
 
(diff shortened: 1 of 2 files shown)


packages/api/test/resolvers/highlight.test.ts
```diff
@@ -52,6 +52,43 @@ const createHighlightQuery = (
   `
 }
 
+const mergeHighlightQuery = (
```
jacksonh, line 55: Nice thanks for adding a test

jacksonh (approved):

PR #289, opened 2022-03-22
Fix uploading endpoint returns 504

sywhb:
send response to client to close connection


jacksonh:
I suspect this wont work. I can't find it in docs, but I think App Engine has a 30s timeout on requests.

jacksonh (approved):

PR #293, opened 2022-03-22
Use lower priority queue for large imports

sywhb:
* add a count query in elastic
  * if priority is not set, check the rate limit of user's libary
    - if 5 articles added in the last minute: use low queue
    - default: use normal queue


jacksonh:
This looks right to me. We probably can wait until we have everything else deployed and tested. Just less moving parts.

PR #295, opened 2022-03-22
Web Reader Highlights [Apple]

satindar:


packages/appreader/additional.d.ts
```diff
@@ -1,9 +0,0 @@
-export {}
```
jacksonh, line 1: Don't we need this?

jacksonh: Oh i see, you converted the file to jsx

jacksonh (approved):

PR #296, opened 2022-03-22
Remove unneeded props for sharing from the article container

jacksonh:


jacksonh:
Will kill this off and work on the event propagation stuff separately.

PR #297, opened 2022-03-22
Separate the mutations out of lower level components

jacksonh:
This will let us handle the mutations differently on native
iOS and should help in avoiding making web calls directly from
the web view, so we can avoid CORs and introduce a caching
layer.


satindar (approved):

jacksonh:
@satindar ok, tested out highlights and reading progress on demo with this deployed. Going to merge now.

PR #312, opened 2022-03-24
Pass html content and title to appreader as HTML

jacksonh:
Passing the content in as HTML means we dont have to worry about
escaping it or any issues with quotes in the htmlContent


jacksonh:
@satindar I think this is a lot safer way for us to pass data around. Not sure if its worth deleting the elements after we get the content from them.

satindar (approved):

PR #315, opened 2022-03-24
Webreader Highlight.js and Mathjax script

satindar:
Still need to load some Mathjax fonts but it's at least rendering the eqations.


apple/OmnivoreKit/Sources/App/Views/WebReader/WebReaderContent.swift
```diff
@@ -45,10 +42,10 @@ struct WebReaderContent {
         </div>
         <script type="text/javascript">
           window.omnivoreEnv = {
-            "NEXT_PUBLIC_APP_ENV": "\(appEnv.rawValue)",
-            "NEXT_PUBLIC_\(appEnv.rawValue.uppercased())_BASE_URL": "\(appEnv.webAppBaseURL.absoluteString)",
-            "NEXT_PUBLIC_\(appEnv.rawValue.uppercased())_SERVER_BASE_URL": "\(appEnv.serverBaseURL.absoluteString)",
-            "NEXT_PUBLIC_\(appEnv.rawValue.uppercased())_HIGHLIGHTS_BASE_URL": "\(appEnv.highlightsServerBaseURL.absoluteString)"
+            "NEXT_PUBLIC_APP_ENV": "prod",
```
satindar, line 45: Not used but if this isn't set to a valid value then the web assertions break.

jacksonh: Got it, maybe we can slim things down a bit more in a future bundle. Not a big deal right now though.

jacksonh (approved):

PR #318, opened 2022-03-25
add success and failure count in assertions

sywhb:


jacksonh (approved):

jacksonh:
Nice, was going to say this would help

PR #325, opened 2022-03-26
Pass createdAt, savedAt, and publishedAt into the appreader

jacksonh:


jacksonh:
This should resolve https://github.com/omnivore-app/omnivore/issues/324

PR #332, opened 2022-03-28
Reload the web reader if its process is terminated

jacksonh:


jacksonh:
I don't think this will actually do anything because we don't load the view in `updateUIView` we probably need to move the `loadHtmlString` in there instead of when the view is created.

satindar (approved):

PR #339, opened 2022-03-29
Feature/search highlights

sywhb:


packages/api/.env.test
```diff
@@ -28,4 +28,6 @@ SEGMENT_WRITE_KEY='test'
 REMINDER_TASK_HANDLER_URL=http://localhost:4000/svc/reminders/trigger
 PUBSUB_VERIFICATION_TOKEN='123456'
 PUPPETEER_TASK_HANDLER_URL=http://localhost:9090/
-
+ELASTIC_URL=http://localhost:9200
+BOOKMARKLET_VERSION="latest"
```
jacksonh, line 32: I think we can remove the bookmarklet environment variables, they shouldn't be used anywhere anymore.

packages/api/src/generated/schema.graphql
```diff
@@ -803,6 +803,7 @@ type Query {
   me: User
   newsletterEmails: NewsletterEmailsResult!
   reminder(linkId: ID!): ReminderResult!
+  search(after: String, first: Int, query: String): SearchResult!
```
jacksonh, line 806: @satindar this new method will replace `GetArticles` for us. Returns a kind of slimmed down object, allows it to be either a deep link to a highlight or a page.

jacksonh (commented):
We will have to figure out how to get this all merged. Maybe we can cherry-pick all the backend API changes and get those, and then separately try to pull the frontend changes into the newest UI (all pretty close to old).

sywhb:
close this pr because we have split it to two pr for backend and web changes:
https://github.com/omnivore-app/omnivore/pull/395
https://github.com/omnivore-app/omnivore/pull/396

PR #342, opened 2022-03-29
Ensure title doesnt get html encoded with new appreader

jacksonh:


jacksonh:
@satindar this fixes this encoding issue in titles, issue was it would get double encoded by saving in the HTML: https://discord.com/channels/844965259462311966/942463305107259432/958380750204461147

satindar (approved):

PR #346, opened 2022-03-30
Improvements to the edit labels modal

jacksonh:


jacksonh:
This is all covered in the new article actions

PR #366, opened 2022-04-01
Label Creation Form

satindar:
Adds UI and network calls to create labels in iOS.

UI is still pretty awful but this is behind a feature flag and will allow us to develop the rest of the labels feature. Follow up PR's will probably be:

- Ability to assign label to link
- Display label on grid/list cards
- Add labels to search options
- Come back and clean up UI once everything is functional
- Drag and Drop implementation

Set `enableLabels` to true in the `FeatureFlag` file to test this out.

### DIFF
--- a/apple/OmnivoreKit/Sources/App/Views/ApplyLabelsView.swift
+++ b/apple/OmnivoreKit/Sources/App/Views/ApplyLabelsView.swift
@@ -2,6 +2,7 @@ import Combine
 import Models
 import Services
 import SwiftUI
+import Views
 
 final class ApplyLabelsViewModel: ObservableObject {
   private var hasLoadedInitialLabels = false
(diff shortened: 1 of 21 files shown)


jacksonh:
I think we want to split this into two parts. We can list the items like you have now on the first view, but then add a `Create New Label` button that opens a modal with the stuff you have on top. 

Does that make sense?

jacksonh:
Or we could have something like a button that toggles the create UI. So instead of what you have now, at the top there is a `Create Label` button, and when you tap that it displays the UI you have now with save/cancel options.

satindar:
> Or we could have something like a button that toggles the create UI. So instead of what you have now, at the top there is a `Create Label` button, and when you tap that it displays the UI you have now with save/cancel options.

Modal sounds like a good idea. We can probably match the emails UI and show the modal when the + button is tapped.

jacksonh (approved):

PR #373, opened 2022-04-04
Reading Progress Update Fix [Apple]

satindar:
Instead of mutating the feed items while the user is in reader mode we store the update in the view model and make the update when the user navigates to the grid view.


jacksonh:
:nice:

jacksonh (approved):

PR #399, opened 2022-04-11
Label filtering ios

satindar:
Adds UI to filter on labels in the library items view.

### DIFF
--- a/apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewIOS.swift
+++ b/apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewIOS.swift
@@ -195,6 +195,10 @@ import Views
               viewModel: viewModel
             )
             .contextMenu {
+              Button(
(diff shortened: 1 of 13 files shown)


apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewIOS.swift
```diff
@@ -198,7 +226,13 @@ import Views
             }
             if #available(iOS 15.0, *) {
               link
-                .swipeActions(edge: .trailing, allowsFullSwipe: true) {
+                .swipeActions(edge: .trailing, allowsFullSwipe: false) {
```
jacksonh, line 229: I think we should remove this one. Just the long press should be enough, and with this user's can't do the full edge to edge swipe to archive, which is probably their most common action.

jacksonh (approved):
Yeah I think just the swipe is the only thing. Looks really clean overall.

PR #417, opened 2022-04-13
Upload highlights and labels data to GCS

sywhb:


packages/api/src/datalayer/pubsub.ts
```diff
@@ -37,17 +36,35 @@ export const createPubSubClient = (): PubsubClient => {
         Buffer.from(JSON.stringify({ userId, email, name, username }))
       )
     },
-    pageUpdated: (page: Partial<Page>, userId: string): Promise<void> => {
+    entityCreated: <T>(
```
jacksonh, line 39: nice. I think this will make pubsub events easier too. We can have just a single event that pushes highlights or pages.

jacksonh (approved):

PR #428, opened 2022-04-14
Drop iOS14 support

satindar:


jacksonh:
🔥 -- I'll tag now before we merge.

jacksonh (approved):

PR #429, opened 2022-04-14
Convert ViewerPublisher to async fetchViewer

satindar:


apple/OmnivoreKit/Sources/App/AppExtensions/Share/ShareExtensionScene.swift
```diff
@@ -80,13 +80,14 @@ final class ShareExtensionViewModel: ObservableObject {
       .store(in: &subscriptions)
 
     // Using viewerPublisher to get fast feedback for auth/network errors
-    services.dataService.viewerPublisher()
```
jacksonh, line 83: Definitely more readable.

jacksonh (approved):

PR #432, opened 2022-04-18
Feature/subscription list resolver

sywhb:


packages/api/src/entity/subscription.ts
```diff
@@ -0,0 +1,48 @@
...
+  unsubscribeMailTo?: string
+
+  @Column('text', { nullable: true })
+  unsubscribeHttpUrl?: string
+
+  @CreateDateColumn()
+  createdAt!: Date
+
+  @UpdateDateColumn()
+  updatedAt!: Date
```
jacksonh, line 47: Will `updatedAt` when a new page is received for a subscription? I think we will want people to be able to order their subscription by last issue received. Also probably want to expose it in the UI so they know when they last received an article from a subscription.

jacksonh (approved):

packages/api/src/entity/subscription.ts
sywhb: Yes we can do that

PR #453, opened 2022-04-20
Add unsubscribe resolver

sywhb:


jacksonh:
Hey @sywhb, is there a way to know if a library item is part of a subscription now? Like in the library will we be able to tell its a subscription so we can add the `Unsubscribe` button?

sywhb:
> Hey @sywhb, is there a way to know if a library item is part of a subscription now? Like in the library will we be able to tell its a subscription so we can add the `Unsubscribe` button?

Yes, we have a `subscription` field in the item. I will add it to the getArticle API response

jacksonh (approved):

packages/api/src/services/subscriptions.ts
```diff
@@ -25,9 +27,57 @@ export const saveSubscription = async (
...
+  } else {
+    throw new Error('No unsubscribe method defined')
+  }
+
+  // set status to unsubscribed
+  subscription.status = SubscriptionStatus.Unsubscribed
+  return getRepository(Subscription).save(subscription)
+}
+
+const sendUnsubscribeEmail = async (
```
jacksonh, line 59: nice that we can handle both methods

PR #460, opened 2022-04-21
Add basic subscriptions UI

sywhb:


sywhb:
Adding generic table component

jacksonh:
With the web code we don't pull in the graphql generated code, so I added the type definition manually.

PR #396, opened 2022-04-08
Feature/search highlights web

sywhb:

PR #467, opened 2022-04-22
Omn 407: [Link to highlight on page]

gitstart-omnivore:
## Description
- Updated the initial function that scroll to reading Position to take into account when the url contains an HREF for highlight.
- Also added for the clicking on highlight to scroll to the highlight position

## House Keeping
- [ ] Added video describing your plan of action ?
- [x] Added Loom video ?
- [x] Is Deployment Link passing ?
- [x] Have you assinged PR to yourself ?
- [x] Is Github action passing ?
- [x] Updated the appropriate tag ?

## Ticket link (if applicable)
[Task Link](https://app.gitstart.com/clients/omnivore/tickets/OMN-407)
[Soapbox Link](https://soapbox.wistia.com/videos/XCflbbMcs8)

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Others

## Remarks

- [ ] I added a new package to achieve this task.
- [ ] I have updated package.json

### DIFF
--- a/packages/web/components/templates/article/ArticleContainer.tsx
+++ b/packages/web/components/templates/article/ArticleContainer.tsx
@@ -6,7 +6,7 @@ import { ArticleSubtitle } from './../../patterns/ArticleSubtitle'
 import { theme, ThemeId } from './../../tokens/stitches.config'
 import { HighlightsLayer } from '../../templates/article/HighlightsLayer'
 import { Button } from '../../elements/Button'
-import { MutableRefObject, useEffect, useState } from 'react'
+import { MutableRefObject, useEffect, useState, useRef } from 'react'
 import { ReportIssuesModal } from './ReportIssuesModal'
 import { reportIssueMutation } from '../../../lib/networking/mutations/reportIssueMutation'
 import { ArticleHeaderToolbar } from './ArticleHeaderToolbar'
@@ -15,6 +15,7 @@ import { updateThemeLocally } from '../../../lib/themeUpdater'
 import { ArticleMutations } from '../../../lib/articleActions'
 import { LabelChip } from '../../elements/LabelChip'
 import { Label } from '../../../lib/networking/fragments/labelFragment'
+import { HighlightLocation, makeHighlightStartEndOffset } from '../../../lib/highlights/highlightGenerator'
 
 type ArticleContainerProps = {
   article: ArticleAttributes
@@ -36,6 +37,11 @@ export function ArticleContainer(props: ArticleContainerProps): JSX.Element {
   const [showShareModal, setShowShareModal] = useState(false)
   const [showReportIssuesModal, setShowReportIssuesModal] = useState(false)
   const [fontSize, setFontSize] = useState(props.fontSize ?? 20)
+  const highlightHref = useRef(window.location.hash ? window.location.hash.split('#')[1] : null)
+  const [highlightReady, setHighlightReady] = useState(false)
+  const [highlightLocations, setHighlightLocations] = useState<
+    HighlightLocation[]
+  >([])
 
   const updateFontSize = async (newFontSize: number) => {
     if (fontSize !== newFontSize) {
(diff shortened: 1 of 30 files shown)


jacksonh:
It looks like with this one we aren't scrolling to the read position in the cases where there isn't a highlight id. I'll try to debug.

packages/web/components/templates/article/ArticleContainer.tsx
```diff
@@ -36,6 +37,11 @@ export function ArticleContainer(props: ArticleContainerProps): JSX.Element {
   const [showShareModal, setShowShareModal] = useState(false)
   const [showReportIssuesModal, setShowReportIssuesModal] = useState(false)
   const [fontSize, setFontSize] = useState(props.fontSize ?? 20)
+  const highlightHref = useRef(window.location.hash ? window.location.hash.split('#')[1] : null)
```
jacksonh, line 40: How come this uses `useRef`? Is it so we don't access window while server side rendering?

packages/web/components/templates/article/ArticleContainer.tsx
```diff
@@ -48,6 +54,21 @@ export function ArticleContainer(props: ArticleContainerProps): JSX.Element {
     updateFontSize(props.fontSize ?? 20)
   }, [props.fontSize])
 
+  // Load the highlights
```
jacksonh, line 57: Did we need to move this out of the highlight layer?

Temitope-Emmanuel: Hello @jacksonh, yeah, the highlights is being added asynchronously to the page and we need to know that it is ready before running the function to scroll to either the last read position or highlight with an id. We could use prop drilling to also achieve this. Another way we can go about this is if we take note of the position of the highlight parent and immediately scroll rather than using the highlight id

packages/web/components/templates/article/ArticleContainer.tsx
Temitope-Emmanuel: We are not expecting to perform any changes after the initial Href of the url on load. That's why I am using useRef, I was not thinking about serverside since the function to scroll to the position runs in a useEffect

gitstart-omnivore: [@jacksonh](https://github.com/jacksonh) Yeah, the highlights are being added asynchronously to the page and we need to know that they are ready before running the function to scroll to either the last read position or highlight with an id. We could use prop drilling to also achieve this. Another way we can go about this is if we take note of the position of the highlight parent and immediately scroll rather than using the highlight id

packages/web/components/templates/article/ArticleContainer.tsx
gitstart-omnivore: [@jacksonh](https://github.com/jacksonh) Yeah, the highlights are being added asynchronously to the page and we need to know that they are ready before running the function to scroll to either the last read position or highlight with an id. We could use prop drilling to also achieve this. Another way we can go about this is if we take note of the position of the highlight parent and immediately scroll rather than using the highlight id

packages/web/components/templates/article/ArticleContainer.tsx
gitstart-omnivore: We are not expecting to perform any changes after the initial Href of the URL on load. That's why we are using useRef, we weren't not thinking about server side since the function to scroll to the position runs in a useEffect.

gitstart-omnivore:
@jacksonh We have fixed the initial scroll to read position issue when there is no highlight id.

packages/web/components/templates/article/ArticleContainer.tsx
jacksonh: Thanks. Personally I'd prefer prop pushing, but I'll test out this version as-is.

jacksonh:
Seems like this is working with read position now. Going to deploy to demo and test a bit more.

jacksonh:
I have folded these changes into https://github.com/omnivore-app/omnivore/pull/396 so I will close this one out.

PR #471, opened 2022-04-23
Add timezone difference in today search query

sywhb:


sywhb:
> Looks good. Lets test it out on demo since there are a bunch of formatting changes.

Deployed this in Demo and looks good

PR #489, opened 2022-04-26
use react-live-chat-loader to defer loading of intercom

jacksonh:


jacksonh:
Will try this again in the future

PR #499, opened 2022-04-27
Keep iOS NavBar hidden

satindar:
Moved the modifier onto NavLink to make sure it's always applied. It can get lost of we apply it further down the tree (according to other comments I found online). Seems to fix it. Haven't reproduced the bug again.

Also fixed the mac app and converted some combine funds to async.


jacksonh:
:nice:

jacksonh (approved):

PR #501, opened 2022-04-28
Fix subscriptions issues

sywhb:


sywhb:
@jacksonh 
This PR fixes two issues:
1. Unsubscribe email failed to send
2. Unsubscribe all the subscriptions once the newsletter email is deleted

packages/api/src/services/subscriptions.ts
```diff
@@ -56,28 +81,22 @@ export const unsubscribe = async (
...
 
-const sendUnsubscribeHttpRequest = async (url: string): Promise<void> => {
-  const response = await fetch(url, {
-    method: 'GET',
-  })
-
-  if (!response.ok) {
-    throw new Error(`Failed to unsubscribe, response: ${response.statusText}`)
+    for (const subscription of subscriptions) {
+      await unsubscribe(subscription)
```
jacksonh, line 97: One small issue I see, is if one fails to unsubscribe and throws an error, then we end this loop and don't attempt the others.

jacksonh (approved):

packages/api/src/services/subscriptions.ts
sywhb: Nice to catch the bug

Issue #508, opened 2022-04-28
Web is sending the segment.identify event too frequently


jacksonh:
In the PrimaryHeader we call the identify function on each re-render:

```
/ Attempt to identify the user if they are logged in.
  useEffect(() => {
    const user = window.analytics?.user().id()
    if (!user && viewerData?.me?.id) {
      window.analytics?.identify({ userId: viewerData?.me?.id })
    }
  }, [viewerData?.me?.id])
  ```
  
  this should only be needed if the user has not been identified yet.

PR #516, opened 2022-04-29
Move the analytics event for link_read out of the container, supply userId

jacksonh:
This prevents events from being sent when there is no user id yet
due to race conditions in analytics setup.


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/508

PR #540, opened 2022-05-04
Show unable to parse if page is failed to save within 30 seconds

sywhb:


packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
```diff
@@ -259,7 +259,11 @@ export function HomeFeedContainer(props: HomeFeedContainerProps): JSX.Element {
         const username = viewerData?.me?.profile.username
         if (username) {
           setActiveCardId(item.node.id)
-          if (item.node.state === State.PROCESSING) {
+          if (
```
jacksonh, line 262: Do we still need this bit with your API change?

sywhb: Actually we didn't make changes to the API. The content `<p>We were unable to parse this page.</p>` is added in the last commit.
We were sending user to `/links/id` page when this item is still processing (for refreshing) but here we need to send user to `/slug` page in order to see the unparsable content.
We can also send the user to `slug` page here and do the refreshing

packages/api/src/resolvers/article/index.ts
```diff
@@ -102,7 +102,7 @@ const FORCE_PUPPETEER_URLS = [
   /twitter\.com\/(?:#!\/)?(\w+)\/status(?:es)?\/(\d+)(?:\/.*)?/,
   /^((?:https?:)?\/\/)?((?:www|m)\.)?((?:youtube\.com|youtu.be))(\/(?:[\w-]+\?v=|embed\/|v\/)?)([\w-]+)(\S+)?$/,
 ]
-const UNPARSEABLE_CONTENT = 'We were unable to parse this page.'
+const UNPARSEABLE_CONTENT = '<p>We were unable to parse this page.</p>'
```
sywhb, line 105: I only added a <p> tag here

packages/api/src/resolvers/article/index.ts
```diff
@@ -434,7 +434,7 @@ export const getArticleResolver: ResolverFn<
 
     if (
       page.state === ArticleSavingRequestStatus.Processing &&
-      new Date(page.createdAt).getTime() < new Date().getTime() - 1000 * 60
+      new Date(page.createdAt).getTime() < new Date().getTime() - 1000 * 30
```
sywhb, line 437: Only reduced the duration here

packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
jacksonh: Ahh ok

packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
```diff
@@ -259,11 +258,7 @@ export function HomeFeedContainer(props: HomeFeedContainerProps): JSX.Element {
         const username = viewerData?.me?.profile.username
         if (username) {
           setActiveCardId(item.node.id)
-          if (item.node.state === State.PROCESSING) {
-            router.push(`/${username}/links/${item.node.id}`)
-          } else {
-            router.push(`/${username}/${item.node.slug}`)
-          }
+          router.push(`/${username}/${item.node.slug}`)
```
jacksonh, line 261: I am not sure if this works because slug can change after the item is saved.

sywhb: Changed to `id`. I will also add a refresh interval in slug page

packages/api/src/resolvers/article/index.ts
jacksonh: I think this should be `page.savedAt` because that is set each time the URL is saved.

jacksonh (approved):

PR #541, opened 2022-05-04
Display labels on the appreader page

jacksonh:


apple/OmnivoreKit/Sources/Models/DataModels/FeedItem.swift
```diff
@@ -58,6 +58,19 @@ public extension LinkedItem {
     return URL(string: pageURLString ?? "")
   }
 
+  var labelsJSONString: String {
```
jacksonh, line 61: It would be nicer to do this in the `LinkedItemLabel` extension but thats in the models.

satindar (approved):

PR #543, opened 2022-05-04
Tune the update speed

sywhb:


jacksonh:
Oh, what does this do?

sywhb:
> Oh, what does this do?

There is an exception thrown while running the update query in elastic

jacksonh (approved):

Issue #544, opened 2022-05-04
YouTube sharing is broken


jacksonh:
It looks like YouTube share URLs have been updated to the following format, causing issues for us parsing video IDs.

https://youtube.com/watch?feature=share&v=RVmm1w5iE88

PR #545, opened 2022-05-04
Fix parsing of video IDs out of shared YouTube URLs

jacksonh:


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/544

PR #552, opened 2022-05-05
Fix batch_size=500 to avoid exceeding max memory limit and requests_p…

sywhb:
…er_second=250 to throttle update requests

Fix batch_size=500 to avoid exceeding max memory limit and requests_per_second=250 to throttle update requests


sywhb:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html

jacksonh (approved):

PR #561, opened 2022-05-06
Faster save by reusing JSDOM

sywhb:


packages/api/src/utils/parser.ts
```diff
@@ -253,9 +254,8 @@ export const parsePreparedContent = async (
     // Format code blocks
     // TODO: we probably want to move this type of thing
     // to the handlers, and have some concept of postHandle
-    if (article?.content) {
-      const cWindow = new JSDOM(article?.content).window
-      cWindow.document.querySelectorAll('code').forEach((e) => {
+    if (article?.dom) {
+      article.dom.querySelectorAll('code').forEach((e) => {
```
jacksonh, line 258: nice. I didn't know that was available.

jacksonh (approved):

PR #566, opened 2022-05-09
Optimize saving performance

sywhb:
* Cache chrome data in a local folder
* Add stealth and adblock plugin to puppeteer to reduce processing time
* Remove Request Interception to avoid conflicts with adblock
* Reduce http call to get jsonld data if title or content or sitename or byline exists

### DIFF
--- a/packages/api/src/utils/parser.ts
+++ b/packages/api/src/utils/parser.ts
@@ -255,22 +255,25 @@ export const parsePreparedContent = async (
     // TODO: we probably want to move this type of thing
     // to the handlers, and have some concept of postHandle
     if (article?.dom) {
-      article.dom.querySelectorAll('code').forEach((e) => {
-        console.log(e.textContent)
-        if (e.textContent) {
-          const att = hljs.highlightAuto(e.textContent)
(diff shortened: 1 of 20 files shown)


jacksonh (approved):

packages/api/src/utils/parser.ts
```diff
@@ -255,39 +255,44 @@ export const parsePreparedContent = async (
...
+    const jsonLdLinkMetadata = (async () => {
+      return getJSONLdLinkMetadata(window.document)
+    })()
 
     Object.assign(article, {
       content: clean,
-      title: article?.title || jsonLdLinkMetadata.title,
-      previewImage: article?.previewImage || jsonLdLinkMetadata.previewImage,
-      siteName: article?.siteName || jsonLdLinkMetadata.siteName,
+      title: article?.title || (await jsonLdLinkMetadata).title,
```
jacksonh, line 290: Does this get called twice if title and byline are missing?

sywhb: It should be only called once

PR #590, opened 2022-05-11
Remove www. from the siteName

sywhb:


sywhb:
Hey @jacksonh , I have removed `www.` from the search api response

packages/api/src/resolvers/article/index.ts
```diff
@@ -877,7 +878,7 @@ export const searchResolver = authorized<
         isArchived: !!r.archivedAt,
         contentReader:
           r.pageType === PageType.File ? ContentReader.Pdf : ContentReader.Web,
-        originalArticleUrl: r.url,
+        originalArticleUrl: cleanDomainInUrl(r.url),
```
jacksonh, line 881: I think we need to keep `originalArticleUrl` the same, since that is used to load the URL if a user wants to go to the original.

Maybe this change needs to be done in the clients. I have a feeling they aren't using `siteName`.

sywhb: Yes. The frontend uses `originalArticleUrl` to get the site name but i forgot about the `originalArticleUrl` is being used in other places too.

jacksonh: Yeah we have siteName in the API, frontends should really be using that instead of parsing itself.

sywhb: Okay. I will replace it

sywhb: @jacksonh This is ready to 🚢

jacksonh: I think we just need to remove this line here, as "Open Original URL" still needs this value

sywhb: Reverted

jacksonh (approved):

PR #597, opened 2022-05-12
Optimize puppeteer and remove chrome-aws-lambda dependencies

sywhb:
And block requests to 'font', 'image', 'stylesheet', 'script', 'media' in puppeteer


jacksonh:
One thing I don't understand is if we need stylesheet at all. Like does readability look up the stylesheets to check things like visibility?

sywhb:
> One thing I don't understand is if we need stylesheet at all. Like does readability look up the stylesheets to check things like visibility?

I think it only looks up inline style but I will double check it

sywhb:
@jacksonh Readability does not use stylesheet but puppeteer looks up stylesheet to remove hidden elements so we probably should not block stylesheet

```const style = window.getComputedStyle(el);```

**Comment by sywhb** [comment]
Also removed blocking of `media` requests as some images won't be shown

**Review by jacksonh — APPROVED** [review]

**Comment by jacksonh** [comment]
I think an issue running in the cloud function here 

```
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
Error
2022-05-13T23:17:17.412Zpuppeteer Error: Could not find expected browser (chrome) locally. Run `npm install` to download the correct Chromium revision (982053). at ChromeLauncher.launch (/workspace/node_modules/puppeteer-core/lib/cjs/puppeteer/node/Launcher.js:105:23) at PuppeteerNode.launch (/workspace/node_modules/puppeteer-core/lib/cjs/puppeteer/node/Puppeteer.js:125:31) at PuppeteerExtra.launch (/workspace/node_modules/puppeteer-extra/dist/index.cjs.js:129:41)
Default
2022-05-13T23:17:17.412Zpuppeteer Unhandled rejection
```

sywhb:
> I think an issue running in the cloud function here
> 
> ```
> This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
> Error
> 2022-05-13T23:17:17.412Zpuppeteer Error: Could not find expected browser (chrome) locally. Run `npm install` to download the correct Chromium revision (982053). at ChromeLauncher.launch (/workspace/node_modules/puppeteer-core/lib/cjs/puppeteer/node/Launcher.js:105:23) at PuppeteerNode.launch (/workspace/node_modules/puppeteer-core/lib/cjs/puppeteer/node/Puppeteer.js:125:31) at PuppeteerExtra.launch (/workspace/node_modules/puppeteer-extra/dist/index.cjs.js:129:41)
> Default
> 2022-05-13T23:17:17.412Zpuppeteer Unhandled rejection
> ```

Yeah. We still need `chrome-aws-lambda` in cloud function

PR #608, opened 2022-05-12
Remove listnav elements from articles

jacksonh:


packages/readabilityjs/test/generate-testcase.js
```diff
@@ -173,7 +173,7 @@ async function fetchSource(url, callbackFn) {
 }
 
 function sanitizeSource(html, callbackFn) {
-  htmltidy(parseHTML(html).serialize(), {
+  htmltidy(html, {
```
jacksonh, line 176: @sywhb wasn't sure if this is right. It seems like .serialize isn't available here, but i also wasn't sure if parseHTML really needed to be called.

sywhb: You are right! `html` is the HTML string but not a DOM object. @jacksonh

sywhb (approved):
LGTM

PR #613, opened 2022-05-13
Fix UpdateLabelResult undefined

sywhb:


jacksonh (approved):
We should probably add a test case for this one.

PR #620, opened 2022-05-13
Use chrome-aws-lambda in the puppeteer GCF

jacksonh:


sywhb:
Thanks @jacksonh for adding this

jacksonh:
No problem. Kind of annoying that you can't tell which packages work and do not work with GCF

PR #622, opened 2022-05-14
Remove single-process from chromium args

sywhb:


sywhb:
Remove single process[https://www.chromium.org/developers/design-documents/process-models/#single-process] args

jacksonh (approved):

PR #627, opened 2022-05-16
Webview share option

satindar:


jacksonh:
Do we have room to make it "Share Original" ?

satindar:
> Do we have room to make it "Share Original" ?

yeah plenty of room for that. i'll update.

jacksonh (approved):

PR #643, opened 2022-05-17
Simplify the label filter modal, add negation

jacksonh:


jacksonh:
Going to clean this up a little more

satindar (approved):

PR #650, opened 2022-05-18
Fix parsing github blog

sywhb:


jacksonh (commented):
So the blocked resources and the fetching of iframe content were the issue?

packages/puppeteer-parse/index.js
```diff
@@ -790,7 +790,7 @@ async function retrieveHtml(page) {
           }
         }
       });
-      return document.documentElement.innerHTML;
+      return document.documentElement.outerHTML;
```
sywhb, line 793: @jacksonh The change fixed the issue.
With `outerHTML`, we get a <html> element which can be parsed correctly.

PR #651, opened 2022-05-18
Store sqlite file in app group file container [iOS]

satindar:
This allows us to access the core data container from the share extension. 

If we want to support batch operations then there's more work involved with setting up persistent history tracking and notifications. I think can avoid that for now and revisit when we want to optimize.


jacksonh (approved):
Will this wipe out the old file?

satindar:
> Will this wipe out the old file?

@jacksonh good question...my guess it just leaves but i'll see if we can try to find it and remove it

jacksonh:
Ah not too worried about removing the old, was more asking if it means a new is created. So any old data is "gone"

satindar:
> Ah not too worried about removing the old, was more asking if it means a new is created. So any old data is "gone"

Yeah pretty sure the old data is "gone" (but still somewhere on disk)

PR #654, opened 2022-05-18
fix/blog back button

jacksonh:
- Remove redundant back buttons
- Add test case for a webflow based blog


sywhb (commented):
I guess we have to regenerate those failed testcases

jacksonh:
Will close this one out.

PR #655, opened 2022-05-19
Fix not updating labels in elastic after updating in postgres

sywhb:


jacksonh:
Was reported this morning too: https://github.com/omnivore-app/omnivore/issues/656

jacksonh (approved):

PR #661, opened 2022-05-19
iOS background fetch

satindar:
A few things/questions left:
- Adjust timing of BG task scheduling (right now it's set to wait at least an hour)
- Max items should we fetch? (right now it's 30)
- Do we need to fetch IDs 10 at a time as it is now? Or just ask for them all at once?
- Additional event tracking
- Performance optimizations (run requests in parallel. skip content fetches when user opens the app)


jacksonh (approved):

satindar:
Gonna merge then come back to the questions/opts

PR #662, opened 2022-05-20
Send background notifications to iOS on save to trigger a sync

jacksonh:


jacksonh:
Closing this as sync is fast now

PR #663, opened 2022-05-20
Fix cannot convert null to object error message

sywhb:
https://console.cloud.google.com/errors/detail/CMOnj7i0xP_xQA?authuser=0&project=omnivore-production


jacksonh (approved):
Any idea why this is null? This will produce a blank page for the user I think, so it would be good to alert somehow.

sywhb:
> 

Yeah. It happens when readability didn't find any content to present. Possibly because all the elements have very low score or there is some error in parsing the page.

I will take a deeper look at this

PR #666, opened 2022-05-20
Reader mode refresh

satindar:
Now that it's CoreData driven we can update the library list while in reader mode (since core data uses the objectID as identity rather than the entire struct. So properties can change without causing identity to change).


jacksonh:
:NICE:

jacksonh (approved):

PR #685, opened 2022-05-21
Simplify the API to create pages from PDFs

jacksonh:
This creates a stub page when an upload is initiated, so the
SaveFile (or create article from file) APIs do not need to be
called.

The PDF uploaded trigger fired by GCS will mark the upload
as completed and update the page status.


sywhb (approved):

sywhb:
Nice!

jacksonh:
Thanks for taking a look at this one @sywhb. New elastic stuff makes this much simpler

jacksonh:
> Nice!

one less API call 🚀

PR #688, opened 2022-05-23
Fix/pdf loading

satindar:
- Import PSPDFKit in the App package
- Use a PDFItem struct as data model rather than passing in the mutable LinkedItem (avoids document being recreated)
- Wrap Document in a StateObject so it persists between re-renders


satindar:
Needs more testing but I think this is the right direction.

jacksonh (approved):

PR #689, opened 2022-05-23
Save and filter by language

sywhb:
* parse language code from document or from open graph locale metadata
* convert language code to language display name and store in elastic when creating article
* support searching items by language: `language: english`

### DIFF
--- a/packages/readabilityjs/Readability.js
+++ b/packages/readabilityjs/Readability.js
@@ -1894,13 +1894,13 @@ Readability.prototype = {
 
     // get site name
     metadata.siteName = jsonld.siteName ||
-      values["og:site_name"];
+      values["og:site_name"] || null;
 
     // get website icon
     const iconLink = this._doc.querySelector(
       "link[rel='apple-touch-icon'], link[rel='shortcut icon'], link[rel='icon']"
     );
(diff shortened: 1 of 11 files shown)


packages/readabilityjs/Readability.js
```diff
@@ -2833,6 +2835,11 @@ Readability.prototype = {
     return false;
   },
 
+  _getLanguage: function(code) {
+    let lang = new Intl.DisplayNames(['en'], {type: 'language'});
```
jacksonh, line 2839: nice, is `en` the fallback value here?

jacksonh:
Note sure if the new version is more correct, but looks like a bunch of failures related to `siteName`

sywhb:
> Note sure if the new version is more correct, but looks like a bunch of failures related to `siteName`

The default value of `siteName` should be null

packages/readabilityjs/Readability.js
sywhb: `en` here is the locale of the display name so the display names would be in English

sywhb: I have added a fallback value = 'English'

PR #698, opened 2022-05-24
Support sort or filter by publish time

sywhb:


sywhb:
> nice. Do we need to run a migration for this?

We need to update the mappings of index which is done in the `db:migrate`

jacksonh (approved):

PR #705, opened 2022-05-25
Lowercase characters when doing text analysis

sywhb:
Hey @jacksonh, to make this change, I need to reindex in elasticsearch because this changes the analyzer and we cannot do that by updating mappings.


jacksonh:
@sywhb makes sense. Can we trigger that from cloud build like we did last time? Is there anything special we need to do?

Also, did you see my other issue with `title`? Not sure if that would also require a re-index.

jacksonh:
oh, i just realized this isn't just for highlights.

jacksonh (approved):

PR #706, opened 2022-05-25
Extract contentType from safari

jacksonh:
This fixes issues where we'd incorrectly identify a PDF as HTML.
Safari will use our preprocess javascript file even if the
file is PDF, so we rely on JS to detect the content type here.

### DIFF
--- a/apple/OmnivoreKit/Sources/Models/PageScrapePayload.swift
+++ b/apple/OmnivoreKit/Sources/Models/PageScrapePayload.swift
@@ -19,11 +19,18 @@ public struct PageScrapePayload {
   public let url: String
   public let contentType: ContentType
 
-  init(url: String, title: String?, html: String?) {
+  init(url: String, title: String?, html: String?, contentType: String?) {
     self.url = url
     self.title = title
     self.html = html
-    self.contentType = url.hasSuffix(".pdf") ? .pdf : .html
+
+    // If the content type was specified and we know its PDF, use that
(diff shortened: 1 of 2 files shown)


jacksonh:
I looked into the UTI stuff on iOS to see if there is a more accurate way to map contentType to PDF, but it seems like that was just a simple string match too.

apple/OmnivoreKit/Sources/Models/PageScrapePayload.swift
```diff
@@ -19,11 +19,18 @@ public struct PageScrapePayload {
...
-  init(url: String, title: String?, html: String?) {
+  init(url: String, title: String?, html: String?, contentType: String?) {
     self.url = url
     self.title = title
     self.html = html
-    self.contentType = url.hasSuffix(".pdf") ? .pdf : .html
+
+    // If the content type was specified and we know its PDF, use that
+    // otherwise fallback to using file extensions.
+    if let contentType = contentType, contentType == "application/pdf" {
```
satindar, line 29: should we use `contentType.contains("pdf")` just in case we get a site that has the mime type wrong?

satindar (approved):

apple/OmnivoreKit/Sources/Models/PageScrapePayload.swift
jacksonh: From what I can tell this is set by Safari not straight from the site headers, but yeah that seems a bit safer.

PR #708, opened 2022-05-25
Fix docker-compose migrate service env

sywhb:


gitstart-omnivore:
Hi @sywhb, Just wanted to point out that commenting out the updateMappings() call was enough, changing the environment variables was not needed.

Changing the environment variables causes connection issues instead.

Thanks for the help!

jacksonh (approved):

PR #709, opened 2022-05-25
Webhook

sywhb:


packages/api/src/routers/svc/webhooks.ts
```diff
@@ -0,0 +1,86 @@
...
+import express from 'express'
+import { readPushSubscription } from '../../datalayer/pubsub'
+import { getRepository } from '../../entity/utils'
+import { Webhook } from '../../entity/webhook'
+import axios, { Method } from 'axios'
+
+export function webhooksServiceRouter() {
+  const router = express.Router()
+
+  router.post('/trigger/:action', async (req, res) => {
```
jacksonh, line 13: nice. was wondering how you would do this.

packages/api/src/schema.ts
```diff
@@ -1617,6 +1713,8 @@ const schema = gql`
     generateApiKey(input: GenerateApiKeyInput!): GenerateApiKeyResult!
     unsubscribe(name: String!): UnsubscribeResult!
     subscribe(name: String!): SubscribeResult!
+    setWebhook(input: SetWebhookInput!): SetWebhookResult!
```
jacksonh, line 1716: Does `setWebhook` create a new one?

sywhb: It can be used to create a new one or update an existing one

packages/api/src/routers/svc/webhooks.ts
sywhb: I will create the pub/sub subscription

jacksonh (approved):

sywhb:
I will add the subscriptions of pub/sub in demo @jacksonh

PR #712, opened 2022-05-26
Better PDF upload handling on iOS

jacksonh:
Three main changes here:
- If a PDF isn't a local file, dont try to download and then
  upload it to the backend. Just treat it as a URL.

- If we are passed PDF data by an app like Files, upload it.

- Wrap the whole operation in a Background task so its not
  killed if the user taps dismiss during the upload of larger
  documents.


jacksonh:
I think this one is good to look at now @satindar

PR #714, opened 2022-05-26
Subscriptions Screen [iOS]

satindar:
Includes swipe to delete and tap to copy email address.


jacksonh:
Looks really good. Can we also add a snackbar to confirm after the subscription has been deleted?

satindar:
> Looks really good. Can we also add a snackbar to confirm after the subscription has been deleted?

Added the snacker in. It's helpful since we can show an error message if the network request fails.

jacksonh:
Yeah, definitely helps as a user to get that confirmation too.

jacksonh (approved):

PR #716, opened 2022-05-27
Api tokens

sywhb:


packages/api/src/apollo.ts
```diff
@@ -47,8 +48,18 @@ const contextFunc: ContextFunction<ExpressContext, ResolverContext> = async ({
...
-  if (token && jwt.verify(token, env.server.jwtSecret)) {
-    claims = jwt.decode(token) as Claims
+  if (token) {
+    try {
+      jwt.verify(token, env.server.jwtSecret) &&
+        (claims = jwt.decode(token) as Claims)
+    } catch (e) {
+      if (e instanceof jwt.JsonWebTokenError) {
+        logger.info(`not a jwt token, checking api key`, { token })
+        claims = await claimsFromApiKey(token)
```
jacksonh, line 58: 👍

packages/api/src/utils/auth.ts
```diff
@@ -1,16 +1,51 @@
...
+  // TODO: generate random string key
+  return uuidv4()
+}
+
+export const hashApiKey = (apiKey: string) => {
+  return crypto.createHash('sha256').update(apiKey).digest('hex')
+}
+
+export const claimsFromApiKey = async (key: string): Promise<Claims> => {
+  const hashedKey = hashApiKey(key)
```
jacksonh, line 26: Do we need to do this hash ourselves, can we do something like this? https://www.meetspaceapp.com/2016/04/12/passwords-postgresql-pgcrypto.html

sywhb: Yeah I think we can. 
Personally I prefer doing hashing/encryption (also the default value adding and unique checking) at the application level in case we need to migrate to another database like mysql. 
LOL. I guess I have PTSD

jacksonh: hahah we would never migrate to another data....oh wait I guess we did already.

jacksonh:
One thing I noticed is we have a few `if (!token || !jwt.verify(token, env.server.jwtSecret)) {` blocks in the routers that we might also want to handle the API tokens. Right now just `save` but its something we should be aware of. Maybe we can create a function for verification.

packages/api/src/utils/auth.ts
```diff
@@ -1,16 +1,81 @@
...
+  } catch (e) {
+    if (e instanceof jwt.JsonWebTokenError) {
+      console.log(`not a jwt token, checking api key`, { token })
+      claims = await claimsFromApiKey(token)
+    } else {
+      throw e
+    }
+  }
+
+  return claims
```
sywhb, line 80: I added a verification function here to get `claims` from either token or api key. Could use polymorphism here though

jacksonh:
This looks good. I guess next step we should probably add some hidden UI to access. Just a basic page `/settings/api`

jacksonh (approved):

sywhb:
> This looks good. I guess next step we should probably add some hidden UI to access. Just a basic page `/settings/api`

Yeah. I will add one

PR #718, opened 2022-05-27
Propagate pdf save errors

satindar:


apple/OmnivoreKit/Sources/Services/DataService/Queries/ArticleContentQuery.swift
```diff
@@ -154,18 +154,27 @@ extension DataService {
...
+                htmlContent: result.htmlContent,
+                highlights: result.highlights
+              )
+            } catch {
+              var message = "unknown error"
+              let basicError = (error as? BasicError) ?? BasicError.message(messageText: "unknown error")
+              if case let BasicError.message(messageText) = basicError {
+                message = messageText
+              }
+              continuation.resume(throwing: ContentFetchError.unknown(description: message))
```
jacksonh, line 170: this is cool

jacksonh (approved):
Looks right to me, we might want to try and update `status` on LinkedItem if the PDF successfully downloads.

PR #719, opened 2022-05-27
fix/pdf uploading improvements

jacksonh:
- clear core data when switch env
- handle pdf fetch errors properly
- set pdf state to success even if download has not completed
- fetch pdf data only if it's missing from LinkedItem
- Improve background uploading of PDFs


jacksonh:
WIP

apple/OmnivoreKit/Sources/App/AppExtensions/Share/ShareExtensionScene.swift
```diff
@@ -41,11 +42,16 @@ final class ShareExtensionViewModel: ObservableObject {
   }
 
   func savePage(extensionContext: NSExtensionContext?) {
+    backgroundTask = UIApplication.shared.beginBackgroundTask(withName: "BACKGROUND")
```
satindar, line 45: Should we add a UUID to this? Could be possible to have two background jobs 🤷🏽‍♂️

jacksonh: yeah, definitely. Maybe we generate the clientRequestId right at the start and pass this all the way through.

apple/OmnivoreKit/Sources/App/PDFSupport/PDFViewerViewModel.swift
```diff
@@ -18,25 +18,15 @@ public final class PDFViewerViewModel: ObservableObject {
   }
 
   public func dataURL(remoteURL: URL) -> URL {
-    if let storedURL = storedURL {
+    // TODO: we should probably not reach this point of localPdfURL is not set
```
satindar, line 21: I think we could get here when the user taps a cell where prefetch is still in process. Or would this cell not even show up until the pdf data is stored on device?

jacksonh: @satindar ah yeah, I couldn't figure out how to handle this, but we want it to be like fetching HTML content. We just show the spinner until the full PDF is downloaded.

apple/OmnivoreKit/Sources/Models/DataModels/FeedItem.swift
```diff
@@ -31,6 +31,7 @@ public extension LinkedItem {
   var unwrappedSlug: String { slug ?? "" }
   var unwrappedTitle: String { title ?? "" }
   var unwrappedPageURLString: String { pageURLString ?? "" }
+  var unwrappedLocalPdfURL: String { localPdfURL ?? "" }
```
satindar, line 34: Should probably remove this property since localPdfURL can actually be optional (the other ones here are just for properties that we know are there but get returned to us as optional because of Obj-C interop)

satindar (approved):

PR #723, opened 2022-05-30
Fix tweet not saved

sywhb:
Either we get the twitter by the link in the raw html or we get it from the iframe.
If we have a node with only one child element which has the `twitter-placeholder` class, keep it


packages/api/src/readability.d.ts
```diff
@@ -72,7 +72,7 @@ declare module '@omnivore/readability' {
      *
      * The response will be null if the processing failed (https://github.com/mozilla/readability/blob/52ab9b5c8916c306a47b2119270dcdabebf9d203/Readability.js#L2038)
      */
-    parse(): Readability.ParseResult | null
+    async parse(): Promise<Readability.ParseResult | null>
```
sywhb, line 75: Could not figure out an alternative to making this `async`

jacksonh (approved):

PR #729, opened 2022-06-01
Support labels in highlights

sywhb:
- Add labels in highlight properties
- Add labels in highlight type


jacksonh:
huh, not as hard as i expected in the data layer

sywhb:
> huh, not as hard as i expected in the data layer

Yeah. Just need to add those methods to set/delete labels in highlights.
I think searching labels in highlights might bring some challenges though

jacksonh (approved):

PR #735, opened 2022-06-02
Add chip to change sort order of linked items

satindar:


jacksonh:
> Maybe something like this would help communicate that relevance is used for search only?
> 
> Screen Shot 2022-06-02 at 11 38 02 AM

Ugh, yeah is there any way we could only display relevance if there was a search term? 😬

satindar:
> Ugh, yeah is there any way we could only display relevance if there was a search term? 😬

We could probably toggle the title between Newest/Relevance depending on if search is active. 

Or maybe just call it "search relevance"?

jacksonh (approved):

PR #739, opened 2022-06-03
Fixes for the sync state UI when errors occur

jacksonh:
If syncing fails, the item is still saved so we don't change the
title.

When syncing fails make sure the red error cloud is shown.


jacksonh:
A couple dumb mistakes I made here @satindar

satindar (approved):

PR #747, opened 2022-06-03
Disable sorting options that need API improvements

jacksonh:


jacksonh:
@satindar I'm going to work on an API update so we have a `readAt`, the problem with the current version using `updatedAt` is unread but recently saved articles will show up before recently read ones, so this option doesn't work well.

satindar (approved):

PR #756, opened 2022-06-04
Set spacing and background colour on share extension

jacksonh:


jacksonh:
@satindar hopefully the last one from me today.

satindar (approved):

PR #760, opened 2022-06-05
Fix issue with ShareExtension.js not executing on iOS

jacksonh:
Returning undefined here fixes issues where pages would fail
to scrape if there was no preview image available.


apple/OmnivoreKit/Sources/Views/SyncingIcon.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  File.swift
+//  
+//
+//  Created by Jackson Harper on 6/5/22.
+//
+
+import Foundation
```
satindar, line 8: code missing or do we need to delete this file?

jacksonh: @satindar yeah a little more to come...XCode just added it to the commit accidentally.

apple/OmnivoreKit/Sources/Views/ShareExtensionView.swift
```diff
@@ -188,12 +190,8 @@ public struct ShareExtensionChildView: View {
   }
 
   private func localImage(from: URL) -> Image? {
```
satindar, line 192: from is kinda confusing as a local variable. maybe url instead

`func localImage(from url: URL) -> Image?`

satindar (approved):

apple/OmnivoreKit/Sources/Services/DataService/Mutations/SavePDF.swift
```diff
@@ -127,7 +126,13 @@ public extension DataService {
     let selection = Selection<MutationResult, Unions.SaveResult> {
       try $0.on(
         saveError: .init { .error(errorCode: (try? $0.errorCodes().first) ?? .unknown) },
-        saveSuccess: .init { .saved(requestId: requestId, url: (try? $0.url()) ?? "") }
+        saveSuccess: .init {
```
jacksonh, line 129: I am just adding a comment here to test inline comments for something.

PR #761, opened 2022-06-05
Remove sf mono font. Add system default

satindar:


satindar:
this might not be totally right. i think it looks for a font called "undefined". which still works! but technically wrong

packages/web/components/templates/article/ArticleContainer.tsx
```diff
@@ -180,8 +187,13 @@ export function ArticleContainer(props: ArticleContainerProps): JSX.Element {
     fontSize,
     margin: marginOverride ?? props.margin ?? 360,
     lineHeight: lineHeightOverride ?? props.lineHeight ?? 150,
-    fontFamily: fontFamilyOverride ?? props.fontFamily ?? 'inter',
-    readerFontColor: highContrastFont ? theme.colors.readerFontHighContrast.toString() : theme.colors.readerFont.toString(),
+    fontFamily:
+      fontFamilyOverride ??
+      props.fontFamily ??
+      (props.isAppleAppEmbed ? undefined : 'inter'),
```
jacksonh, line 193: So here you think it gets set to a string named "undefined"?

jacksonh: I wonder if "unset" would work

satindar: yeah because `fontFamilyOverride` could be "undefined". I'll see if "unset" works just to make it less confusing.

jacksonh (approved):

PR #763, opened 2022-06-06
API key UI

sywhb:
- Add get api keys UI
- Add generateApiKey mutation in Web
- Add generate api key modal
- Add copy key modal
- Add revoke api key mutation in Web
- Add revoke api key modal
- Add default expires at and validation


jacksonh (approved):

sywhb:
> One small change, I think we can rename the `api-keys` file to `api` so `/settings/api` is the URL.

Okay!

PR #766, opened 2022-06-06
Fix text content missing when saving from extension

sywhb:
- Add <html> tag in prepared content from extension
- Add test case for danluu


jacksonh:
re-running checks on this one...

PR #780, opened 2022-06-09
Update bundle.js

jacksonh:
Removes the reader controls in app


jacksonh:
I think I can close this one out as the rebuilt bundle for the margins PR will obsolete it. /cc @satindar

PR #787, opened 2022-06-13
Fix: Inactive Reader Nav 

satindar:
For some reason the back button and web prefs button gets disabled when we open a SafariView as a sheet. No idea why but using `fullScreenCover` instead doesn't create the same issue.


jacksonh:
nice, seems like an improvement too

jacksonh (approved):

PR #788, opened 2022-06-13
iPhone Reader Margins

satindar:


jacksonh:
Going to test this one out locally a bit.

packages/web/components/templates/article/ArticleContainer.tsx
```diff
@@ -228,11 +243,6 @@ export function ArticleContainer(props: ArticleContainerProps): JSX.Element {
           '@md': {
             maxWidth: 1024 - styles.margin,
```
satindar, line 244: Maybe better to just use `appEmbedMaxWidthPercentage` here too. Could work better for iPad.

jacksonh:
Doesn't seem to cause any regressions with web

jacksonh (approved):

PR #813, opened 2022-06-16
advanced search

sywhb:
- Add title, description, content as keyword in search.ts
- Allow match keyword
- Allow matching title, description, content keyword in elastic


jacksonh:
Nice one. Love it

jacksonh (approved):

PR #816, opened 2022-06-17
WIP: Refactor/article content fetches

satindar:
Trying to move more functions into DataService. Also splitting functions that make a network request and write to core data. Hopefully isolating the networking code will make things easier to follow by reducing noise.


satindar:
I think this one is good now. Deploying to TF right now.

No huge changes here. Mostly moving things around and resolving warnings. Thinking of moving all the networking code to it's own package. That way we can thin down DataService and have less code around the harder to understand parts of the codebase.

Will work on that in the next PR.

jacksonh (approved):

Issue #752, opened 2022-06-03
When navigating to an article page on web, focus isn't moved to the article component so keyboard navigation doesn't work


jacksonh:
Can replicate on web starting on library navigate to an article with arrow keys, hit enter, attempt to scroll with arrow keys. Note how scrolling with arrow keys and space doesn't work.

PR #821, opened 2022-06-17
Do all scroll watching on the main window

jacksonh:
This moves all scrolling from child divs to the main window. This
improves our keyboard handling, as focus will be given to the
body element, not the child div when navigating by keyboard
commands, so arrow keys and space bar will work after navigating
to the reader with the keyboard commands.


jacksonh:
Fixes this issue: https://github.com/omnivore-app/omnivore/issues/752

PR #833, opened 2022-06-21
Use a Radix dialog instead of dropdown, add OpenDyslexic font, prefetch fonts

jacksonh:
The main change here is using Dialog instead of dropdown. The
dialog uses a fixed position, so when the underlying document
dimensions change based on a font change, the dialog will still
be positioned properly.


satindar (approved):
Looks good! Much more readable and neat trick on using the top and left absolute positions. Only suggestion is to rename `readeringSettings`. Maybe it was meant to be `readerSettings`?

PR #836, opened 2022-06-22
Some debug code

jacksonh:


jacksonh:
Will get rid of this one now

PR #849, opened 2022-06-23
Delete account API

satindar:
Trying to write some backend code 😬


packages/api/src/resolvers/user/index.ts
```diff
@@ -355,3 +360,38 @@ export const signupResolver: ResolverFn<
...
+>(async (_, { userID }, { models, claims, log, authTrx }) => {
+  const user = await models.user.get(userID)
+
+  if (!user || user.id !== claims.uid) {
+    return {
+      errorCodes: [DeleteAccountErrorCode.Unauthorized],
+    }
+  }
+
+  const deleteUserResult = await authTrx((tx) =>
```
sywhb, line 377: We might also need to delete this user's saved data in elastic

packages/api/src/resolvers/user/index.ts
```diff
@@ -355,3 +360,38 @@ export const signupResolver: ResolverFn<
...
 }
+
+export const deleteAccountResolver = authorized<
+  DeleteAccountSuccess,
+  DeleteAccountError,
+  MutationDeleteAccountArgs
+>(async (_, { userID }, { models, claims, log, authTrx }) => {
+  const user = await models.user.get(userID)
+
+  if (!user || user.id !== claims.uid) {
```
sywhb, line 371: We may return `USER_NOT_FOUND` if user is not found

sywhb (commented):
LGTM. we might also add a unit test for this api change

satindar:
> LGTM. we might also add a unit test for this api change

@sywhb Thanks for reviewing this. I added the tests for deleting a user...just not sure how great they are 😄 . They do pass though!

packages/api/src/resolvers/user/index.ts
```diff
@@ -355,3 +361,47 @@ export const signupResolver: ResolverFn<
...
+      errorCodes: [DeleteAccountErrorCode.Unauthorized],
+    }
+  }
+
+  const deleteUserResult = await authTrx((tx) =>
+    models.user.deleteUser(claims.uid, tx)
+  )
+
+  // delete this user's pages in elastic
+  await deletePagesByParam({ userId: userID }, { uid: userID, pubsub })
```
sywhb, line 389: hey @satindar i added deleting this user's pages in elastic

sywhb (approved):

packages/api/src/resolvers/user/index.ts
satindar: > hey @satindar i added deleting this user's pages in elastic

Nice! Do you mind deploying this to demo? I can do some manual testing on iOS after that.

sywhb: sorry for the delay. deployed to demo! @satindar

packages/api/src/resolvers/user/index.ts
```diff
@@ -355,3 +361,47 @@ export const signupResolver: ResolverFn<
...
+      resolver: 'deleteAccountResolver',
+      uid: claims.uid,
+    },
+  })
+
+  if ((deleteUserResult as UserData).id !== undefined) {
+    return { userID }
+  } else {
+    return {
+      errorCodes: [DeleteAccountErrorCode.Forbidden],
```
satindar, line 404: @sywhb Tried to delete an account a few times but it looks like it's ending up here and sending a FORBIDDEN error code. Doesn't seem to be a type error either since the account is still intact when I logout and in again 🤔 

Do you mind checking the logs to see if you find anything? I don't think I have access for them. Only have access to Sentry which doesn't have anything helpful here.

satindar: Screen Shot 2022-07-07 at 8 59 35 AM

sywhb: @satindar cool. I will see what I can find

sywhb: @satindar finally i figured out the issue is with the db permission. fixed it and deploying to demo

sywhb: @satindar deployed on demo! verified in my local env. let me know if it works :)

sywhb:
@satindar i merged this pr firstly to deploy the backend

PR #850, opened 2022-06-23
Fix search by is:read or is:unread getting empty result by using the correct property name

sywhb:


packages/api/test/resolvers/article.test.ts
```diff
@@ -1003,5 +1003,22 @@ describe('Article API', () => {
         expect(res.body.data.search.edges[4].node.id).to.eq(highlights[0].id)
       })
     })
+
+    context('when is:unread is in the query', () => {
```
jacksonh, line 1007: Thanks for getting a test in

jacksonh (approved):

packages/api/test/resolvers/article.test.ts
sywhb: i should have done this earlier 😢

PR #883, opened 2022-06-26
fix/unarchive article if saved again

sywhb:
- Update property unarchived=null instead of undefined in the page document in elastic
- Add test


packages/api/src/resolvers/article/index.ts
```diff
@@ -338,7 +338,7 @@ export const createArticleResolver = authorized<
         articleToSave.id = newPageId
       } else {
         // update existing page's state from processing to succeeded
-        articleToSave.archivedAt = archive ? saveTime : undefined
```
jacksonh, line 341: Classic. Think this has happened a few times.

jacksonh (approved):

PR #885, opened 2022-06-27
Fix getting nodes from null document element

sywhb:


sywhb:
https://ottawacitizen.com/news/national/defence-watch/canadian-special-forces-operating-in-ukraine-new-york-times-reports

jacksonh (approved):
Seems like some improvement in the test pages but always scary seeing changes in the test pages.

jacksonh:
Let's make sure we don't deploy this one to prod yet. Should test a bunch on demo first.

PR #893, opened 2022-06-28
add highlights to search results

sywhb:
- Add highlights in search item
- Include highlights in search page result in elastic
- Add test


jacksonh (approved):
nice one.

PR #910, opened 2022-06-30
macOS Share Extension Saving

satindar:
Updates share extension for macOS. Needs to work outside of a background op since they don;t exist on mac.


jacksonh (approved):

jacksonh:
Maybe what we should do on Mac is use something closer to the Web UI that we've been designing. 

Screen Shot 2022-07-08 at 13 30 27

jacksonh:
That square layout might make more sense on desktop

rupin90:
this is a HUGE PR 😓

PR #888, opened 2022-06-27
#863 - Allow opening the create api token dialog on the /settings/api page via a URL

rupin90:
Commit one - just includes to open the Generate API key modal.

Commit two - includes changes for prepopulation. 

Let me know any feedback.

PR #911, opened 2022-07-01
never expire tokens

jacksonh:
- Allow opening the create api token dialog on the /settings/api page via a URL
- adding prepopulation for input
- Set select color and background colors so OS themes dont override them
- fixed bug for input text showing undefined value
- Added never as one of the options
- Add support for "never" expiring tokens by setting to the Date MAX_AGE
- Better printing of never expiring tokens


jacksonh:
This is most of https://github.com/omnivore-app/omnivore/pull/888 -- I just left out the last few commits as I noticed some issues with changing selection values but I want to get never expiring tokens live

PR #924, opened 2022-07-05
Always fall back to scrapingbee if there is an exception

sywhb:


jacksonh:
Nice one, thanks. Going to figure out how to force this to throw an exception.

jacksonh:
Hey @sywhb I'll merge this one and try and test out on demo before deploying.

PR #928, opened 2022-07-06
Increase luminous for labels

rupin90:
Leveraging color2k library for some color conversion methods.

It is still not the best view but it does generate a lighter shade if luminance value is below 0.5.  I have also tried to do HSL values conversions and all but it would completely change the color not really make it glowy. Lot of conversions and lot of code but the output is not exactly how we want it to be.


jacksonh:
i think the light mode is really good now. The one issue I see is dark mode. When the label color is too dark I think we need to switch to a tinted white text.

rupin90:
I have been looking to see if there is anyway to lighten it up, I have played around with HSLA and RGBA, its not increasing the readability.

Yeah tinted :D

jacksonh:
Screenshot of GitHub dark mode

Screen_Shot_2022-07-11_at_16 09 41

jacksonh (approved):
Fantastic improvement. Just need to update to less lightening `lighten(props.color, 0.2)`

rupin90:
🙏🏽

PR #931, opened 2022-07-06
Reply to the original sender of the forwarded email

sywhb:


packages/api/src/routers/svc/emails.ts
```diff
@@ -89,6 +89,7 @@ export function emailsServiceRouter() {
         to: newsletterEmail.user.email,
         subject: `Fwd: ${data.subject}`,
         html: data.html,
+        replyTo: data.from,
```
sywhb, line 92: @jacksonh according to gnu:
```‘Reply-To’
An address to which replies should be sent, instead of ‘From’. This is used if, for some reason, your ‘From’ address cannot receive replies.```

jacksonh: ohh this is simpler than I expected.

jacksonh (approved):

jacksonh:
Will deploy and test

PR #934, opened 2022-07-07
fix/unsubscribe email not saved

sywhb:
- Move parseUnsubscribeHeader to index.ts
- Publish unsubscribe header info in the non-newsletter queue
- Save unsubscribe email address or http url of newsletter emails


sywhb:
@jacksonh

PR #951, opened 2022-07-12
feature/typeahead search

sywhb:
- update dependencies
- make title as search_as_you_type type in elastic
- add typeahead gql schema
- add searchAsYouType in elastic
- add typeahead api implementation
- rename typeahead to typeaheadSearch
- add test


sywhb:
hey @jacksonh this pr requires to reindex pages in elastic because we need to replace the type of `title` with `search_as_you_type`

jacksonh (approved):

PR #954, opened 2022-07-12
Extension: support right click saveurl and shortcut key to save current tab

jacksonh:


pkg/extension/src/scripts/background.js
```diff
@@ -644,6 +698,17 @@ function init () {
...
+  browserApi.contextMenus.create({
+    id: "log-selection",
+    title: "Save to Omnivore",
+    contexts: ["link"],
+    onclick: async function(obj) {
+      executeAction(async function (currentTab) {
+        await saveUrl(currentTab, obj.linkUrl)
+      })
+    },
+  });
```
sywhb, line 711: 👍

sywhb (approved):
LGTM

PR #968, opened 2022-07-15
debug content-fetch

sywhb:


jacksonh:
👁️

packages/puppeteer-parse/index.js
```diff
@@ -722,7 +744,7 @@ async function retrieveHtml(page) {
           }
         })();
       }),
-      page.waitForTimeout(5000), //5 second timeout
+      await page.waitForTimeout(1000), //5 second timeout
```
jacksonh, line 747: I think this comment might be inaccurate now.

jacksonh (commented):
Is this reverting to a previous commit? I noticed some changes like the articleSavingRequest and not sure if those changes are needed?

sywhb:
> Is this reverting to a previous commit? I noticed some changes like the articleSavingRequest and not sure if those changes are needed?

no, these are changes you made only in `content-fetch` but not in `puppeteer-parse` previously.
i did not check whether this is needed but just to sync them in case :)

packages/puppeteer-parse/index.js
sywhb: updated

PR #973, opened 2022-07-15
Use chrome-aws-lambda in GCF

jacksonh:
- Use chrome-aws-lambda in GCF
- Remove puppeteer-core package in GCF


jacksonh:
@sywhb I had to do this to get the GCF deploy to work.

sywhb:
> @sywhb I had to do this to get the GCF deploy to work.

@jacksonh ahh. it's the chromium args again?

sywhb (approved):

sywhb:
@jacksonh i think the chromium path in env is not right

jacksonh:
@sywhb Yeah I think when run in GCF it sets the path itself, but this package knows the correct paths to use.

sywhb:
@jacksonh yeah i think this package searches through many places so it can find the correct one :)

shall i merge it?

Issue #895, opened 2022-06-28
Improve table used on the `/settings/api` page


jacksonh:
Video shows a few problems with this table:

https://www.loom.com/share/a5b689adaa43447ba5a199afb6b4fc67

1. It looks like the header elements and the values aren't aligned, you can see when i resize the window the headers are not above the correct elements anymore.
2. The date should be formatted in a more readable local time
3. In mobile the view doesn't make a lot of sense. We might need to add labels to the values

jacksonh:
Two more issues I just found. These happen in Safari:

- The `Expires On` on the far right extends past the header when resizing
- The text colour for the input dropdown is white on a white backround

jacksonh:
I believe this resolves the last issue I brought up: https://github.com/omnivore-app/omnivore/commit/c2ce6a478b80a5f7a7db54f6f29e7d1bbc198e99

rupin90:
I think we can close this issue now? I am closing it for now, I can reopen if needed.

PR #976, opened 2022-07-15
#895 - Improve table used on the /settings/api page

rupin90:
Here is the behavior of the Table on different screen sizes. Finally was able to lay out the CSS. Somehow it wasn't as straight forward as I thought.
Leveraged NPM package - https://github.com/coston/react-super-responsive-table

Remaining Items:-
- [X] - border and background CSS for the table rows
- [X]  - Any feedback

https://user-images.githubusercontent.com/107437442/180309832-dcfebc78-6ffd-40fe-b609-c8729df81711.mp4


jacksonh:
I think this is just starting out, right? I notice that when scaling the table down the headers don't collapse correctly:

https://www.loom.com/share/e796b6f5aae444ec9aae7b736e4d5d76

jacksonh (approved):

jacksonh:
@rupin90 oh sorry i mean they dont move with the table, we do want to show them

jacksonh:
Hey @rupin90 I noticed two things:

1. It looks like we should import TableR instead of table. `Type error: '"../../components/elements/Table"' has no exported member named 'Table'. Did you mean 'TableR'?`

2. For some reason the z-index or something is higher with the table than the modals. When I open the modal all the headers of the table appear on top of the modal.

Screen Shot 2022-07-19 at 15 31 56

rupin90:
> Hey @rupin90 I noticed two things:
> 
> 1. It looks like we should import TableR instead of table. `Type error: '"../../components/elements/Table"' has no exported member named 'Table'. Did you mean 'TableR'?`
> 2. For some reason the z-index or something is higher with the table than the modals. When I open the modal all the headers of the table appear on top of the modal.
> 
> Screen Shot 2022-07-19 at 15 31 56

Thank you, wil look into it tonight 🙏

rupin90:
> 

I have updated the name to `TableNew`, I can update to any other name if you have any suggestions. This is simply to avoid conflict with the library. It is working fine.

Z-index issue is fixed now. Modal didn't have any z-index so I just gave it to Modal since that would be always be the top view if `open`.

rupin90:
Also I have tried different ways to customize the layout of the table - it has been very tricky so I have asked the community on how to customize the look and feel for the table - https://github.com/coston/react-super-responsive-table/issues/382

The `css` prop/attribute does not work on these elements. :(.

I also tried the typical way adding a className but the output is like 
``` <div class=[object object]>........ </div>``` 

Not sure if this was faced before in our codebase.

CC: @jacksonh @satindar

**Comment by jacksonh** [comment]
Hey @rupin90 this is really good. 

I just pushed a commit to show how to style these imported elements, i also renamed the import just so we could use `Table` elsewhere. 

with stitches we use this `stytled` function to make a component stylable. It also works on strings like `styled('br', { .. a bunch of css`. You can add styling in that `styled` call, or any component that you've made styled, then you can use the `css` attribute.

**Comment by rupin90** [comment]
> Hey @rupin90 this is really good.
> 
> I just pushed a commit to show how to style these imported elements, i also renamed the import just so we could use `Table` elsewhere.
> 
> with stitches we use this `stytled` function to make a component stylable. It also works on strings like `styled('br', { .. a bunch of css`. You can add styling in that `styled` call, or any component that you've made styled, then you can use the `css` attribute.

Aah nice, Thank you 🙏🏽 . Will do it just now.

File: packages/web/components/elements/Table.tsx [hunk 1]
```diff
@@ -102,7 +102,7 @@ export function TableNew(props: TableProps): JSX.Element {
           },
         }}
       >
-        <Table>
+        <ResponsiveTable>
```
rupin90: 👏🏽  Cool!

packages/web/components/elements/Table.tsx
```diff
@@ -106,7 +106,7 @@ export function Table(props: TableProps): JSX.Element {
           },
         }}
       >
-        <StyledTable>
+        <StyledTable css={{ /* can also style here now */ }}>
```
rupin90, line 109: Thank you

rupin90:
@jacksonh  - Thank you, this is PR ready

jacksonh:
Hey @rupin90 one thing I notice is in small there's a black outline we don't need:


Screen Shot 2022-07-21 at 15 41 05

rupin90:
> Hey @rupin90 one thing I notice is in small there's a black outline we don't need:
> 
> Screen Shot 2022-07-21 at 15 41 05

You mean the black border around the table? okay I can remove that.

rupin90:
### Hey @jacksonh   - how is this?

Screen Shot 2022-07-22 at 4 25 57 PM

jacksonh:
This looks great, thanks. I'm going to merge it now.

PR #989, opened 2022-07-19
email-login, email-signup, confirm-email and reset-password router

sywhb:
- add status to user table
- create `PENDING` user first after email-signup
- add sender email address in env
- add email confirmation & password reset token generation
- send confirmation email with link and token to pending user after email-signup
- add sinon and sinon-chai for faking sendMail
- replace login & signup gql resolver with rest router
- send confirmation email to pending user when trying to login
- add confirm-email router to set user status to `ACTIVE`
- remove membership from user table
- send password reset email with link and token to user when user forgot password
- add reset password router to update new password
- add tests for above
- use dynamic template in sendgrid


packages/api/src/events/reports/content_display_report_created.ts
```diff
@@ -28,7 +28,7 @@ export class ContentDisplayReportSubscriber
     if (!env.dev.isLocal) {
       // If we are in the local environment, just log a message, otherwise email the report
       await sendEmail({
-        to: 'feedback@omnivore.app',
+        to: env.sender.feedback,
```
jacksonh, line 31: 👍

packages/api/src/services/create_user.ts
```diff
@@ -68,6 +72,9 @@ export const createUser = async (input: {
         email: input.email,
         sourceUserId: input.sourceUserId,
         password: input.password,
+        status: input.pendingConfirmation
+          ? StatusType.Pending
+          : StatusType.Active,
```
jacksonh, line 77: 👍

jacksonh:
This all looks correct to me. We could use a few API tests I think.

sywhb:
> This all looks correct to me. We could use a few API tests I think.

Hey @jacksonh, I have added integration tests here.

jacksonh:
I'll fix these conflicts that I created :-/

packages/api/src/routers/auth/auth_router.ts
```diff
@@ -417,49 +437,203 @@ export function authRouter() {
...
+
+        await createUser({
+          email,
+          provider: 'EMAIL',
+          sourceUserId: email,
+          name,
+          username: lowerCasedUsername,
+          pictureUrl,
+          bio,
+          password: hashedPassword,
```
jacksonh, line 461: @sywhb should we be setting status pending here?

PR #1004, opened 2022-07-21
Add Save to Omnivore Android app

jacksonh:


jacksonh:
Tested this on device last night and it crashed, probably something with the preferences.

satindar (approved):

satindar:
@jacksonh Added another Android app in here. It's pretty messed up but going to merge this one anyways so we have at least have an Android base directory to work in.

PR #1010, opened 2022-07-25
Load Intercom on /support page

satindar:


jacksonh:
Cool, tested this one out and it works.

jacksonh (approved):

Issue #1018, opened 2022-07-26
Not able to open label modal on pdf reader page


sywhb:

jacksonh:
Working now.

PR #1019, opened 2022-07-26
Display toolbar on wide screens for PDFs

jacksonh:


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/1018

PR #1021, opened 2022-07-26
Save articles by sending emails to omnivore inbox

sywhb:
- Add function isProbablyArticle to test if a forwarded email contains an article to save
- Save article from forwarding emails
- Save emails with a subject containing `omnivore:` (case-insensitive) or from a sender email registered with us


jacksonh (approved):
Nice work. A few people have asked for this

PR #1031, opened 2022-07-28
Menu Component Start

rupin90:
Have leveraged a different library that fits our requirements - https://github.com/azouaoui-med/react-pro-sidebar
- [X] Menu is up with the basic Menu Items 
- [X] Story Book for Menu
- [X] Dynamic Items for Menu
- [x] CSS work for the menu theme/background

All the items of this PR are completed. This is PR ready. I have removed  some code in the `homefeedContainer.tsx` file because now the functionality is supported through Menu and Address bar URL. 
Here is the specific code that was removed - https://github.com/omnivore-app/omnivore/pull/1031/commits/f0d4c6fd7ffec0e74b70ed7fe305653f6ebc7234 

I just have separate commit in case we need to add this back.


jacksonh:
@rupin90 yeah regarding Q1, there are three sets of items, the first are static and are basically like this:

https://github.com/omnivore-app/omnivore/blob/main/packages/web/components/templates/homeFeed/HomeFeedContainer.tsx#L58

The second is the subscriptions and can be fetched like this: 

https://github.com/omnivore-app/omnivore/blob/main/packages/web/pages/settings/subscriptions.tsx#L12

Finally is the labels, which can be fetched like this: 

https://github.com/omnivore-app/omnivore/blob/main/packages/web/pages/settings/labels.tsx#L150

I think something like this could basically work:


```
type MenuItemData = {
   label: string
   query: string
   icon: string (probably should be an image or something)
}

// ... 

export const MenuComponent = () => {
     const { labels } = useGetLabelsQuery()
     const { subscriptions } = useGetSubscriptionsQuery()
   
     const menuItems = useMemo(() => {
         let result = [
              { label: 'Read Now', query: '!label:Newsletter', icon: 'none' } // probably a better way to do icons
             // list the rest of the items here (Inbox, Read Later, Highlights, Today, Newsletters)
         ]
         result += subscriptions.map((s) => { label: s.name, query: `subscription:"${s.subscription}"`, icon: 'subscription' }
         result += labels.map((l) => { label: l.name, query: `label:"${l.name}"`, icon: 'label' }
     }, [labels, subscriptions])
}
```

jacksonh:
For the style overrides, if you need to write pure css you could try putting it in `packages/web/styles/globals.css`

jacksonh:
Hey Rupin, could we make the menu bar sort of appear as part of the grid? I think the whole menu bar should be below the top elements like the search bar. 

Here is how its currently structured:

Screen Shot 2022-08-08 at 13 42 25


Here is kind of the structure we are looking for:

Screen Shot 2022-08-08 at 13 41 52

rupin90:
> Hey Rupin, could we make the menu bar sort of appear as part of the grid? I think the whole menu bar should be below the top elements like the search bar.
> 
> Here is how its currently structured:
> 
> Screen Shot 2022-08-08 at 13 42 25
> 
> Here is kind of the structure we are looking for:
> 
> Screen Shot 2022-08-08 at 13 41 52

okay, here is the update/screenshot on the Menu. Let me know if this is fine. Pushed a commit.

QUESTION - 

Screen Shot 2022-08-08 at 4 16 51 PM

jacksonh:
@rupin90 yeah that looks like what we are going for. I think we will have to play with the size but structurally seems closer to this design. 

Screen Shot 2022-08-09 at 11 07 54

jacksonh:
I think now its just all CSS?

rupin90:
> I think now its just all CSS?

Yeah, just for the theming, should be done in next few hours. 
@jacksonh

rupin90:
All the items of this PR are completed. This is PR ready. I have removed  some code in the `homefeedContainer.tsx` file because now the functionality is supported through Menu and Address bar URL. 


Here is the specific code that was removed - https://github.com/omnivore-app/omnivore/pull/1031/commits/f0d4c6fd7ffec0e74b70ed7fe305653f6ebc7234 

I just have separate commit in case we need to add this back.

CC: @jacksonh

jacksonh:
Hey @rupin90 this is good. I have a new library page I've been working on, what I will do is remove the menu from /home and stick it in this temp library page for a bit while we test it out.

jacksonh:
I think the collapse feature might be really nice for us too. Good choice on this library.

rupin90:
> Hey @rupin90 this is good. I have a new library page I've been working on, what I will do is remove the menu from /home and stick it in this temp library page for a bit while we test it out.

Okay sounds good, I saw the commits, thank you for setting the active menu items. 

I saw that you have added the code back  in homefeedcontainer.tsx.  Is that temporary or we are using it for something? Just trying to gain an understanding.

rupin90:
> I think the collapse feature might be really nice for us too. Good choice on this library.

yeah collapse feature is nice especially for mobile/small screen devices, but I believe for that we will need icons for all the menu items. 🙏🏽 .

packages/web/components/templates/Menu.tsx
```diff
@@ -0,0 +1,171 @@
...
+    subscriptions.map((s) =>
+      subscriptionsList.push({
+        label: s.name,
+        query: `subscription:"${s.subscription}"`,
+        icon: 'subscription',
+        href: `?q=subscription:"${s.subscription}"`,
+      })
+    )
+  }
+  return [{ labels: [...labelsList], subscriptions: [...subscriptionsList] }]
```
jacksonh, line 61: I don't think this needs to be an array anymore. I think if we just return`{ labels: [...labelsList], subscriptions: [...subscriptionsList] }` some of the below code gets simpler.

jacksonh: I just pushed a small change to do this.

PR #1035, opened 2022-07-29
feature/sync api

sywhb:
- Add `updatesSince` API in GQL schema
- Add DELETED state to the page doc
- Add soft delete of pages by updating page state to DELETED
- Add `updatesSince` API resolver
- Add integration tests for `updatesSince` API resolver


jacksonh (approved):
Be good to do some testing on demo. Mostly making sure deleted items don't show up because of soft delete

Issue #999, opened 2022-07-21
Readwise integration


ashinonyx:
Really liking Omnivore so far, it's just missing one thing that would make it almost perfect for me: Readwise integration!

I've reached out to them and they have noted my request, but they also suggested that this could be worked on Omnivore's end as well, using [Readwise's API](https://readwise.io/api_deets).

jacksonh:
Hey @ashinonyx yeah I think the Readwise team is correct here. Its probably up to use to push highlights into Readwise. 

Can you just quickly explain what you'd like to do? I assume what you want is, when you add a highlight or note in Omnivore, its pushed to Readwise?

ashinonyx:
@jacksonh Yes, that's exactly it. I'd like to see highlights made in Omnivore get pushed to Readwise

jacksonh:
@ashinonyx thanks. I'll look at their API again, I think this should be possible, its not too different from our webhooks support.

ashinonyx:
@jacksonh Sounds great! Thanks for the help!

jacksonh:
Hey @ashinonyx we are starting to test this integration now. Would you be interested in testing?

jonoberg:
@jacksonh I can also help with testing

jacksonh:
@jonoberg awesome. I will shoot you an email tomorrow with some instructions. Thanks!

ashinonyx:
@jacksonh sure, I'd be happy to help!

jacksonh:
@ashinonyx @jonoberg if you go to this URL, you can click the Readwise button: https://omnivore.app/settings/integrations -- from there you can add your readwise API token to Omnivore. That should start a process where your highlights are synced into readwise. 

If you want to disconnect, you can just go to the same page and click `Remove`

jacksonh:
Actually, we just found one issue we need to work out. Sorry about that, should be ready this weekend.

jacksonh:
@jonoberg @ashinonyx sorry for the delay, this is working now:

To test:

go to this URL, and click the Readwise button: https://omnivore.app/settings/integrations -- from there add your Readwise API token (there is a link to get your token from Readwise). That should start a process where your highlights are synced into Readwise. Historical highlights will be imported, and any new highlights in Omnivore.

jonoberg:
I just set it up and it looks to have worked great. Thanks for making this happen!

howarddo2208:
[...]

PR #1036, opened 2022-08-01
feature/readwise integration api

sywhb:
- Add integrations table in db
- Add integration entity
- Add createIntegration API to the GQL schema
- Add createIntegration API implementation


sywhb:
#999

sywhb:
Adding integration tests...

packages/db/migrations/0091.do.integrations.sql
```diff
@@ -0,0 +1,25 @@
...
+-- Description: Create integrations table
+
+BEGIN;
+
+CREATE TYPE omnivore.integration_type AS ENUM ('READWISE');
+
+CREATE TABLE omnivore.integrations (
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v1mc(),
+    user_id uuid NOT NULL REFERENCES omnivore.user ON DELETE CASCADE,
+    "type" omnivore.integration_type NOT NULL,
```
jacksonh, line 12: 👍

packages/api/src/resolvers/integrations/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  try {
+    const user = await getRepository(User).findOneBy({ id: uid })
+    if (!user) {
+      return {
+        errorCodes: [SetIntegrationErrorCode.Unauthorized],
+      }
+    }
+
+    const integrationToSave: Partial<Integration> = {
```
jacksonh, line 32: Do we need to sanitize these input values? I guess GQL kind of does already, and then the DB will verify that its a valid type.

packages/api/src/services/integrations.ts
```diff
@@ -0,0 +1,148 @@
...
+): Promise<boolean> => {
+  switch (type) {
+    case IntegrationType.Readwise:
+      return validateReadwiseToken(token)
+    default:
+      return false
+  }
+}
+
+const validateReadwiseToken = async (token: string): Promise<boolean> => {
```
jacksonh, line 51: Good idea validating right here.

jacksonh (approved):

jacksonh:
I guess we will have to do some testing on demo.

jacksonh (approved):

PR #1043, opened 2022-08-04
Text to speech (iOS)

satindar:


jacksonh:
Yeah, i think the iOS voice just isn't good enough for us vs using something like Amazon Polly.

satindar:
> Yeah, i think the iOS voice just isn't good enough for us vs using something like Amazon Polly.

The voice can be changed in device settings. So for users that use speech a lot they'll get a voice they're familiar with. Can try some more once we move forward on this feature. Also implement Polly so we can compare.

TBH the main reason I'd like to go with native if we can is to avoid the AWS SDK 😱

satindar:
Closing since this was just for testing the iOS synthesizer out.

PR #1044, opened 2022-08-04
Fix not getting correct sender email address from forwarding email by Gmail rules

sywhb:
Gmail will add some headers when forwarding emails by some rules:
* `X-Forwarded-To: recipient@omnivore.app`
* `X-Forwarded-For: sender@omnivore.app recipient@omnivore.app`
and the `from` and `to` headers become the original sender and recipient's email addresses.

In this case, we are sending the original sender email address to Pub/Sub instead of the user's Omnivore account email address.
We fixed it by getting the Omnivore account email from the `X-Forwarded-For` header and sending it to Pub/Sub.

### DIFF
--- a/packages/inbound-email-handler/src/index.ts
+++ b/packages/inbound-email-handler/src/index.ts
@@ -76,7 +76,7 @@ export const inboundEmailHandler = Sentry.GCPFunction.wrapHttpFunction(
       const forwardedTo = headers['x-forwarded-to']?.toString().split(',')[0]
       // x-forwarded-for is a space separated list of email address
       // the first one is the forwarding email sender and the last one is the recipient
-      // e.g. 'X-Forwarded-For: sender@omnivore.app recipient@omnivore.app
+      // e.g. 'X-Forwarded-For: sender@omnivore.app recipient@omnivore.app'
       const forwardedFrom = headers['x-forwarded-for']?.toString().split(' ')[0]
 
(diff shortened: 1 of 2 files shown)


packages/inbound-email-handler/src/index.ts
```diff
@@ -42,51 +42,60 @@ export const getNewsletterHandler = (
...
+      const html = parsed['html']
+      const text = parsed['text']
+
+      // headers added when forwarding email by some rules in Gmail
+      // e.g. 'X-Forwarded-To: recipient@omnivore.app'
+      const forwardedTo = headers['x-forwarded-to']?.toString().split(',')[0]
+      // x-forwarded-for is a space separated list of email address
+      // the first one is the forwarding email sender and the last one is the recipient
+      // e.g. 'X-Forwarded-For: sender@omnivore.app recipient@omnivore.app
+      const forwardedFrom = headers['x-forwarded-for']?.toString().split(' ')[0]
```
jacksonh, line 80: I guess this is the key to the whole thing here.

jacksonh (approved):

packages/inbound-email-handler/src/index.ts
sywhb: Yeah, I have updated the description.
To sum up, the issue is Gmail changes the `from` to the original sender email address instead of the Omnivore account email address.

PR #1046, opened 2022-08-05
Use parsed from, to and subject from email headers

sywhb:
Screenshot 2022-08-05 at 10 03 56 AM
The email subject in headers may contain encoded strings so we need to use parsed ones instead.


sywhb:
Looks like this works:
Screenshot 2022-08-05 at 11 44 58 AM

jacksonh (approved):

PR #1060, opened 2022-08-10
fix/integrations db permissions

sywhb:
- Grant permissions of integrations table to omnivore_user
- Separate migrations from test
- Delete user with right permission


sywhb:
@jacksonh This is ready for review :)

sywhb:
Hey @jacksonh, is it good to merge now?

jacksonh (approved):

PR #1065, opened 2022-08-11
Android Login 

satindar:


jacksonh:
I'm going to merge this one so we can do proper Android PRs

jacksonh:
Oh nevermind, we need to get the backend parts working first.

packages/api/src/routers/auth/mobile/mobile_auth_router.ts
```diff
@@ -16,8 +16,9 @@ export function mobileAuthRouter() {
   const router = express.Router()
 
   router.post('/sign-in', async (req, res) => {
-    const { token, provider } = req.body
-    const payload = await createMobileSignInResponse(token, provider)
+    const { token, provider, source } = req.body
```
jacksonh, line 19: @satindar so existing `iOS` is not passing source yet, right?

packages/api/src/routers/auth/google_auth.ts
```diff
@@ -57,16 +57,22 @@ export const validateGoogleUser = async (
...
+const iosClientId = env.google.auth.iosClientId
+const webClientId = env.google.auth.clientId
+
+const googleWebClient = new OAuth2Client(webClientId)
+const googleIOSClient = new OAuth2Client(iosClientId)
 
 export async function decodeGoogleToken(
-  idToken: string
+  idToken: string,
+  isAndroid: boolean
```
jacksonh, line 68: Instead of passing `isAndroid` in here, it might be more clear to have the calling code pass the clientId. Something like:

```
router.post('/sign-in', async (req, res) => {
    const { token, provider } = req.body
    const payload = await createMobileSignInResponse(token, provider)
    const { token, provider, source } = req.body
    const payload = await createMobileSignInResponse(clientIdForSource(source), token, provider)
```

jacksonh:
I'll merge this one as-is so we can test it better on demo.

PR #1066, opened 2022-08-11
feature/tts backend

sywhb:
- Create Amazon Polly text to speech integration
- Add tests
- Upload audio file to GCP with public access right


packages/api/src/utils/uploads.ts
```diff
@@ -102,21 +101,18 @@ export const generateUploadFilePathName = (
...
+  filePath: string,
   data: Buffer,
-  contentType: string
+  options?: { contentType?: string; public?: boolean },
+  selectedBucket?: string
 ): Promise<void> => {
-  // if (env.dev.isLocal) {
-  //   return
-  // }
+  await storage
```
jacksonh, line 110: I was thinking about this this morning, and I think we can make these audio files non-public. We just need to returned a signed URL that can download the file when we do the redirect.

packages/db/migrations/0093.do.speech.sql
```diff
@@ -0,0 +1,28 @@
...
+-- Description: Add speech table containing text to speech audio_url and speech_marks
+
+BEGIN;
+
+CREATE TABLE omnivore.speech (
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v1mc(),
+    user_id uuid NOT NULL REFERENCES omnivore.user ON DELETE CASCADE,
+    elastic_page_id TEXT NOT NULL,
+    voice text,
+    audio_url text NOT NULL,
```
jacksonh, line 12: If we did do the signed URLs we'd have to store this differently, we'd probably have to store the bucket + file path and then create a signed URL when we return.

packages/api/src/routers/article_router.ts
```diff
@@ -61,5 +70,97 @@ export function articleRouter() {
       articleSavingRequestId: result.id,
     })
   })
+
+  router.get(
+    '/:id/:outputFormat',
+    cors<express.Request>(corsConfig),
+    async (req, res) => {
+      const id = req.params.id
+      const outputFormat = req.params.outputFormat
```
jacksonh, line 79: Can we verify here that `id` and `outputFormat` are not null? Maybe verify `id` not null and `outputFormat` is in a list of items like `['mp3', 'speechMarks'].contains(outputFormat)`

packages/api/src/utils/uploads.ts
sywhb: Yes, it makes sense to me

jacksonh (approved):
I think we could re-work this a little to have private MP3 files, we'd need to return signed URLs from the get MP3 API, but it would provide some extra security so its probably worth the extra effort.

packages/db/migrations/0093.do.speech.sql
sywhb: Yeah. I will also add a `state` field

packages/api/src/routers/article_router.ts
sywhb: Yes, def

packages/api/src/entity/speech.ts
```diff
@@ -0,0 +1,48 @@
...
+  id!: string
+
+  @ManyToOne(() => User, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'user_id' })
+  user!: User
+
+  @Column('text')
+  elasticPageId!: string
+
+  @Column('text', { default: '' })
```
jacksonh, line 31: nice

packages/api/src/routers/svc/speech.ts
```diff
@@ -0,0 +1,97 @@
...
+import { buildLogger } from '../../utils/logger'
+
+const logger = buildLogger('app.dispatch')
+
+export function speechServiceRouter() {
+  const router = express.Router()
+
+  router.options('/', cors<express.Request>({ ...corsConfig, maxAge: 600 }))
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  router.post('/', async (req, res) => {
```
jacksonh, line 18: I think we need to authenticate this API? I don't see a token validation check here

jacksonh (changes requested):
Could an external unauthenticated call be made to `/svc/text-to-speech'`?

sywhb:
Hey @jacksonh, I added a JWT token validation to `/svc/text-to-speech`

packages/api/src/routers/svc/speech.ts
```diff
@@ -0,0 +1,113 @@
...
+    const token = req.query.token as string
+    try {
+      if (!(await getClaimsByToken(token))) {
+        logger.info('Unauthorized request', { token })
+        return res.status(200).send('UNAUTHORIZED')
+      }
+    } catch (error) {
+      logger.error('Unauthorized request', { token, error })
+      return res.status(200).send('UNAUTHORIZED')
+    }
```
sywhb, line 33: Added authentication here

jacksonh: Does the user token get passed when this is called?

packages/api/src/utils/createTask.ts
```diff
@@ -325,4 +328,45 @@ export const enqueueSyncWithIntegration = async (
...
+  pageId: string
+): Promise<string> => {
+  const { GOOGLE_CLOUD_PROJECT } = process.env
+  const payload = {
+    userId,
+    pageId,
+  }
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const token = await signToken({ uid: userId }, env.server.jwtSecret, {
```
jacksonh, line 342: Ah I see it, nice way of doing it.

jacksonh (approved):

PR #1082, opened 2022-08-15
WIP: iOS UX and Audio Sessions for TTS

jacksonh:
- add text to speech play button on web links
- remove extra nonsense line of code
- Sample of playing back an mp3
- WIP: AudioSession and miniplayer for iOS TTS


apple/OmnivoreKit/Sources/App/Views/WebReader/WebReaderContainer.swift
```diff
@@ -68,6 +71,34 @@ struct WebReaderContainerView: View {
         .scaleEffect(navBarVisibilityRatio)
         Spacer()
       #endif
+      if FeatureFlag.enableTextToSpeechButton {
+        Button(
+          action: {
```
satindar, line 76: Maybe we could move this action into audioSession? So we would just call something like `audioSession.handleUserTap` and then we could use it elsewhere.

apple/OmnivoreKit/Sources/App/Views/AudioPlayer/MiniPlayer.swift
```diff
@@ -0,0 +1,87 @@
...
+
+  init<PresentingView>(
+    presentingView: PresentingView
+  ) where PresentingView: View {
+    self.presentingView = AnyView(presentingView)
+  }
+
+  public var body: some View {
+    GeometryReader { geometry in
+      ZStack(alignment: .center) {
```
satindar, line 26: This is pretty neat! Our good friend ZStack making to work again.

satindar:
This looks great so far! The EnvironmentObject use looks correct to me.

apple/OmnivoreKit/Sources/App/Views/WebReader/WebReaderContainer.swift
jacksonh: Yeah I tried to do that and then had to move it out, probably still could. There are two different play buttons displayed and they act a little differently. The play button on the top nav bar will play the current item, and if you are listening to the current item, pause it. But if you are listening to a different item and hit that play button, it will play the new item

The play button on the miniplayer just toggles play/pause though.

apple/OmnivoreKit/Sources/Utils/ImageCache.swift
```diff
@@ -27,24 +27,29 @@ public final class ImageCache {
...
-    cache.totalCostLimit = 1024 * 1024 * 50 // 50 MB
+    // cache.totalCostLimit = 1024 * 1024 * 1024 * 50 // 50 MB
+    print("CREATING IMAGE CACHE: ", self)
   }
 
   private func image(_ url: URL) -> PlatformImage? {
     var cachedImage: PlatformImage?
     queue.sync {
-      cachedImage = cache.object(forKey: url as AnyObject)
+      cachedImage = cache.object(forKey: NSString(string: url.absoluteString))
```
jacksonh, line 40: @satindar I'll remove these debugging lines, but I discovered when we use `URL` as the key the cache is never hit. I guess its not doing a proper equality check or something.

satindar: Oh wow! Guess I never tested this thoroughly 🙃

jacksonh: I wonder if it used to? Also it might not even matter because the NSURLCache stuff could be used when we try to re-download. 

Kind of weird though, I didn't see anything on that blog post about it.

apple/OmnivoreKit/Sources/Services/AudioSession/AudioSession.swift
```diff
@@ -110,4 +113,41 @@ public class AudioSession: ObservableObject {
       print("play time in ms: ", Int(player.currentTime * 1000))
     }
   }
+
+  func setupRemoteControl() {
+    UIApplication.shared.beginReceivingRemoteControlEvents()
```
satindar, line 118: Nice

jacksonh:
I think this is probably mergable now. It would be awesome to do a testflight build. We will need to make sure @sywhb's API stuff is deployed, but this version does work. Just slow to do the initial load because thats triggering the conversion.

apple/OmnivoreKit/Sources/Services/AudioSession/AudioSession.swift
```diff
@@ -0,0 +1,240 @@
...
+
+    Task {
+      do {
+        try await downloadAudioFile(pageId: pageId)
+      } catch {
+        print("FAILED TO DOWNLOAD AUDIO URL")
+        print(error.localizedDescription)
+      }
+
+      DispatchQueue.main.async {
```
jacksonh, line 76: Probably need to figure out a cleaner way of doing this.

apple/OmnivoreKit/Sources/Services/AudioSession/AudioSession.swift
```diff
@@ -0,0 +1,240 @@
...
+    state = .loading
+
+    let pageId = item!.unwrappedID
+
+    Task {
+      do {
+        try await downloadAudioFile(pageId: pageId)
+      } catch {
+        print("FAILED TO DOWNLOAD AUDIO URL")
+        print(error.localizedDescription)
```
jacksonh, line 73: Right now we intentionally fall through here, so the error gets detected and handled on the main thread.

jacksonh:
Probably best to merge this one now, and we can fix bugs off of it.

PR #1088, opened 2022-08-15
Adding background color to active item and cleaning up

rupin90:
- Adding background color to the active menu item for both the themes
 - Setting active status for the selected menu item

Video - 
https://www.awesomescreenshot.com/video/10518687?key=8a6c52bdb4ccf2a908058c2b3b851bf2


jacksonh:
Ah this looks so good. Thank you!

jacksonh (approved):

PR #1089, opened 2022-08-17
Fix the style for cancel buttons on non-login pages

jacksonh:


sywhb (approved):
Fast!

jacksonh:
haha PR that broke it was fast too

PR #1099, opened 2022-08-18
Fix: 1091 - Add padding on the left of the menu

rupin90:
Active background didn't have left padding, so added that.


jacksonh:
Looks so nice.

jacksonh (approved):

PR #1101, opened 2022-08-18
Fix/1096 - Added line seprator above labels and subscriptions

rupin90:
Merging 1091 and 1092 fixes before would make this PR look correct.

### DIFF
--- a/packages/web/styles/menu.css
+++ b/packages/web/styles/menu.css
@@ -25,8 +25,12 @@
   padding: 5px 20px;
 }
 
-.pro-sidebar .pro-menu .pro-menu-item.pro-sub-menu {
- border-top: 1px solid var(--colors-grayBg);
+.pro-sidebar .pro-menu .pro-menu-item.pro-sub-menu:before {
+  content: '';
+  display: block;
+  width: 90%;
+  border: 0.5px solid var(--colors-grayBg);
+  margin-left: 10px;
 }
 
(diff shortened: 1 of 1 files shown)


jacksonh:
Hey Rupin, I think there's one small issue with this. The borders extends all the way to the right for me, and is maybe a bit to far on the left. I think its kind of a pain to fix this because we can't just use border...right?

Screen Shot 2022-08-19 at 13 29 53

rupin90:
> Hey Rupin, I think there's one small issue with this. The borders extends all the way to the right for me, and is maybe a bit to far on the left. I think its kind of a pain to fix this because we can't just use border...right?
> 
> Screen Shot 2022-08-19 at 13 29 53

Apparently even after a decade of Frontend legendary change/ upgrade in the world, CSS never got the `border-length` kinda property. 🤷🏽 . lol.

Used an alternative to control the border length, I believe this should work.

CC: @jacksonh

jacksonh (approved):

jacksonh:
CSS!

jacksonh:
@rupin90 is there an easy way to add a little more padding between `Newsletters` and `Labels`? I think it just needs 2px more maybe. It feels a little too tight, but otherwise this is good. We can try to do that separately.

jacksonh (approved):

rupin90:
> @rupin90 is there an easy way to add a little more padding between `Newsletters` and `Labels`? I think it just needs 2px more maybe. It feels a little too tight, but otherwise this is good. We can try to do that separately.

Will do that separately

PR #1109, opened 2022-08-19
Fix/1080 - Adding background on the hover of a menu item

rupin90:
video - https://www.awesomescreenshot.com/video/10600482?key=4a18f66abb92df9d41d15c2252dfd9dc


jacksonh (approved):

jacksonh:
Ah this is so much nicer.

jacksonh (approved):

PR #1114, opened 2022-08-22
feature/queue tts in the api

sywhb:
- Add function to set speech to failed state
- Throw error if synthesizing was cancelled
- Set speech state to FAILED if encountered error during transcription
- Stop queueing transcription after calling createArticle API
- Enqueue TTS task in the get speech of article API if speech not found

### DIFF
--- a/packages/api/src/routers/article_router.ts
+++ b/packages/api/src/routers/article_router.ts
@@ -103,18 +103,28 @@ export function articleRouter() {
         order: {
           createdAt: 'DESC',
         },
+        relations: ['user'],
       })
-      if (existingSpeech?.state === SpeechState.COMPLETED) {
-        logger.info('Found existing completed speech', {
-          audioUrl: existingSpeech.audioFileName,
-          speechMarksUrl: existingSpeech.speechMarksFileName,
-        })
-        return res.redirect(await redirectUrl(existingSpeech, outputFormat))
-      }
-      if (existingSpeech?.state === SpeechState.INITIALIZED) {
-        logger.info('Found existing in progress speech')
-        // retry later
-        return res.status(202).send('Speech is in progress')
+      if (existingSpeech) {
+        if (existingSpeech.user.id !== uid) {
+          logger.info('User is not allowed to access speech of the article', {
+            userId: uid,
+            articleId,
+          })
+          return res.status(401).send({ errorCode: 'UNAUTHORIZED' })
+        }
+        if (existingSpeech.state === SpeechState.COMPLETED) {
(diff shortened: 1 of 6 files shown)


packages/api/src/routers/article_router.ts
```diff
@@ -72,12 +72,13 @@ export function articleRouter() {
...
-    '/:id/:outputFormat',
+    '/:id/:outputFormat/:voice?',
     cors<express.Request>(corsConfig),
     async (req, res) => {
-      const id = req.params.id
+      const articleId = req.params.id
       const outputFormat = req.params.outputFormat
-      if (!id || !['mp3', 'speech-marks'].includes(outputFormat)) {
+      const voice = req.params.voice
+      if (!articleId || !['mp3', 'speech-marks'].includes(outputFormat)) {
```
jacksonh, line 81: nice

packages/api/src/routers/article_router.ts
```diff
@@ -107,69 +114,30 @@ export function articleRouter() {
...
-      } catch (error) {
-        logger.error('Text to speech error', { error })
-        // update state
-        await getRepository(Speech).update(speech.id, {
-          state: SpeechState.FAILED,
-        })
-        res.status(500).send('Text to speech error')
-      }
+      // enqueue a task to convert text to speech
+      const taskName = await enqueueTextToSpeech(uid, speech.id)
```
jacksonh, line 138: Making sure I understand: if the speech has been set to INITIALIZED, we return 202, if there is no speech, we create an initialized item and return 202 also. Right?

sywhb: Yes

packages/api/src/routers/svc/speech.ts
```diff
@@ -21,68 +21,57 @@ export function speechServiceRouter() {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       body: req.body,
     })
+    const { userId, speechId } = req.body as {
```
jacksonh, line 24: Is there a way we can do the token check here before setting failure state? Otherwise I think people could set existing speeches to failed that they don't own.

packages/api/src/routers/article_router.ts
sywhb: We return different response body

packages/api/src/routers/svc/speech.ts
sywhb: My mistake here. I will validate token first

jacksonh (approved):

PR #1115, opened 2022-08-22
Make a full screen audio player that can be swapped with the miniplayer

jacksonh:


jacksonh:
@satindar can you take a look at this one? I think testing will only work on demo, and the first time you try to play an article you will get an error message, but if you wait a few seconds then try again it should work.

apple/OmnivoreKit/Sources/Services/AudioSession/AudioSession.swift
```diff
@@ -19,24 +19,35 @@ public enum AudioSessionState {
...
+  @Published public var duration: TimeInterval = 0
+  @Published public var timeElapsedString: String?
+  @Published public var durationString: String?
+
   let appEnvironment: AppEnvironment
   let networker: Networker
 
   var timer: Timer?
   var player: AVAudioPlayer?
+  var downloadTask: Task<Void, Error>?
```
satindar, line 43: nice. better way to track cancellations.

apple/OmnivoreKit/Sources/Services/AudioSession/AudioSession.swift
```diff
@@ -225,6 +326,12 @@ public class AudioSession: ObservableObject {
       throw BasicError.message(messageText: "audioFetch failed. no response or bad status code.")
     }
     print("httpResponse: ", httpResponse)
+    if let httpResponse = result?.1 as? HTTPURLResponse, httpResponse.statusCode == 202 {
```
satindar, line 329: Do we need to poll the API here so we know when the download is complete?

jacksonh: Oh yeah that is a good idea, maybe we could do like a permanent snackbar in that state or a different indicator than the spinner

Or maybe just some data on the miniplayer...

apple/OmnivoreKit/Sources/Services/AudioSession/AudioSession.swift
jacksonh: Yeah I really like that `Task.isCancelled` thing, so you don't even need to track state.

jacksonh:
I'll work on the pending stuff in a separate PR. Probably need to make some UI for it.

PR #1117, opened 2022-08-22
Added space in menu items so it doesn't look so tight

rupin90:
Added space in menu items so it doesn't look so tight


jacksonh (approved):

jacksonh:
Will merge this one now 👍

PR #1130, opened 2022-08-24
Feature 1079 -Updating list view to the latest design

rupin90:
- [x] Adding Author
- [x] Adding origin
- [x] Adding Card menu

### Design -
Screen Shot 2022-08-24 at 12 45 14 AM


jacksonh:
Whaaaa. So fast

rupin90:
> Whaaaa. So fast

haha no it wasn't ready.

rupin90:
@jacksonh  - Let me know if this is what was all needed. It seemed like most of the components were there just some CSS was needed to match the design. Let me know any feedback. Thank you.

jacksonh:
This is what I am seeing locally:

Screen Shot 2022-08-29 at 12 03 34

Looks like image layout and maybe author name are the main issues.

rupin90:
@jacksonh  
- So I updated the view with the condition that image won't be displayed if it is List Layout.  So for the Grid Layout the image for the card will be displayed. 
- I have increased the font weight for the author and updated the margin around. 
- There is a pop up menu for both Grid and List layout.


I think this PR now fixes the necessary changes for both Grid and List layout according to the design.

rupin90:
Merging this should tackle 1079 and 1097 github issues.

jacksonh:
Looks good. There actually is a design for a version with images too, but lets merge this one. I think we might tweak a bit in design also. Maybe can tighten it up a little bit and make it more compact.

jacksonh (approved):

PR #1131, opened 2022-08-24
feature/transcription rule

sywhb:
- Add listenedAt to Page in elastic
- Update listenedAt in the API
- Add function to test if a page should be transcribed
- Add a pubsub subscription listening on the new article to automatically synthesize article


packages/api/src/routers/article_router.ts
```diff
@@ -118,6 +119,13 @@ export function articleRouter() {
             audioUrl: existingSpeech.audioFileName,
             speechMarksUrl: existingSpeech.speechMarksFileName,
           })
+          await updatePage(
+            existingSpeech.elasticPageId,
+            {
+              listenedAt: new Date(),
```
jacksonh, line 125: nice. This will probably be very useful

packages/api/src/services/speech.ts
```diff
@@ -1,9 +1,77 @@
...
+export const shouldSynthesize = async (
+  userId: string,
+  page: Page
+): Promise<boolean> => {
+  const [recentListenedPage, count] = (await searchPages(
+    {
+      dateFilters: [
+        {
+          field: 'listenedAt',
+          startDate: new Date(Date.now() - 30 * 24 * 60 * 60 * 1000),
```
jacksonh, line 29: 30 days, right?

jacksonh (approved):
Looks good. I wonder a bit about the elastic performance, but I guess its probably pretty good. Just a normal search.

packages/api/src/services/speech.ts
sywhb: Yes

PR #1132, opened 2022-08-24
Update a few colours and strings in the Android welcome and login screens

jacksonh:


android/Omnivore/app/src/main/java/app/omnivore/omnivore/ui/auth/EmailLogin.kt
```diff
@@ -46,7 +50,9 @@ fun EmailLoginView(viewModel: LoginViewModel, onAuthProviderButtonTap: () -> Uni
       }
 
       ClickableText(
-        text = AnnotatedString("Continue with Google/Apple ->"),
+        text = AnnotatedString("Return to Social Login"),
```
satindar, line 53: Makes way more sense! Was struggling to think of something

android/Omnivore/app/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,7 @@
 <resources>
     <string name="app_name">Omnivore</string>
-    <string name="gcp_id">590528454690-mpmbvcb4c3ifmnojmrultt37eo8f64at.apps.googleusercontent.com</string>
-    <string name="welcome_title">Everything you read.\nSafe, organized, and\neasy to share.</string>
+    <string name="gcp_id">590528454690-g535h6sd708andjrvrmqgaj1sergtqi9.apps.googleusercontent.com</string>
```
satindar, line 3: @jacksonh Did this one work for you?

jacksonh: oh no, let me revert that. Was just trying it out.

jacksonh: Thanks for catching, that would have been a pain to debug.

android/Omnivore/app/src/main/java/app/omnivore/omnivore/ui/auth/EmailLogin.kt
jacksonh: haha same. Technically we don't need anything since Android has a back button. Was also thinking just `Cancel`

satindar (approved):

PR #1135, opened 2022-08-25
Add debug logging to mobile auth endpoint

jacksonh:


jacksonh:
will just deploy this one then delete.

jacksonh:
ok, this isn't needed.

PR #1139, opened 2022-08-25
Fix text to speech crash, fix default voice

jacksonh:
Fix the default voice, dont force unwrap items


jacksonh:
@satindar guess we need to update testflight with this one.

satindar (approved):

PR #1141, opened 2022-08-26
Fix content-fetch timeout issue

sywhb:


jacksonh:
Ah yes, should have timeouts on everything.

I wonder if we need even lower. We will want it to be lower than the timeout CloudTask expects, otherwise CloudTask will cancel it and we wont really get an error.

sywhb:
> Ah yes, should have timeouts on everything.
> 
> I wonder if we need even lower. We will want it to be lower than the timeout CloudTask expects, otherwise CloudTask will cancel it and we wont really get an error.

Google Could Task runs for max of 10 minutes by default

jacksonh (approved):

PR #1146, opened 2022-08-28
Android Apple Login Endpoint

satindar:
Simple endpoint for an Android WebView to extract the Apple id_token. Once it's fetched we can use the same endpoints that iOS uses for a pending user and an already registered user.

### DIFF
--- a/packages/api/src/routers/auth/mobile/mobile_auth_router.ts
+++ b/packages/api/src/routers/auth/mobile/mobile_auth_router.ts
@@ -12,6 +12,8 @@ import {
 } from './sign_up'
 import { createMobileAccountCreationResponse } from './account_creation'
 import { env } from '../../../env'
+import { corsConfig } from '../../../utils/corsConfig'
+import cors from 'cors'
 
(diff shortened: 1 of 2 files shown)


packages/api/src/routers/auth/mobile/mobile_auth_router.ts
```diff
@@ -60,5 +61,12 @@ export function mobileAuthRouter() {
     res.status(payload.statusCode).json(payload.json)
   })
 
+  router.post('/android-apple-redirect', (req, res) => {
+    const { id_token } = req.body
+    return res.redirect(
+      `${env.client.url}/android-apple-token?token=${id_token as string}`
```
satindar, line 67: This url will actually never get called. Android WebView has a hook called `shouldOverrideUrlLoading` that we'll use to match this url. Then we parse out the token, dismiss then WebView, then send the apple token to the viewModel.

jacksonh (approved):
Lets make sure to test Apple login on demo

PR #1147, opened 2022-08-29
Fix issues with theme not being properly updated in iOS when the system dark/light mode changes

jacksonh:
- Make event names match in Swift and JS
- isDark is a string in Swift so we need to check for "true"
- Add script to update the bundle file


satindar (approved):

packages/appreader/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "build": "webpack --mode production"
+    "build": "webpack --mode production",
+    "build-and-update": "yarn build && cp build/bundle.js ../../apple/OmnivoreKit/Sources/Views/Resources/bundle.js"
```
satindar, line 7: nice

jacksonh: haha i swear I've written that multiple times and never remember to commit.

PR #1148, opened 2022-08-29
feature/separate queues for synthesis

sywhb:
- Add text-to-speech queue name and location to env
- Use text-to-speech queue
- Add the priority to request param and use the correct queue accordingly
GET `/:id/:outputFormat/:priority/:voice?`

### DIFF
--- a/packages/api/src/util.ts
+++ b/packages/api/src/util.ts
@@ -65,8 +65,6 @@ interface BackendEnv {
     reminderTaskHanderUrl: string
     integrationTaskHandlerUrl: string
     textToSpeechTaskHandlerUrl: string
-    textToSpeechName: string
-    textToSpeechLocation: string
   }
   fileUpload: {
     gcsUploadBucket: string
@@ -95,6 +93,9 @@ interface BackendEnv {
   azure: {
     speechKey: string
     speechRegion: string
+  }
+  gcp: {
+    location: string
   }
 }
 
@@ -150,6 +151,7 @@ const nullableEnvVars = [
   'TEXT_TO_SPEECH_TASK_HANDLER_URL',
   'AZURE_SPEECH_KEY',
   'AZURE_SPEECH_REGION',
+  'GCP_LOCATION',
 ] // Allow some vars to be null/empty
 
 /* If not in GAE and Prod/QA/Demo env (f.e. on localhost/dev env), allow following env vars to be null */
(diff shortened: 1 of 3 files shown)


packages/api/src/util.ts
```diff
@@ -233,6 +235,8 @@ export function getEnv(): BackendEnv {
     reminderTaskHanderUrl: parse('REMINDER_TASK_HANDLER_URL'),
     integrationTaskHandlerUrl: parse('INTEGRATION_TASK_HANDLER_URL'),
     textToSpeechTaskHandlerUrl: parse('TEXT_TO_SPEECH_TASK_HANDLER_URL'),
+    textToSpeechName: parse('TEXT_TO_SPEECH_QUEUE_NAME'),
```
jacksonh, line 238: I wonder if we can just hardcode these queue names. Doesn't seem like we will ever want to dynamically change the names

packages/api/src/util.ts
```diff
@@ -233,6 +235,8 @@ export function getEnv(): BackendEnv {
...
+    textToSpeechLocation: parse('TEXT_TO_SPEECH_QUEUE_LOCATION'),
```
jacksonh, line 239: I also wonder if we can get this from our current location, or if the API defaults to that. We always want all infra in the same location, so maybe we don't need to specify.

packages/api/src/utils/createTask.ts
```diff
@@ -240,10 +236,17 @@ export const enqueueParseRequest = async (
     return ''
   }
 
+  // use GCF url for low priority tasks
+  const taskHandlerUrl =
```
jacksonh, line 240: Hard to tell from the diff, would it be cleaner if content fetch also passed the queue name in?

packages/api/src/util.ts
jacksonh: I looked at what google provides and I guess not. But maybe we could have a single `GCP_LOCATION`

sywhb: Haha. Yes, I also just looked at `cloud_tasks_client.ts`...
Yeah, I agree to hardcode these configs. I was thinking about having all these configs as env var is for self-hosting.

packages/api/src/utils/createTask.ts
sywhb: Yes, let me change this

packages/api/src/util.ts
jacksonh: yeah no real need to have lots of secrets.

jacksonh (approved):

packages/api/src/util.ts
sywhb: Thanks!

PR #1149, opened 2022-08-29
perf/remove client side image layout

jacksonh:
- Make event names match in Swift and JS
- isDark is a string in Swift so we need to check for "true"
- Add script to update the bundle file
- Remove the client side image layout code


jacksonh:
Closing this one, will retest it on a fresh branch.

PR #1150, opened 2022-08-29
Android Apple token handling

satindar:
Allows for Apple sign in on Android.


jacksonh (approved):

jacksonh:
Never understood that FlowActivity naming

satindar:
> Never understood that FlowActivity naming

Ha thought that you came up with that! Guess you don't mind the change then.

PR #1152, opened 2022-08-29
Update the HTML to SSML function to use child text items instead of using the textContent attribute'

jacksonh:
- Add a new HTML to SSML function
- Use the new HTML -> SSML function

### DIFF
--- a/packages/text-to-speech/src/htmlToSsml.ts
+++ b/packages/text-to-speech/src/htmlToSsml.ts
@@ -131,16 +131,20 @@ export type SSMLItem = {
   textItems: string[]
 }
 
-export type VoiceOptions = {
-  primary: string
-  secondary: string
+export type SSMLOptions = {
+  primaryVoice: string
+  secondaryVoice: string
+  rate: string
+  language: string
 }
 
-const startSsml = (element: Element, voices: VoiceOptions): string => {
+const startSsml = (element: Element, options: SSMLOptions): string => {
   const voice =
-    element.nodeName === 'BLOCKQUOTE' ? voices.secondary : voices.primary
+    element.nodeName === 'BLOCKQUOTE'
+      ? options.secondaryVoice
+      : options.primaryVoice
   return `
-<speak xmlns="http://www.w3.org/2001/10/synthesis" xmlns:mstts="http://www.w3.org/2001/mstts" xmlns:emo="http://www.w3.org/2009/10/emotionml" version="1.0" xml:lang="en-US"><voice name="${voice}"><prosody rate="0%" pitch="0%">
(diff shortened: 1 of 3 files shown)


packages/text-to-speech/src/htmlToSsml.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export type VoiceOptions = {
+  primary: string
+  secondary: string
+}
+
+const startSsml = (element: Element, voices: VoiceOptions): string => {
+  const voice = element.nodeName === 'BLOCKQUOTE' ? voices.secondary : voices.primary
+  return `
+<speak xmlns="http://www.w3.org/2001/10/synthesis" xmlns:mstts="http://www.w3.org/2001/mstts" xmlns:emo="http://www.w3.org/2009/10/emotionml" version="1.0" xml:lang="en-US"><voice name="${voice}"><prosody rate="0%" pitch="0%">
```
sywhb, line 143: Maybe we should put rate=`1` or `default` according to the doc:
```Indicates the speaking rate of the text. You can express rate as:
A relative value, expressed as a number that acts as a multiplier of the default. For example, a value of 1 results in no change in the rate. A value of 0.5 results in a halving of the rate. A value of 3 results in a tripling of the rate.
A constant value:
x-slow
slow
medium
fast
x-fast
default```

sywhb: Maybe we should put pitch=`default` and some value:
```Indicates the baseline pitch for the text. You can express the pitch as:
An absolute value, expressed as a number followed by "Hz" (Hertz). For example, <prosody pitch="600Hz">some text</prosody>.
A relative value, expressed as a number preceded by "+" or "-" and followed by "Hz" or "st" that specifies an amount to change the pitch. For example: <prosody pitch="+80Hz">some text</prosody> or <prosody pitch="-2st">some text</prosody>. The "st" indicates the change unit is semitone, which is half of a tone (a half step) on the standard diatonic scale.
A constant value:
x-low
low
medium
high
x-high
default```

sywhb: We should maybe add language attribute to the `speak` element too

sywhb: `xml:lang="en-US"`

jacksonh: Ah thanks, I will copy over the values you had

jacksonh: I think i C&P'd this header from Azure demo app

jacksonh: ok updated these ones to pass in from config. I left pitch as hard coded.

PR #1159, opened 2022-08-30
Home Feed and WebView (Android)

satindar:
- extracy logout button to it's own comosable function
- add a home view model
- WIP: Render list of linked items


satindar:
This one is probably good to merge now. Has the basic infinite list on the home feed and we're able to render the webview by making the call to /app web routes.

Main thing missing now is PDF reader. But also missing search, labels, filters, offline mode, audio etc. 😄

jacksonh (approved):

PR #1160, opened 2022-08-30
Text to Speech Reader Shortcut

satindar:


satindar:
Need to fix the header colors and dismiss icon. I think modal is going to be the way to go here since nav view is going to be hard to wrestle into the mini player (if it would even work). 

There might also be a way to layer the mini player (in collapsed mode) back over the webview modal 🤔 

Simulator Screen Shot - iPhone 13 mini - 2022-08-30 at 15 02 17

satindar:
@jacksonh Updated this to use NSNotification instead. Turned out to be pretty easy since we already had the viewModel setup for this.

jacksonh:
Oh nice, going to try and test this out.

jacksonh:
Notification thing is really cool. I changed this to make it tap on title, and then also added a menu item. That way advanced users can use the single tap, but there's still an easy to figure out version also.

jacksonh:
Glad you figured out how to do this outside of a sheet.

PR #1161, opened 2022-08-31
Create a cloud function to steam speech mp3

sywhb:


packages/text-to-speech/src/index.ts
```diff
@@ -85,301 +56,13 @@ const updateSpeech = async (
...
-  prosodyElement.appendChild(htmlElement)
-
-  return speakElement.outerHTML.replace(/&nbsp;|\n/g, '')
-}
-
 export const textToSpeechHandler = Sentry.GCPFunction.wrapHttpFunction(
   async (req, res) => {
-    console.debug('New text to speech request', req)
+    console.info('Text to speech request received')
     const token = req.query.token as string
```
jacksonh, line 62: Could we pull this from the `Authorization` header the same way we do in API?

I think it would be nice if you can supply either way, as URL param or in `Authorization` header.

sywhb: Yes, good advice!

packages/text-to-speech/src/index.ts
```diff
@@ -388,26 +71,94 @@ export const textToSpeechHandler = Sentry.GCPFunction.wrapHttpFunction(
...
+)
+
+export const textToSpeechStreamingHandler = Sentry.GCPFunction.wrapHttpFunction(
+  async (req, res) => {
+    console.debug('Text to speech steaming request', req)
+    if (!process.env.JWT_SECRET) {
+      console.error('JWT_SECRET not exists')
+      return res.status(500).send({ errorCodes: 'JWT_SECRET_NOT_EXISTS' })
+    }
+    const token = (req.query.token || req.headers.authorization) as string
```
sywhb, line 131: Pull token from Authorization header too

jacksonh (approved):

PR #1163, opened 2022-09-02
PDF Viewer - Android

satindar:
- add a pdfviewer compose function
- WIP: trying to render pdf


satindar:
Closing this since the Compose API for PSPDFKit is not quite what we want yet. Their support dev said this way of doing it will take another 3-6 months for them to release. 

Going to try using an activity instead.

PR #1164, opened 2022-09-02
System Bar Padding - Android

satindar:


jacksonh:
Looks like there is still some color issue here with the bars: https://cdn.discordapp.com/attachments/942463305107259432/1015060730107678750/Screenshot_20220902-084806.jpg

satindar:
> Looks like there is still some color issue here with the bars: https://cdn.discordapp.com/attachments/942463305107259432/1015060730107678750/Screenshot_20220902-084806.jpg

Got the status bar fixed now. It's always black with light icons. Still need to update the feed item card styles.

jacksonh (approved):

PR #1165, opened 2022-09-02
WIP: iOS streaming audio

jacksonh:
- WIP: add CachingPlayerItem and switch to AVPlayer for streaming audio
- messing around with authed URL
- Update to use URLRequest so we can post data


satindar:
@jacksonh I pushed up a few commits. One to avoid passing in LinkedItem and instead using a struct that has the itemID and metadata needed (like you mentioned in Discord). The other one removes an explicit try call.

Overall this looks really good 👍🏽

PR #1167, opened 2022-09-06
feature/utterances

sywhb:
- Update wordsCount in page
- Change rate to 1.333
- Read Utterance input


packages/api/src/resolvers/article/index.ts
```diff
@@ -257,6 +258,9 @@ export const createArticleResolver = authorized<
 
       const saveTime = new Date()
       const slug = generateSlug(parsedContent?.title || croppedPathname)
+      const wordsCount = wordsCounter(parsedContent?.content || '', {
+        isHtml: true,
+      }).wordsCount
```
jacksonh, line 263: nice. I'm picturing in the future we could have a bunch of metadata on pages that we expose through a REST API, like get /<pageid>/metadata and it just returns key value data for things like word count.

jacksonh: I guess one worry is does this function throw on error, or just return zero? We probably don't want to completely fail if an error occurs.

packages/api/src/resolvers/article/index.ts
```diff
@@ -289,6 +293,7 @@ export const createArticleResolver = authorized<
         readingProgressAnchorIndex: 0,
         state: ArticleSavingRequestStatus.Succeeded,
         language: parsedContent?.language,
+        wordsCount,
```
jacksonh, line 296: So right now we save it but aren't exposing it in GQL, right? (That is fine with me)

packages/api/src/resolvers/article/index.ts
sywhb: I checked the source code and what it does is convert HTML to text and count words.
It is possible that the converting throws an error so I will write an exception handler to capture any error and return 0 instead.

Yeah, totally agree with the REST API for metadata!

packages/api/src/resolvers/article/index.ts
sywhb: Yes

packages/text-to-speech/src/index.ts
```diff
@@ -130,41 +149,29 @@ export const textToSpeechStreamingHandler = Sentry.GCPFunction.wrapHttpFunction(
...
-        speechMarksStream,
+        ...utteranceInput,
+        textType: 'utterance',
       }
-      res.set({
-        'Content-Type': 'audio/mpeg',
-        'Transfer-Encoding': 'chunked',
+      const { audioStream, speechMarks } = await synthesizeTextToSpeech(input)
+      res.send({
+        audioData: audioStream.read().toString('hex'),
```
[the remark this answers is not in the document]
sywhb: I will check again. Maybe creating a buffer would be better

jacksonh (approved):

jacksonh (approved):

PR #1170, opened 2022-09-07
feature/parse utterances

sywhb:
- Convert html to speech file and return in the API
- Stop counting the words in SSML tags for speech marks
- New endpoint to fetch speech-file: `/api/article/[id]/speech-file`


packages/api/src/routers/article_router.ts
```diff
@@ -93,16 +74,16 @@ export function articleRouter() {
   })
 
   router.get(
-    '/:id/:outputFormat/:priority/:voice?',
+    '/:id/:outputFormat/:priority?/:voice?/:secondaryVoice?',
```
jacksonh, line 77: I wonder if we should make some of the optional stuff query params, it might be more clear to have `/:id/:outputFormat?voice=Jenny...&secondaryVoice=Guy..&priority=high`

jacksonh (approved):

packages/api/src/routers/article_router.ts
sywhb: Yes, it would be def better

sywhb: @jacksonh Updated

jacksonh (approved):

PR #1173, opened 2022-09-07
fix/speech

sywhb:
- Exclude empty text in utterances
- Replace speech-file with speech in the route
- Add pageId to speech-file
- Change rate to string type
- Do not throw error if text is empty
- Escape entityName in ssml

### DIFF
--- a/packages/api/src/routers/article_router.ts
+++ b/packages/api/src/routers/article_router.ts
@@ -102,7 +102,7 @@ export function articleRouter() {
         },
       })
 
-      if (outputFormat === 'speech-file') {
+      if (outputFormat === 'speech') {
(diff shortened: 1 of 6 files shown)


packages/api/src/routers/article_router.ts
```diff
@@ -29,7 +29,7 @@ interface SpeechInput {
   secondaryVoice?: string
   priority?: 'low' | 'high'
 }
-const outputFormats = ['mp3', 'speech-marks', 'speech-file']
+const outputFormats = ['mp3', 'speech-marks', 'speech']
```
jacksonh, line 32: 👍

jacksonh (approved):

PR #1182, opened 2022-09-09
Fix not setting confirmation code from Google Japan confiramtion emails

sywhb:
- Match both email from and subject
- Match unicode parentheses in the subject

### DIFF
--- a/packages/inbound-email-handler/test/newsletter.test.ts
+++ b/packages/inbound-email-handler/test/newsletter.test.ts
@@ -14,17 +14,39 @@ import { MorningBrewHandler } from '../src/morning-brew-handler'
 
 describe('Confirmation email test', () => {
   describe('#isConfirmationEmail()', () => {
-    it('returns true when email is from Gmail Team', () => {
-      const from = 'Gmail Team <forwarding-noreply@google.com>'
+    let from: string
+    let subject: string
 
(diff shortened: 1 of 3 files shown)


jacksonh:
Can we change the address in the test cases? Can just use test@omnivore.app

sywhb:
> Can we change the address in the test cases? Can just use [test@omnivore.app](mailto:test@omnivore.app)

Sorry, I shouldn't have disclosed the real address

packages/inbound-email-handler/test/newsletter.test.ts
```diff
@@ -14,17 +14,39 @@ import { MorningBrewHandler } from '../src/morning-brew-handler'
...
-      const subject = `(#${code}) Gmail Forwarding Confirmation - Receive Mail from sam@omnivore.com`
+      code = '123456789'
+      subject = `(#${code}) Gmail Forwarding Confirmation - Receive Mail from test@omnivore.app`
+
+      expect(getConfirmationCode(subject)).to.equal(code)
+    })
+
+    it('returns the confirmation code from the Google Japan email', () => {
+      code = '123456789'
+      subject = `（#${code}）SWG の転送の確認 - test@omnivore.app からメールを受信`
```
sywhb, line 49: Replaced with fake address @jacksonh

jacksonh (approved):

PR #1184, opened 2022-09-12
fix/html to text error

sywhb:
- Catch Maximum call stack size exceeded error when converting html to text and fallback to linkedom
- Strip Emojis from title

### DIFF
--- a/packages/text-to-speech/src/htmlToSsml.ts
+++ b/packages/text-to-speech/src/htmlToSsml.ts
@@ -273,7 +273,7 @@ const textToUtterance = ({
   voice?: string
   isHtml?: boolean
 }): Utterance => {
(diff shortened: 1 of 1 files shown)


jacksonh:
Interesting. Is that error because of JSDOM or our parser?

sywhb:
> Interesting. Is that error because of JSDOM or our parser?

It is because of https://github.com/html-to-text/node-html-to-text/issues/44
They didn't fix this properly and seems that if the number of nested items exceeded a certain amount, it would crash

sywhb:
I think https://github.com/html-to-text/node-html-to-text uses https://github.com/fb55/htmlparser2

jacksonh (approved):

PR #1186, opened 2022-09-12
Bump up image size to 160*160

sywhb:


jacksonh (approved):

jacksonh:
Be interesting to see if this causes noticeable performance differences on the clients.

sywhb:
> Be interesting to see if this causes noticeable performance differences on the clients.

Yeah, let's see how is the performance affected on demo

PR #1193, opened 2022-09-14
Make sure we access currentViewer on the main thread

jacksonh:


jacksonh:
@satindar I'm going to merge this one so I can do a build. Let me know what you think when you have a chance to look at it.

PR #1201, opened 2022-09-15
Update UI after seeking, dont try to detect end of stream using currentTime

jacksonh:
This helps with two issues:
- after seeking sometimes you will see the loading spinner even
after audio has resumed.

- sometimes if we are still pulling data we might detect that
we are at the end of the stream.

### DIFF
--- a/apple/OmnivoreKit/Sources/Views/FeedItem/HomeFeedCardView.swift
+++ b/apple/OmnivoreKit/Sources/Views/FeedItem/HomeFeedCardView.swift
@@ -77,7 +77,7 @@ public struct FeedCard: View {
         .padding(.top, 8)
       }
     }
-    .padding(.top, 10)
+    .padding(.top, 0)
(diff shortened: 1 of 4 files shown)


apple/OmnivoreKit/Sources/Views/FeedItem/HomeFeedCardView.swift
```diff
@@ -77,7 +77,7 @@ public struct FeedCard: View {
         .padding(.top, 8)
       }
     }
-    .padding(.top, 10)
```
jacksonh, line 80: @satindar from what I can tell, SwiftUI is now adding the spacing before the item also, so we don't need this padding. Either that or its padding from the cell edge insets

jacksonh:
Doing a TestFlight build off this branch.

jacksonh:
Going to merge these iOS patches and create a new build. Will also adjust the padding a bit more on feed items.

PR #1207, opened 2022-09-18
On Safari PDF Highlights are not setting opacity

jacksonh:


jacksonh:
deploying this to demo

PR #1209, opened 2022-09-18
Android PDF Reader

satindar:
- add pspdfkit android dep
- use pspdfkit documentloader when selecting a pdf item
- add pdf reader composable


satindar:
Currently hitting this exception when tapping on a pdf item from the home view list:

```
2022-09-18 08:33:30.289 22597-22597/app.omnivore.omnivore E/AndroidRuntime: FATAL EXCEPTION: main
    Process: app.omnivore.omnivore, PID: 22597
    com.pspdfkit.jetpack.compose.NonFragmentActivityException: You need to use a fragment activity in order to use the DocumentView composable.
        at com.pspdfkit.jetpack.compose.DocumentViewKt.DocumentView(SourceFile:259)
        ```

Issue #1197, opened 2022-09-15
There is a strange border appearing around "Read Later" button when the button is clicked


sywhb:
This happens on Web.
<img width="477" alt="Screenshot 2022-09-15 at 9 04 43 AM" src="https://user-images.githubusercontent.com/12963965/190306288-0f5d382d-8954-4263-ae66-a5c3d3c43a6d.png">

<img width="488" alt="Screenshot 2022-09-15 at 11 21 46 AM" src="https://user-images.githubusercontent.com/12963965/190306793-3d3372c2-d881-4703-a90e-d2798f55171a.png">

jacksonh:
Ah yes, this is some of the default browser stuff. I will fix this one.

jacksonh:
PR here: https://github.com/omnivore-app/omnivore/pull/1210

PR #1210, opened 2022-09-19
Only use button active item focus in modal buttons

jacksonh:
This is useful for keyboard tabbing but we don't want it on
all items as it creates a weird effect.

Fixes: https://github.com/omnivore-app/omnivore/issues/1197


jacksonh:
I'll merge this one to test on demo.

PR #1213, opened 2022-09-19
Generate a queue of player items when speech synthesis starts

jacksonh:
This should better handle cases where playback catches up to
audio synthesis/download, and will have the player items put
into a wait state.

This does make seeking a little less performant, as AVPlayer
wont aggresively pre-load as we were before.


jacksonh:
Will merge this one now. Seems to have resolved issues of playback getting stuck

PR #1216, opened 2022-09-19
Bundled Web Reader - Android

satindar:
- add ArticleContent class and stub in webreader
- move LinkedItem into models package
- add web assets to android
- load bundled resources into web reader


satindar:
Running into a few runtime errors right now. Hard to debug since it's compiled TS. 

Screen Shot 2022-09-19 at 2 27 42 PM

satindar:
Resolved the issue above (c.map is not a function). 

Main issue now is that it's prepending the Android asset path to the ima urls 🤔

jacksonh (approved):

satindar:
Merging this one before starting on context menus.

PR #1222, opened 2022-09-21
WebView Text Selection Menu - Android

satindar:


jacksonh:
Interesting, so Android doesn't let us hook into the default menu system quite like iOS does?

satindar:
> Interesting, so Android doesn't let us hook into the default menu system quite like iOS does?

The ActionMode way I used is the preferred way according to some Android dev advocates. There is also this API which seems to be for adding to an existing menu:

https://developer.android.com/reference/android/view/Menu#add(int,%20int,%20int,%20int)

Haven't been able to make that one work yet though.

jacksonh (approved):
Nice work on this one. Really tricky to get all these languages working together.

PR #1227, opened 2022-09-22
fix/axios newsletter parsing

sywhb:
- Add test case for Axios AM newsletter
- Add Axios Chicago test case
- Remove the table with text content less than 20 because these are ads


packages/api/src/utils/parser.ts
```diff
@@ -262,7 +262,6 @@ export const parsePreparedContent = async (
       const codeBlocks = article.dom.querySelectorAll('code')
       if (codeBlocks.length > 0) {
         codeBlocks.forEach((e) => {
-          console.log(e.textContent)
```
jacksonh, line 265: 👍

PR #1230, opened 2022-09-23
Escape HTML entities in Youtube and Twitter handlers

sywhb:
- Escape youtube title and author name
- Escape HTML entities in Twitter title and description
- Escape HTML entities in puppeteer-parse

### DIFF
--- a/packages/content-fetch/youtube-handler.js
+++ b/packages/content-fetch/youtube-handler.js
@@ -37,6 +37,7 @@ exports.youtubeHandler = {
 
     const oembedUrl = `https://www.youtube.com/oembed?format=json&url=` + encodeURIComponent(`https://www.youtube.com/watch?v=${videoId}`)
     const oembed = (await axios.get(oembedUrl.toString())).data;
+    // escape html entities in title
     const title = _.escape(oembed.title);
     const ratio = oembed.width / oembed.height;
     const thumbnail = oembed.thumbnail_url;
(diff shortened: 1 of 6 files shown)


packages/content-fetch/youtube-handler.js
```diff
@@ -36,11 +37,13 @@ exports.youtubeHandler = {
 
     const oembedUrl = `https://www.youtube.com/oembed?format=json&url=` + encodeURIComponent(`https://www.youtube.com/watch?v=${videoId}`)
     const oembed = (await axios.get(oembedUrl.toString())).data;
-    const title = oembed.title;
+    // escape html entities in title
+    const title = _.escape(oembed.title);
```
jacksonh, line 41: So if the input was `This "is quoted"` would the output here be `This \"is quoted\"`?

sywhb:
Screenshot 2022-09-21 at 1 37 52 PM

packages/content-fetch/youtube-handler.js
jacksonh: Oh and then we unescape when we set the title in the database?

sywhb:
Screenshot 2022-09-23 at 4 50 13 PM

jacksonh:
Awesome, thank you for verifying. Lots of encoding/decoding going on.

packages/content-fetch/youtube-handler.js
sywhb: Yes, it will become `This \"is quoted\"` in the database

jacksonh (approved):

PR #1238, opened 2022-09-26
feature/split large utterance

sywhb:
- Make each `<li>` an utterance to shorter the utterance length and add a pause after the utterance ends
- Start at index 3, because the frontend starts two nodes above the backend
- Split big SSML into small chunks of less than 200 chars

### DIFF
--- a/packages/text-to-speech/src/htmlToSsml.ts
+++ b/packages/text-to-speech/src/htmlToSsml.ts
@@ -55,8 +55,7 @@ const ANCHOR_ELEMENTS_BLOCKED_ATTRIBUTES = [
 
 function ssmlTagsForTopLevelElement() {
   return {
-    opening: `<p>`,
-    closing: `</p>`,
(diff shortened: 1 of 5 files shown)


packages/text-to-speech/src/htmlToSsml.ts
```diff
@@ -106,9 +104,9 @@ function parseDomTree(pageNode: Element) {
 
   visitedNodeList.shift()
   visitedNodeList.forEach((node, index) => {
-    // We start at index 2, because the frontend starts one node above us
-    // on the #readability-content element that wraps the entire content.
-    node.setAttribute('data-omnivore-anchor-idx', (index + 2).toString())
+    // We start at index 3, because the frontend starts two nodes above us
```
jacksonh, line 107: hah thanks

jacksonh (approved):
Seems like best thing to do is merge and test as much as possible on demo.

PR #1246, opened 2022-09-27
feature/subscription icon

sywhb:
- Add icon field to subscriptions table
- Add icon to the subscription entity
- Save icon in the subscription table
- Use parsed image as subscription icon
- Return subscription icon proxy url in the list subscriptions api response


jacksonh (approved):

jacksonh:
So would this save something like the feature image for an article? I think what we need is something like this image https://substackcdn.com/image/fetch/w_1272,c_limit,f_webp,q_auto:best,fl_progressive:steep/https%3A%2F%2Fbucketeer-e05bbc84-baa3-437e-9518-adb32be77984.s3.amazonaws.com%2Fpublic%2Fimages%2F1dad529e-b61e-4313-85c7-3c978c3d094f_768x768.png from the https://slowboring.com blog. Its sort of like favicon for the site, not the article.

sywhb:
> So would this save something like the feature image for an article? I think what we need is something like this image https://substackcdn.com/image/fetch/w_1272,c_limit,f_webp,q_auto:best,fl_progressive:steep/https%3A%2F%2Fbucketeer-e05bbc84-baa3-437e-9518-adb32be77984.s3.amazonaws.com%2Fpublic%2Fimages%2F1dad529e-b61e-4313-85c7-3c978c3d094f_768x768.png from the https://slowboring.com blog. Its sort of like favicon for the site, not the article.

Yes, you are right! I have been working on using the favicon or a logo hardcoded in the email content

sywhb:
@jacksonh Actually I found that they don't have a favicon in the HTML. Maybe it's because of the email content.

sywhb:
I added some test pages which I get from the email content and none of them has a favicon.
Maybe I should try to retrieve the icon of each newsletter in the handler if there is any

jacksonh (approved):
NIce work. Lots of work for some small icons but it will make our UI much easier to scan.

packages/api/src/utils/parser.ts
```diff
@@ -621,3 +621,18 @@ export const parseEmailAddress = (from: string): addressparser.EmailAddress => {
   }
   return { name: '', address: from }
 }
+
+export const fetchFavicon = async (
+  url: string
+): Promise<string | undefined> => {
+  try {
+    // get the correct url if it's a redirect
+    const response = await axios.head(url)
```
jacksonh, line 630: Do we know what the default axios timeouts are? Might want to make sure its kind of short for these.

packages/api/src/utils/parser.ts
```diff
@@ -621,3 +621,18 @@ export const parseEmailAddress = (from: string): addressparser.EmailAddress => {
...
+    const realUrl = response.request.res.responseUrl
+    const domain = new URL(realUrl).hostname
+    return `https://api.faviconkit.com/${domain}/32`
```
jacksonh, line 633: NIce. I think they might have rate limiting but lets see. Can also sign up for their service hopefully.

jacksonh (approved):

packages/api/src/utils/parser.ts
sywhb: The default is `0`. I better put a `timeout: 1000` (1s) here

packages/api/src/utils/parser.ts
sywhb: Yes and if it doesn't work well, we can always fetch the HTML content from the domain website and parse the `metadata` from the HTML

packages/api/src/utils/parser.ts
sywhb: Maybe `5s` is better if it is slow

PR #1257, opened 2022-09-29
Consolidate newsletters and other handlers

sywhb:
- Init newsletter-handler
- Replace newsletter-handler with content-handler
- Add handlers for content-fetching


jacksonh:
nice

jacksonh (approved):

PR #1259, opened 2022-09-29
Set the "since" to be the exact time when the last page was added to ensure all the deletion happen after that

sywhb:


jacksonh:
Classic.

jacksonh (approved):

PR #1263, opened 2022-09-30
Automation of appstore screenshots

jacksonh:
- This doesn't completely automate the process but makes it
easier to create screenshots across devices.


jacksonh:
@satindar this sort of works. I'll put together some docs on how I run it, its more like a script for the user to follow than an automation.

PR #1268, opened 2022-10-02
Add the getting started guide as markdown

jacksonh:


jacksonh:
Need to cleanup this markdown, then create an HTML export.

PR #1270, opened 2022-10-03
Split utterance into chunks of 256 chars

sywhb:


jacksonh (approved):
Probably best thing is to deploy and we can test.

sywhb:
> Probably best thing is to deploy and we can test.

Deployed in Demo @jacksonh

Issue #1265, opened 2022-09-30
Apple Login button frequently isn't available


jacksonh:
Seems like the URL we fetch the icon from: https://appleid.cdn-apple.com/appleid/button?color=black&width=261&height=40&type=continue often doesn't work. We should just import this asset into our images instead of relying on Apple's version.

jacksonh:
![appleid_button@1x (1)](https://user-images.githubusercontent.com/75189/193264643-156a1f82-d2ae-49e5-9ea3-3340d351fdd8.png)

jacksonh:
There's a tool here to make downloadable buttons but its failing right now: https://appleid.apple.com/signinwithapple/button

jacksonh:
![appleid_button@2x](https://user-images.githubusercontent.com/75189/193727374-12a671df-3625-4341-93e3-e87a8c95f688.png)

jacksonh:
Fixed

PR #1273, opened 2022-10-04
Embed the apple id button so it doesnt fail when the apple CDN returns 404

jacksonh:


jacksonh:
We lose some localization by doing it this way, but that is better than the button disappearing: https://github.com/omnivore-app/omnivore/issues/1265

PR #1276, opened 2022-10-04
improve utterance break

sywhb:
- Fix tests
- if we hit 256, look back for first ending sentence within 80 chars


jacksonh:
I AM EXCITE

jacksonh (approved):
Probably want a few more tests for edge cases with this, but looks good.

PR #1279, opened 2022-10-05
Use NLP lib to better detect sentence boundary and split utterances

sywhb:


jacksonh (approved):
"Our new AI powered backend detects key phrases"

PR #1284, opened 2022-10-06
Exposes port 9090:8080 in the container

sywhb:


docker-compose.yml
```diff
@@ -140,8 +140,8 @@ services:
       context: .
       dockerfile: ./packages/content-fetch/Dockerfile-local
```
sywhb, line 141: @jacksonh why are we using a `Dockerfile-local` here? Is it something different from the `Dockerfile`?

jacksonh: @sywhb i think there was some issue running Dockerfile locally with docker-compose vs Dockerfile which was running in App Engine. 

Probably an issue with the Chromium dependency

sywhb: I didn't have such an issue. Maybe it is good to remove the `Dockerfile-local` now?

jacksonh: @sywhb yeah probably ok now. Maybe need to do a full test with --force-recreate also

jacksonh (approved):

docker-compose.yml
sywhb: Yeah, I tested

PR #1285, opened 2022-10-06
use content handler in api

sywhb:
- Remove custom content-handler in packages/api
- Calls preParseContent function in content-handler package before parsing content


jacksonh (approved):

jacksonh:
I suppose we should hold off on this one in prod until monday, but could get it all merged in demo

PR #1294, opened 2022-10-10
Recent searches feature

sywhb:


sywhb:
> Nice timing, just improving this on the client

Lol. Yeah, I saw that

packages/api/src/entity/search_history.ts
```diff
@@ -0,0 +1,29 @@
...
+export class SearchHistory {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string
+
+  @ManyToOne(() => User, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'user_id' })
+  user!: User
+
+  @Column('varchar', { length: 255 })
+  keyword!: string
```
jacksonh, line 22: Can we make this "term" ?

jacksonh (changes requested):

packages/api/src/entity/search_history.ts
sywhb: Yes 👍 that is more accurate

sywhb:
Hey @jacksonh this is ready for review

packages/api/src/entity/search_history.ts
sywhb: Updated

jacksonh (approved):

PR #1296, opened 2022-10-10
Format Typescript Files

zfir:
Hacktoberfest PR


satindar:
@zfir Hi thanks for taking a look at the formatting of TS files here!

We have config files and scripts in place that use a combo of prettier and ESLint. It looks like you ran a script at the root level but it's likely different than the config we have so things are going to change back once someone else runs the scripts.

PR #1298, opened 2022-10-11
feature/add popular read when signup

sywhb:
- Add popular reads to new users
- Add iOS popular reads to new iOS users
- Set device of the request in Redis
- Device info expires in 10 mins


sywhb:
Renamed `X-Device` to `X-OmnivoreClient`

packages/api/src/services/popular_reads.ts
```diff
@@ -88,6 +94,37 @@ export const addPopularRead = async (
...
+export const addPopularReadsForNewUser = async (
+  userId: string,
+  isIOSUser = false
+): Promise<void> => {
+  await addPopularReads(
+    userId,
+    'omnivore_get_started',
+    'power_read_it_later',
+    'omnivore_organize',
+    isIOSUser ? 'omnivore_ios' : 'omnivore_android'
```
jacksonh, line 124: I think this is wrong, because the user could be web. What we might want to do is replace isIOSUser with `client` and then have something like:

```
defaultReads = ['omnivore_get_started', 'power_read_it_later', 'omnivore_organize']

var reads = defaultReads

switch(client) {
  case iOS:
     reads += 'omnivore_ios'
}
```

One other thing we could consider is language too, but that might be too complicated for now.

sywhb: This is just a `TODO` because I haven't got the starting guide for iOS user so I just put rubbish here :)

packages/api/src/events/user/user_created.ts
```diff
@@ -114,3 +115,18 @@ export class PublishNewUserEvent implements EntitySubscriberInterface<Profile> {
...
+export class AddPopularReadsToNewUser
+  implements EntitySubscriberInterface<Profile>
+{
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+  listenTo() {
+    return Profile
+  }
+
+  async afterInsert(event: InsertEvent<Profile>): Promise<void> {
+    const isIOSUser = event.entity.user.source === 'APPLE'
```
jacksonh, line 129: Discussion in Discord about this, but we will need to somehow get this from request header information here instead of from the auth provider source.

sywhb:
Hey @jacksonh I saved the `X-OmnivoreClient` in the request context and read it when adding popular reads

packages/api/package.json
```diff
@@ -53,6 +53,7 @@
     "dot-case": "^3.0.4",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
+    "express-http-context": "^1.2.4",
```
jacksonh, line 56: oh cool, i was wondering about something like this.

packages/api/src/server.ts
```diff
@@ -104,6 +105,16 @@ export const createApp = (): {
     app.use('/api/', apiLimiter)
   }
 
+  // set client info in the request context
+  app.use(httpContext.middleware)
+  app.use('/api/', (req, res, next) => {
```
jacksonh, line 110: nice, this will help a lot. Could be really useful with some analytics stuff.

packages/api/src/services/popular_reads.ts
```diff
@@ -88,6 +95,51 @@ export const addPopularRead = async (
...
+    const pageId = await addPopularRead(userId, name)
+    results.push({
+      pageId,
+      name,
+      status: pageId
+        ? ArticleSavingRequestStatus.Succeeded
+        : ArticleSavingRequestStatus.Failed,
+    })
+  }
+  return results
```
jacksonh, line 113: nice, could `addPopularRead` throw? I guess if one fails all probably will anyways.

packages/api/src/services/popular_reads.ts
```diff
@@ -88,6 +95,51 @@ export const addPopularRead = async (
...
+  const defaultReads = [
+    'omnivore_get_started',
+    'power_read_it_later',
+    'omnivore_organize',
+  ]
+
+  // get client from request context
+  const client = httpContext.get('client') as string | undefined
+
+  switch (client) {
```
jacksonh, line 128: For now, because we don't have these guides, these will just fail to add, but nothing bad will happen I think.

jacksonh:
Only thing I want to be careful with is the `_web`, `_ios`, `_android` docs. Since those haven't been created yet, we need to make sure we don't accidentally deploy something that adds temp content to new user's libraries.

packages/api/src/services/popular_reads.ts
sywhb: We handler the exception when creating pages in elastic so it should not throw any error but the `pageId` could be `undefined` if the `name` is not found in the array of `popularReads`, so it will add a failed result in the array of results

packages/api/src/services/popular_reads.ts
sywhb: Yes, we will simply ignore these nonexistent guide

sywhb:
> 

Yes, we have some tests to ensure that

jacksonh (approved):
I'll create a new PR to update the article too.

PR #1300, opened 2022-10-11
Web Prefs - Android

satindar:
- bump android version
- change navbar offset on scroll


jacksonh (approved):
:nice:

satindar:
@jacksonh Update on this one: Main task left is to store and apply the selected font. I'll push a beta build after that.

I'll open an issue with a checklist of tasks we need to complete before we can push this to a prod Play Store build. Some things like pdf view thumbnails and highlight saving, highlight menu on webview, and then all the assets/copy needed for the store.

PR #1301, opened 2022-10-12
Improvements to substack article handling

jacksonh:
- Dont exclude substacks in article headers
- Remove the subscriber dialog that was part of the header before


sywhb (approved):

jacksonh:
Hopefully no tests break, adding a new one too

PR #1308, opened 2022-10-13
Create a content handler for GitHub pages

jacksonh:
This will search for the markdown <article> element and make it
the only page element so we add a repo's README.


sywhb:
Hah, seems the `GithubHandler` class is not included in this PR

jacksonh:
oops. Classic move for me

jacksonh:
@sywhb Checked in now

sywhb (approved):
@jacksonh LGTM

PR #1313, opened 2022-10-16
Mac App Read Later Fix

satindar:


jacksonh (approved):
Solid old school Mac development here

PR #1314, opened 2022-10-17
Empty state for the Highlights View

jacksonh:


jacksonh:
@satindar can you take a look at this one?

PR #1334, opened 2022-10-19
Add the iOS Saving Links article as a default read for new Omnivore users on iOS

jacksonh:
This also re-orders things a bit, making sure the getting started
guide is always the topmost in a user's library.

### DIFF
--- a/packages/api/src/services/popular_reads.ts
+++ b/packages/api/src/services/popular_reads.ts
@@ -124,7 +124,6 @@ export const addPopularReadsForNewUser = async (
   const defaultReads = [
     'omnivore_organize',
     'power_read_it_later',
-    'omnivore_get_started',
   ]
 
   // get client from request context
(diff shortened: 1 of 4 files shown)


packages/api/src/services/popular_reads.ts
```diff
@@ -142,6 +141,9 @@ export const addPopularReadsForNewUser = async (
       break
   }
 
+  // We always want this to be the top-most article in the user's
+  // list. So we save it last to have the greatest saved_at
+  defaultReads.push('omnivore_get_started')
```
sywhb, line 146: Nice

sywhb (approved):
LGTM 👍

sywhb:
@jacksonh We also need to update the test because now the ios guide is available

jacksonh:
eek tests! Thanks, updating now.

PR #1336, opened 2022-10-20
Delete highlights, labels and other metadata when soft delete pages

sywhb:


packages/api/src/resolvers/article/index.ts
```diff
@@ -643,10 +649,16 @@ export const setBookmarkArticleResolver = authorized<
         return { errorCodes: [SetBookmarkArticleErrorCode.NotFound] }
       }
 
-      // delete the page
+      // delete the page and its metadata
       const deleted = await updatePage(
         pageRemoved.id,
-        { state: ArticleSavingRequestStatus.Deleted },
+        {
```
jacksonh, line 655: Do we also clear content and originalHTML when we delete?

sywhb: We do not clear them. hmm, my concern is we could use them for some analytics and stuff @jacksonh

sywhb: I'm open to clear them too since we are not using them now

packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
```diff
@@ -975,7 +989,9 @@ function HomeFeedGrid(props: HomeFeedContentProps): JSX.Element {
       )}
       {showRemoveLinkConfirmation && (
         <ConfirmationModal
-          message={'Are you sure you want to remove this link?'}
+          message={
+            'Are you sure you want to remove this link? All your highlights will be deleted.'
```
sywhb, line 993: @jacksonh I mention the highlights deleted in the confirmation modal. The wording could be changed

jacksonh: How about:

```
Are you sure you want to remove this item? All associated notes and highlights will be deleted.'
```

sywhb: Better!

jacksonh: I'll make it the same on iOS

sywhb: Okay. Updated

jacksonh (approved):

PR #1337, opened 2022-10-20
Fix alert/row flicker when removing items on iOS

jacksonh:
This also updates the message to be consistent with Web.


sywhb (approved):

jacksonh:
@satindar I'm merging this one so i can do a build, but do you mind taking a look? Main change was moving the alert to the list, not the item.

satindar:
> @satindar I'm merging this one so i can do a build, but do you mind taking a look? Main change was moving the alert to the list, not the item.

Yeah this is a lot better now. Didn't really notice how bad it was before. Nice catch on that!

PR #1338, opened 2022-10-20
Split the expanded player into a new view and use a sheet to present it

jacksonh:


satindar:
Works pretty well on iPhone. I'll check iPad too.

PR #1348, opened 2022-10-25
Add new action bar at bottom of iOS reader that shows when user reaches end

jacksonh:


jacksonh:
@satindar can you take a look at this one? More ZStack shenanigans.

satindar:
> @satindar can you take a look at this one? More ZStack shenanigans.

Looks good! I added an opacity animation that happens when it's presented. Not sure how to make it fade out though.

Also added the button color to the asset catalog.

PR #1353, opened 2022-10-27
Persist integration in db before creating cloud task to sync with integrations

sywhb:


jacksonh (approved):
Looks good, hopefully we have UPDATE rights on that table, not just INSERT

sywhb:
> Looks good, hopefully we have UPDATE rights on that table, not just INSERT

Yeah, we have both

PR #1358, opened 2022-10-28
WIP: New themes and reader preferences for iOS

jacksonh:


apple/OmnivoreKit/Sources/Views/Theme.swift
```diff
@@ -0,0 +1,54 @@
...
+      if theme.rawValue == themeName {
+        return theme
+      }
+    }
+    return nil
+  }
+}
+
+public enum ThemeManager {
+  @AppStorage(UserDefaultKey.themeName.rawValue) public static var currentThemeName = "System"
```
satindar, line 53: This is neat! Had no idea you could do this.

PR #1359, opened 2022-10-28
Rules API

sywhb:


packages/api/src/schema.ts
```diff
@@ -1912,6 +1912,93 @@ const schema = gql`
...
+
+  type Rule {
+    id: ID!
+    name: String
+    query: String!
+    actions: [RuleAction!]!
+    triggerType: RuleTriggerType!
+    createdAt: Date!
+    updatedAt: Date!
+  }
```
sywhb, line 1929: @jacksonh `query` is the search filter and `actions` referred to a list of "actions"

packages/api/src/schema.ts
```diff
@@ -1912,6 +1912,93 @@ const schema = gql`
...
+    type: RuleActionType!
+    value: String
+  }
+
+  enum RuleActionType {
+    ADD_LABEL
+    ARCHIVE
+    MARK_AS_READ
+    SEND_NOTIFICATION
+  }
```
sywhb, line 1941: @jacksonh Action type is an enum. For starters, I added these four

packages/api/src/schema.ts
```diff
@@ -1912,6 +1912,93 @@ const schema = gql`
...
+
+  enum RuleTriggerType {
+    ON_PAGE_UPDATED
+    CRON
+  }
```
sywhb, line 1946: @jacksonh Rule trigger type is an enum value of either `ON_PAGE_UPDATED` or `CRON`.
I use `ON_PAGE_UPDATED` not `ON_PAGE_CREATED` because there is no content of the page when it is created before parsing

packages/api/src/schema.ts
```diff
@@ -1912,6 +1912,93 @@ const schema = gql`
...
+    BAD_REQUEST
+  }
+
+  input SetRuleInput {
+    id: ID
+    name: String
+    query: String!
+    actions: [RuleActionInput!]!
+    triggerType: RuleTriggerType!
+  }
```
sywhb, line 1963: @jacksonh `setRule` API can be used to create or update rule based on the existence of `id` field

packages/api/src/schema.ts
jacksonh: For RuleTriggerType, would the client have to set that when creating a rule?

sywhb: Probably no. The `RuleTriggerType` should bind to the `action` so I probably should have a mapping in the backend and remove this from the GQL request

jacksonh: Yeah I guess we'd have a list and it would be part of it.

jacksonh (approved):

PR #1377, opened 2022-11-01
Fix not correctly set confirmation code from Gmail in Spanish

sywhb:
* Forward text if html is not found in the email
* Fix not correctly set confirmation code from Gmail in Spanish

### DIFF
--- a/packages/inbound-email-handler/src/newsletter.ts
+++ b/packages/inbound-email-handler/src/newsletter.ts
@@ -10,7 +10,7 @@ interface Unsubscribe {
 const EMAIL_CONFIRMATION_CODE_RECEIVED_TOPIC = 'emailConfirmationCodeReceived'
 const CONFIRMATION_EMAIL_SENDER_ADDRESS = 'forwarding-noreply@google.com'
 // check unicode parentheses too
-const CONFIRMATION_CODE_PATTERN = /^[(（]#\d+[)）]/u
+const CONFIRMATION_CODE_PATTERN = /\d+/u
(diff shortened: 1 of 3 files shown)


packages/inbound-email-handler/src/newsletter.ts
```diff
@@ -55,7 +55,7 @@ export const getConfirmationCode = (subject: string): string | undefined => {
   if (matches) {
     // get the number code only
     // e.g. (#123456) => 123456
-    return matches[0].slice(2, -1)
```
jacksonh, line 58: I wonder if what we should do in the UX is just show the whole string. We'd have to change the UI a bit but the REGEX's and slicing seems a little brittle.

jacksonh (approved):

packages/inbound-email-handler/src/newsletter.ts
sywhb: Yeah, I was thinking about it too.
One idea is we can always forward every confirmation email with an instruction or a link for setting the confirmation code in Omnivore
@jacksonh

jacksonh: Yeah, maybe creating an email with the info in it works best. The whole user interface is maybe just too confusing and brittle, but we could make a nice email.

jacksonh (approved):

packages/inbound-email-handler/src/newsletter.ts
sywhb: I will create a ticket for this

PR #1380, opened 2022-11-01
Improve note saving reliability on iOS

jacksonh:
- WIP: New themes and reader preferences for iOS
- Better handling of errors when dispatching annotation events to JS
- Update appreader bundle
- Add missing import
- Revert some theme changes, update reader, fix theme clearing
- Better handling of errors when dispatching annotation events to JS
- Revert some theme stuff


apple/OmnivoreKit/Sources/App/Views/WebReader/WebReader.swift
```diff
@@ -84,11 +85,20 @@ struct WebReader: PlatformViewRepresentable {
   private func updatePlatformView(_ webView: WKWebView, context: Context) {
     if annotationSaveTransactionID != context.coordinator.lastSavedAnnotationID {
       context.coordinator.lastSavedAnnotationID = annotationSaveTransactionID
-      (webView as? OmnivoreWebView)?.dispatchEvent(.saveAnnotation(annotation: annotation))
+      do {
```
jacksonh, line 88: @satindar this is the main change here, we let the WebReader control either the modal is closed and only close it if saving succeeds.

apple/OmnivoreKit/Sources/Services/AudioSession/SpeechSynthesizer.swift
```diff
@@ -120,7 +120,7 @@ struct SpeechSynthesizer {
   func createPlayerItems(from: Int) -> [SpeechItem] {
     var result: [SpeechItem] = []
 
-    for idx in from ..< document.utterances.count {
+    for idx in from ..< min(7, document.utterances.count) {
```
jacksonh, line 123: this was in here for some debuging 😬

apple/OmnivoreKit/Sources/Views/Article/OmnivoreWebView.swift
```diff
@@ -31,42 +32,62 @@ public final class OmnivoreWebView: WKWebView {
...
-      }
-    })
   }
 
   @available(*, unavailable)
   required init?(coder _: NSCoder) {
     fatalError("init(coder:) has not been implemented")
   }
 
+  public func updateTheme() {
```
jacksonh, line 42: Most stuff from here down is just to propagate errors when dispatching events.

apple/OmnivoreKit/Sources/Views/Article/OmnivoreWebView.swift
```diff
@@ -340,25 +409,35 @@ public enum WebViewDispatchEvent {
...
+      case let .updateTheme(themeName: themeName):
+        return "event.themeName = '\(themeName)';"
+      case let .updateFontSize(size: size):
+        return "event.fontSize = '\(size)';"
+      case let .updateColorMode(isDark: isDark):
+        return "event.isDark = '\(isDark)';"
+      case let .updateFontFamily(family: family):
+        return "event.fontFamily = '\(family)';"
+      case let .saveAnnotation(annotation: annotation):
+        let encoder = JSONEncoder()
```
jacksonh, line 429: This is the only real change here. Without this, any note with JSON symbols like `"` would fail to save.

apple/OmnivoreKit/Sources/Views/FontSizeAdjustmentPopoverView.swift
```diff
@@ -69,81 +69,112 @@ public enum WebFont: String, CaseIterable {
       .navigationTitle("Reader Font")
     }
 
+    var themePicker: some View {
```
jacksonh, line 72: This part is not actually used yet.

jacksonh:
I built this on top of my theme PR at first because it also touched the appreader, but all the theme changes are commented out here.

apple/OmnivoreKit/Sources/App/Views/WebReader/WebReader.swift
```diff
@@ -84,11 +85,19 @@ struct WebReader: PlatformViewRepresentable {
...
+        try (webView as? OmnivoreWebView)?.dispatchEvent(.saveAnnotation(annotation: annotation))
```
satindar, line 89: Does the dispatchEvent call sometime fail?

jacksonh: @satindar I think technically it can throw from JS. I haven't been able to force it, but we were kind of silently failing before.

jacksonh: Well, I did make it so generating the event can throw if we can't serialize the event data.

satindar: Nice. Should be more robust now 👍🏽

PR #1389, opened 2022-11-02
Add a realistic voice api provider

sywhb:


sywhb:
Hey @jacksonh, this is a starting point for getting and setting character count in Redis for rate limiting which doesn't take care of concurrency yet

packages/api/src/services/features.ts
```diff
@@ -0,0 +1,67 @@
...
+    return feature
+  }
+
+  // opt in to feature for the first 1000 users
+  const count = await getRepository(Feature).countBy({
+    name: FeatureName.UltraRealisticVoice,
+    grantedAt: Not(IsNull()),
+  })
+
+  let grantedAt: Date | null = new Date()
```
jacksonh, line 43: Technically there could be a small race condition where more than 1000 could be created, but only a few so I don't think it matters in this case.

packages/api/src/services/features.ts
```diff
@@ -0,0 +1,67 @@
...
+    user: { id: uid },
+    name: FeatureName.UltraRealisticVoice,
+    grantedAt,
+  })
+}
+
+export const signFeatureToken = (feature: Feature): string => {
+  return jwt.sign(
+    {
+      userid: feature.user.id,
```
jacksonh, line 59: Not sure if we need to copy but we use `uid` for a regular token.

Also, I'm not sure if the JWT library handles expiration itself if we use these names:

```
  const iat = Math.floor(Date.now() / 1000)
  const exp = Math.floor(
    new Date(Date.now() + 1000 * 60 * 60 * 24 * expireInDays).getTime() / 1000
  )
  return jwt.sign({ uid: userId, iat, exp }, env.server.jwtSecret)
```

sywhb: Yeah, the JWT library handles the `iat` and `exp` itself and I have updated the token to be expired in 1 day.
Do you think it is enough?

sywhb: Also renamed `userid` to `uid` in the JWT token

packages/api/src/services/features.ts
sywhb: Yeah, the race condition could happen. I could probably do a `insert` and `select` in a transaction

jacksonh (approved):

PR #1394, opened 2022-11-06
Feat/dragndrop

rupin90:
This is still in progress

 - [x]  user can drop any file on the library page.  (I do have temporary code that shows the name of the file being dropped right on the page)
 - [x] File upload APIs 
 - [x] Dragndrop UI
 - [x] File upload progressbar


jacksonh:
Hey @rupin90 could we also add this to `HomeFeedContainer.tsx` so we can use it on the old library as-well as the new?

rupin90:
> Hey @rupin90 could we also add this to `HomeFeedContainer.tsx` so we can use it on the old library as-well as the new?

Yep yep, can do that today and will update this.

What about the file upload? API?

rupin90:
> Is the plan with this feature to parse the content of the PDF document? If that's the case; as far as I know, parsing PDF is going to be [PIA](https://wiki.c2.com/?PdfSucks=).
> 
> There is going to be a need for a decent amount of tests with different types of PDFs and how to handle them. For example, some PDF files are not parsable without OCR.
> 
> > What are the file formats that we want to accept? Is it just Pdf?
> 
> I would at least add EPUB to the list. Maybe also text files (.txt) and Markdown files (.md), and/or latex
> 
> > Are we accepting multiple uploads at once?
> 
> A queue like system will be nice to have, but can be a separate feature. Also, adding the ability to send files through the API would be a nice addition.

I think AFAIK it is going to be simple file upload. For now we are just doing one upload at a time.

Thank you for some nice to haves, definitely will consider to continue to evolve this feature. 
For this PR we are adding drag and drop with single file upload.

I hope that answers all your questions.

jacksonh:
Hey @Bad3r we actually have a PDF viewer so we can render native PDFs, and allow users to highlight + add notes. Thanks for mentioning epub, we had actually just been talking about it, so its good to hear other people are interested.

jacksonh:
@Bad3r in terms of parsing the PDFs, we do use pdf.js on the backend to try and extract all the content of the PDF to create a search index for it. I think Logseq uses the same library.

rupin90:
> It looks like there is some z-order issue, i see the UI show up behind the cards
> 
> Screen Shot 2022-11-21 at 20 57 43

should be good now, Having major internet issues 😓

jacksonh (approved):

PR #1406, opened 2022-11-11
set voice in user personalization

sywhb:
- Set voice and secondary voice in user personalization
- Fix typeorm issue

### DIFF
--- a/packages/api/src/utils/corsConfig.ts
+++ b/packages/api/src/utils/corsConfig.ts
@@ -1,5 +1,8 @@
+import { env } from '../env'
+
 export const corsConfig = {
   credentials: true,
+  // allow https://studio.apollographql.com for local env
   origin: [
     'https://omnivore.app',
     'https://dev.omnivore.app',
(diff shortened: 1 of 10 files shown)


jacksonh (approved):

packages/api/src/utils/corsConfig.ts
```diff
@@ -9,5 +9,6 @@ export const corsConfig = {
     'https://web-demo.omnivore.app',
     'http://localhost:3000',
     'lsp://logseq.io',
+    'https://studio.apollographql.com',
```
jacksonh, line 12: I think with this one it will only work locally anyways, right?

packages/api/src/utils/corsConfig.ts
```diff
@@ -9,5 +12,6 @@ export const corsConfig = {
     'https://web-demo.omnivore.app',
     'http://localhost:3000',
     'lsp://logseq.io',
+    env.dev.isLocal && 'https://studio.apollographql.com',
```
sywhb, line 15: @jacksonh I updated the code and now it will only allow Apollo sandbox access in local env

jacksonh: I didn't realize that syntax works, cool

PR #1408, opened 2022-11-11
Replace filter clause with must clause in the search query in elastic

sywhb:


jacksonh:
Cool, we can probably deploy this to demo for the weekend.

sywhb:
> Cool, we can probably deploy this to demo for the weekend.

Yes, I will do lots of testing during the weekend too

jacksonh (approved):

PR #1417, opened 2022-11-15
Better content extraction of title and description for GitHub URLs

jacksonh:


packages/content-handler/src/websites/github-handler.ts
```diff
@@ -12,10 +12,29 @@ export class GitHubHandler extends ContentHandler {
 
   async preParse(url: string, dom: Document): Promise<Document> {
     const body = dom.querySelector('body')
+    const head = dom.querySelector('head')
```
sywhb, line 15: Where is `head` being used?

packages/content-handler/src/websites/github-handler.ts
```diff
@@ -12,10 +12,29 @@ export class GitHubHandler extends ContentHandler {
...
+      body.replaceChildren(article)
+
+      // Attempt to set the author also. This is available on repo homepages
+      // but not on things like PRs. Ideally we want PRs and issues to have
+      // author set to the author of the PR/issue.
+      if (linkAuthor && linkAuthor.textContent) {
+        const author = dom.createElement('span')
+        author.setAttribute('rel', 'author')
+        author.innerHTML = linkAuthor.textContent
+        article.appendChild(author)
```
sywhb, line 30: Nice!

packages/content-handler/src/websites/github-handler.ts
```diff
@@ -12,10 +12,29 @@ export class GitHubHandler extends ContentHandler {
...
+      }
+    }
+
+    // Remove the GitHub - and repo org from the title
+    const twitterTitleContent = twitterTitle?.getAttribute('content')
+    if (twitterTitle && twitterTitleContent) {
+      twitterTitle.setAttribute('content', twitterTitleContent.replace(/GitHub - .*\//, ''))
```
sywhb, line 37: The `content` is like "omnivore: Omnivore is a complete, open source read-it-later solution for people who like reading.". Do we want to keep the repo name?

sywhb (approved):
LGTM 👍

packages/content-handler/src/websites/github-handler.ts
jacksonh: @sywhb the idea is we will get the repo name as the author.

packages/content-handler/src/websites/github-handler.ts
jacksonh: Took so long to find 😬

PR #1424, opened 2022-11-16
fix/gdcvault parsing issue

sywhb:
- Make headless = false in local testing
- Add puppeteer-extra plugins
- Add gdcvault parsing test-case
- Allow kaltura videos through in readablity
- Make overlay element an unlikelyCandidates and give player element positive score


jacksonh:
Didn't we have issues with `AdblockerPlugin` before? We can probably put on demo and test out some more.

jacksonh (approved):

sywhb:
> Didn't we have issues with `AdblockerPlugin` before? We can probably put on demo and test out some more.

I guess they fixed the issue and now we have no problem using the plugin

PR #1425, opened 2022-11-16
Improve how we handle loading the initial items and recently saved items on iOS

jacksonh:
This PR contains some performance fixes and new embedded highlights. When a user saves a page, we embed some markers around their selected text to attempt to create a highlight on the backend. 

The backend will find these markers, and generate diff patches from them using an algorithm similar to the frontend's highlight generator.

TODO: 
- [ ] update UI to display highlights in the share extension if found
- [ ] Make sure readability does not remove a highlight section


apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
```diff
@@ -91,49 +91,53 @@ import Views
...
     let syncResult = try? await dataService.syncLinkedItems(since: lastSyncDate, cursor: nil)
     if syncResult != nil {
       lastItemSyncTime = dateFormatter.string(from: syncStartTime)
     }
+    // If possible start prefetching new pages in the background
+    if let itemIDs = syncResult?.updatedItemIDs,
+       let username = dataService.currentViewer?.username,
+       itemIDs.count > 0
+    {
+      Task {
```
satindar, line 121: Looks like the Task block isn't needed now that the function is marked as async.

apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
```diff
@@ -159,17 +163,33 @@ import Views
...
+  }
+
+  func loadItems(dataService: DataService, audioController _: AudioController, isRefresh: Bool) async {
+    let syncStartTime = Date()
+    let start = CFAbsoluteTimeGetCurrent()
+
+    isLoading = true
+    showLoadingBar = true
+
+    _ = await [
```
satindar, line 179: Interesting syntax! Do you know if this loads concurrently?

apple/OmnivoreKit/Sources/Models/DataModels/FeedItem.swift
```diff
@@ -12,6 +12,16 @@ public struct LinkedItemQueryResult {
   }
 }
 
+public struct LinkedItemSyncResult {
```
satindar, line 15: nice

apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
jacksonh: What I was trying to achieve here is allowing the function to return without waiting on this call, but I might be misunderstanding what happens here.

apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
jacksonh: I believe so, but I think a taskGroup is a bit more clear. I'll switch to that

apple/OmnivoreKit/Sources/Models/DataModels/FeedItem.swift
jacksonh: more structs 😬

packages/api/src/services/save_page.ts
```diff
@@ -139,14 +143,37 @@ export const savePage = async (
...
       }
     }
   }
 
+  if (pageId && parseResult.highlightData) {
+    const highlight = {
+      updatedAt: new Date(),
+      createdAt: new Date(),
+      userId: ctx.uid,
+      elasticPageId: pageId,
```
sywhb, line 160: We used to have this in the database and we might not need it anymore

packages/api/src/utils/highlightGenerator.ts
```diff
@@ -0,0 +1,328 @@
...
+        pushNode(node)
+      }
+    } else {
+      if (!nonParagraphTagsRegEx.test((node as Element).tagName))
+        newParagraph = true
+    }
+
+    for (
+      let i = 0, len = node.childNodes.length;
+      !reachedEndNode && i < len;
```
sywhb, line 78: I guess we could check `reachedEndNode` before checking the `nodeType` on line 63

packages/api/src/utils/highlightGenerator.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  getTextNodes(rootNode)
+
+  return {
+    textNodes,
+    articleText,
+  }
+}
+
+export function findEmbeddedHighlight(
```
sywhb, line 97: There is some mixed use of `function` and `arrow const` here. May I know if there is any reason for that?

packages/api/src/utils/parser.ts
```diff
@@ -16,6 +16,11 @@ import { ILike } from 'typeorm'
 import { v4 as uuid } from 'uuid'
 import addressparser from 'addressparser'
 import { preParseContent } from '@omnivore/content-handler'
+import {
+  findEmbeddedHighlight,
+  EmbeddedHighlightData,
+} from './highlightGenerator'
+import { HighlightData } from '../datalayer/highlight/model'
```
sywhb, line 23: Maybe we could use `import { Highlight } from '../../elastic/types'`

packages/api/src/utils/highlightGenerator.ts
```diff
@@ -0,0 +1,328 @@
...
+  if (!patch) throw new Error('Invalid patch')
+  const textNodes = articleTextNodes.textNodes
+
+  const { highlightTextStart, highlightTextEnd } = selectionOffsetsFromPatch(
+    articleTextNodes.articleText,
+    patch
+  )
+
+  // Searching for the starting text node using interpolation search algorithm
+  const textNodeIndex = interpolationSearch(
```
sywhb, line 191: Nice 👍

packages/api/src/services/save_page.ts
```diff
@@ -139,14 +143,37 @@ export const savePage = async (
...
+      ...parseResult.highlightData,
+    }
+
+    if (
+      !(await addHighlightToPage(pageId, highlight, {
```
sywhb, line 165: Not sure if it will help to speed up a bit if we save the highlight with the page together in one call

sywhb (approved):
Looks great! Although I didn't really look into a few of the functions like `getSurroundingText` thoroughly since we copied those from frontend code

packages/api/src/services/save_page.ts
sywhb: I mean the property `elasticPageId`

jacksonh: Oh, so its like a back link from the highlight to the page? It does look like its being set now for other highlights

packages/api/src/utils/highlightGenerator.ts
jacksonh: Yeah, which do you prefer? I'll make it consistent. I think in the past we mostly used arrow const

packages/api/src/services/save_page.ts
sywhb: Oh then maybe we better keep it

packages/api/src/utils/parser.ts
jacksonh: Thanks, not even needed, just some junk VS Code accidentally added.

packages/api/src/services/save_page.ts
jacksonh: Yeah maybe for a future optimization.

packages/api/src/utils/highlightGenerator.ts
jacksonh: Let me leave this one as-is for now, I think i need to add some extra tests for the paragraph wrapping stuff, and it could effect the else block below.

PR #1426, opened 2022-11-16
Use greedy matching on the GitHub title regex

jacksonh:
This fixes issues with titles that have slashes (/) in their title
as the regex match would look for the last possible matching
slash.

### DIFF
--- a/packages/content-handler/src/websites/github-handler.ts
+++ b/packages/content-handler/src/websites/github-handler.ts
@@ -35,7 +35,7 @@ export class GitHubHandler extends ContentHandler {
     if (twitterTitle && twitterTitleContent) {
       twitterTitle.setAttribute(
         'content',
-        twitterTitleContent.replace(/GitHub - .*\//, '')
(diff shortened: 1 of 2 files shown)


packages/content-handler/src/websites/github-handler.ts
```diff
@@ -35,7 +35,7 @@ export class GitHubHandler extends ContentHandler {
     if (twitterTitle && twitterTitleContent) {
       twitterTitle.setAttribute(
         'content',
-        twitterTitleContent.replace(/GitHub - .*\//, '')
+        twitterTitleContent.replace(/GitHub - (.*?)\//, '')
```
sywhb, line 38: Nice. Could there be a slash in the `owner` name?

jacksonh: eek, good catch. I did just look it up and they can not `All characters must be either a hyphen ( - ) or alphanumeric.
`

sywhb: Thanks!

sywhb (approved):

PR #1429, opened 2022-11-17
Fix crash in readability when stripping class names

jacksonh:
- Fix issue where stripping classNames could cause a crash

className can be null or can be a dictionary, so we need to verify its a non-null string before calling startsWith on it.


sywhb (approved):
LGTM

PR #1446, opened 2022-11-23
Add label actions in rule engine

sywhb:
- Add add label action in the rule engine
- Validate pageId before adding labels


jacksonh:
I guess we don't create if it doesn't exist, like the Newsletter label. Seems kind of hard to do that from the rule though.

jacksonh (approved):

jacksonh:
Yeah that makes sense. Pretty easy to use a dropdown. Probably each action type can have a nice custom UI, for now I'm making it very generic

PR #1449, opened 2022-11-23
Use search API to filter page in rule engine

sywhb:


jacksonh (approved):
This looks good to me, how do you feel about it now that implementing?

sywhb:
I feel more confident now. 
I also made a small change to run the filter for each rule before calling actions and have a `Promise.all` for all the actions; so the actions will not be affected by each other.

jacksonh:
Oh the promise.all thing is really nice, in the future we could probably add a special token to bypass rate limiting.

sywhb:
Yeah, rate-limiting is a pain in the arse. Maybe for now we only allow max 10 rules per user?

PR #1456, opened 2022-11-24
Allow changing byline in update api

sywhb:


jacksonh:
hah, that was fast!

sywhb:
Thanks!

packages/api/src/resolvers/update/index.ts
```diff
@@ -39,6 +39,7 @@ export const updatePageResolver = authorized<
     id: input.pageId,
     title: input.title || undefined,
     description: input.description || '',
+    author: input.byline ?? undefined,
```
jacksonh, line 42: I wonder, if old clients call this API, they will be sending undefined, so if the author is already set, would it then get set to undefined?

sywhb: no, it won't

jacksonh: oh because it would be undefined and removed from what is being set. Thanks!

jacksonh (approved):

packages/api/src/resolvers/update/index.ts
sywhb: yeah

PR #1475, opened 2022-11-28
Allow setting labels from the highlights list view on iOS

jacksonh:
- Update GQL schema adding labels to highlights
- Add labels to highlights
- Allow setting labels on highlights


jacksonh:
This will also give us the square-ish labels on web, but they seem to look ok there too.

PR #1478, opened 2022-11-29
Fix NYTimes first few paragraph missing

sywhb:
The first few paragraphs are deleted by readability because there are too many links in them but these links are not advertisements so I added the class name of the element of the paragraphs to the skipping list when cleaning the nodes.

Test case added too

### DIFF
--- a/packages/readabilityjs/Readability.js
+++ b/packages/readabilityjs/Readability.js
@@ -2785,7 +2785,7 @@ Readability.prototype = {
           return false;
         }
 
-        var parentClasses = node.parentNode.classList;
+        var parentClasses = node.parentNode?.classList || [];
         var haveToRemove =
           !this._isOmnivoreNode(node) && (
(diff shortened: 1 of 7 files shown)


packages/readabilityjs/Readability.js
```diff
@@ -2782,14 +2785,15 @@ Readability.prototype = {
           return false;
         }
 
+        var parentClasses = node.parentNode.classList;
```
jacksonh, line 2788: Can this be null?

sywhb: Nice catch! Thank you

packages/readabilityjs/Readability.js
```diff
@@ -2782,14 +2785,15 @@ Readability.prototype = {
           return false;
         }
 
+        var parentClasses = node.parentNode?.classList || [];
```
sywhb, line 2788: @jacksonh Default `parentClasses` is an empty array now

jacksonh (approved):
Will have to make sure we do some testing on demo.

PR #1480, opened 2022-11-29
Webview Highlight Taps - Android

satindar:
- send highlightID in json payload send to native app
- generate new js bundles for native mobile apps


satindar:
@jacksonh Will probably need some follow up fixes but this is working now.

jacksonh (approved):

PR #1484, opened 2022-11-30
Update savedAt in update page api

sywhb:


packages/api/src/resolvers/update/index.ts
```diff
@@ -37,9 +37,10 @@ export const updatePageResolver = authorized<
 
   const pageData = {
     id: input.pageId,
-    title: input.title || undefined,
-    description: input.description || '',
+    title: input.title ?? undefined,
+    description: input.description ?? undefined,
     author: input.byline ?? undefined,
+    savedAt: input.savedAt ? new Date(input.savedAt) : undefined,
```
jacksonh, line 43: Looks good, if input is invalid it will just throw which i think is what we want.

jacksonh (approved):

PR #1487, opened 2022-11-30
saved filters api

sywhb:
- Add filters table and entity
- Update table name
- Add saved filter api


jacksonh:
I think the only missing thing here is the ability to reorder them which probably people would want.

sywhb:
> I think the only missing thing here is the ability to reorder them which probably people would want.

Let me add a `position` or `order` param here

jacksonh (commented):
All looks good to me. I need to implement on iOS to see how the move API works with what iOS and React provide for reordering lists.

PR #1490, opened 2022-12-01
Fix parsing guardian news

sywhb:


jacksonh (approved):

jacksonh:
I think this one passes now right? Maybe we should just merge the tests

PR #1492, opened 2022-12-01
Make the web notebook UI better match the iOS UI

jacksonh:


jacksonh:
Going to deploy this one to demo

PR #1504, opened 2022-12-04
Feat/dragdrop - This is for the Library List page

rupin90:
***LibraryList*** page was not completed with the **HomeFeedContainer** so adding same drag and drop functionality on the **LibraryList** page


jacksonh:
Will close this one out for now as we have some new library design that we might work into the existing library page.

PR #1508, opened 2022-12-05
join group api

sywhb:
- Add unique constraint on user_id and group_id to group_membership table
- Grant update permission on invite, group_membership and group tables
- Add joinGroup API


packages/api/src/services/create_group.ts
```diff
@@ -91,3 +91,59 @@ export const getRecommendationGroups = async (
 export const getInviteUrl = (invite: Invite) => {
   return `${homePageURL()}/invite/${invite.code}`
 }
+
+export const joinGroup = async (
```
jacksonh, line 95: Maybe should be in a file name `join_group` or we could rename this file to `groups`

jacksonh (approved):
Might want to change that file name before merging.

packages/api/src/services/create_group.ts
sywhb: Renamed this file to `groups.ts` @jacksonh

PR #1516, opened 2022-12-08
recommend highlight api

sywhb:
- Add recommendHighlights API
- Recommend highlights
- Allow recommended With Highlights

### DIFF
--- a/packages/api/src/generated/schema.graphql
+++ b/packages/api/src/generated/schema.graphql
@@ -1029,6 +1029,7 @@ type Mutation {
   moveLabel(input: MoveLabelInput!): MoveLabelResult!
   optInFeature(input: OptInFeatureInput!): OptInFeatureResult!
   recommend(input: RecommendInput!): RecommendResult!
+  recommendHighlights(input: RecommendHighlightsInput!): RecommendHighlightsResult!
   reportItem(input: ReportItemInput!): ReportItemResult!
   revokeApiKey(id: ID!): RevokeApiKeyResult!
   saveArticleReadingProgress(input: SaveArticleReadingProgressInput!): SaveArticleReadingProgressResult!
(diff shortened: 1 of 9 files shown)


packages/api/src/schema.ts
```diff
@@ -2318,6 +2344,9 @@ const schema = gql`
     createGroup(input: CreateGroupInput!): CreateGroupResult!
     recommend(input: RecommendInput!): RecommendResult!
```
jacksonh, line 2345: I wonder if its better to just extend the `recommend` API? like `includeHighlights: bool`. At the time of recommending, does the backend have access to all the highlights?

packages/api/src/generated/schema.graphql
```diff
@@ -1248,16 +1249,40 @@ enum RecommendErrorCode {
...
+
+type RecommendHighlightsSuccess {
+  success: Boolean!
+}
+
 input RecommendInput {
   groupIds: [ID!]!
   note: String
   pageId: ID!
+  recommendedWithHighlights: Boolean
```
sywhb, line 1279: @jacksonh There is `recommendedWithHighlights` in the `RecommendInput`

packages/api/src/schema.ts
sywhb: We have both methods to recommend highlights.
Just want to be more flexible. You can just use the original `recommend` api if you want

packages/api/src/generated/schema.graphql
jacksonh: Ah i see, so from the client perspective I would probably always use that option. Will probably add a toggle on the share box that people can use to enable it.

sywhb: So you want me to remove the recommendHighlight API?

jacksonh: @sywhb I don't think that we need it now, I can see how maybe in the future we might want to share a single highlight or something though.

Its probably ok to leave it in.

sywhb: Cool. We can merge it now

jacksonh (approved):

jacksonh:
Going to merge this one so I can test it on demo

PR #1521, opened 2022-12-08
Add delete permission to group and membership tables

sywhb:


jacksonh:
classic

packages/api/src/services/groups.ts
```diff
@@ -172,19 +172,25 @@ export const leaveGroup = async (
...
-        (m) => m.isAdmin && m.user.id !== user.id
+        (m) => m.isAdmin && m.id !== membershipId
       )
       if (!hasAdmin) {
         const newAdmin = group.members.find((m) => !m.isAdmin)
         if (!newAdmin) {
-          // delete the group if there are no more members
-          await t.getRepository(Group).delete({ id: group.id })
-          return true
+          throw new Error('No user found')
```
jacksonh, line 193: So it wont delete the group anymore? (That is ok with me for now)

jacksonh (approved):

packages/api/src/services/groups.ts
sywhb: It will still delete the group (I moved the code to the other place)

sywhb: @jacksonh

packages/api/src/services/groups.ts
```diff
@@ -172,19 +172,25 @@ export const leaveGroup = async (
       throw new Error('User not in group')
     }
 
+    const membershipId = membership.id
     await t.getRepository(GroupMembership).remove(membership)
 
+    if (group.members.length === 1) {
```
jacksonh, line 178: Ah i see it!

jacksonh (approved):

packages/api/src/services/groups.ts
```diff
@@ -172,19 +172,25 @@ export const leaveGroup = async (
...
+      // delete the group if there are no more members
+      await t.getRepository(Group).remove(group)
+
+      return true
+    }
+
```
sywhb, line 184: I move the deletion of group here to fix a bug @jacksonh

PR #1524, opened 2022-12-08
Bump Android Deps

satindar:
- update compose version
- update gradle version
- update apollo android version
- bump multiple dependency versions
- bump android target sdk to 33


satindar:
@jacksonh Will need to update Android Studio (to Dolphin) to run the app with these changes.

satindar:
This will give us access to the new pull to refresh UI.

jacksonh (approved):

PR #1525, opened 2022-12-09
Add recommendation to the original page too

sywhb:


jacksonh:
One thing I just thought, with the default rules would this trigger a notification to the current user?

sywhb:
> One thing I just thought, with the default rules would this trigger a notification to the current user?

Yeah, it would

jacksonh (approved):

PR #1527, opened 2022-12-09
Upsize max highlight char length to 6000

sywhb:


jacksonh (approved):

packages/web/lib/highlights/highlightGenerator.ts
```diff
@@ -12,7 +12,7 @@ const highlightTag = 'omnivore_highlight'
 const highlightClassname = 'highlight'
 const highlightWithNoteClassName = 'highlight_with_note'
 const articleContainerId = 'article-container'
-export const maxHighlightLength = 2000
+export const maxHighlightLength = 6000
```
sywhb, line 15: @jacksonh I changed the limit in Web too.

Does it require any changes in the iOS client?

jacksonh: I think the client should work ok, I might need to re-run to generate the sanitization types

PR #1534, opened 2022-12-12
For URLs that will be forced to backend fetch dont send full content from the extension

jacksonh:
This is useful with YouTube, which has massive page sizes.


sywhb (approved):
LGTM

PR #1535, opened 2022-12-12
Check event type in the filter

sywhb:


jacksonh (approved):
Nice, seems like we just need to deploy and test.

PR #1536, opened 2022-12-12
Do not save recent searches for rule based queries

sywhb:


packages/api/src/resolvers/article/index.ts
```diff
@@ -908,7 +909,10 @@ export const searchResolver = authorized<
 
   // save query, including advanced search terms, in search history
   if (params.query) {
-    await saveSearchHistory(claims.uid, params.query)
+    const client = httpContext.get('client') as string | undefined
+    // don't save search history for rule based queries
+    client !== 'rule-handler' &&
```
jacksonh, line 914: was wondering how you'd do this 👍

packages/rule-handler/src/filter.ts
```diff
@@ -66,6 +66,7 @@ export const search = async (
         headers: {
           Cookie: `auth=${auth};`,
           'Content-Type': 'application/json',
+          'X-OmnivoreClient': 'rule-handler',
```
jacksonh, line 69: `client`  set to `X-OmnivoreClient` in middleware?

sywhb: Yeah, we set the `client` in the `httpContext` in middler

sywhb:
```
  // set client info in the request context
  app.use(httpContext.middleware)
  app.use('/api/', (req, res, next) => {
    const client = req.header('X-OmnivoreClient')
    if (client) {
      httpContext.set('client', client)
    }
    next()
  })
```

PR #1555, opened 2022-12-17
Fix: Android offline crashes

satindar:


satindar:
This fixes the crash but still feels wrong. Apollo is crashing when network connections fail rather than returning an error. I wrapped everything in try/catch blocks but wondering of I'm using Apollo wrong. It should return an error when the calls fail. Will do some more research or try to wrap our own function around the `execute` call to apollo.

satindar:
Also saw a crash from Segment when offline. Only happened once though.

jacksonh (approved):
I wouldn't be surprised if this is how you are supposed to do it in Android, I'd expect some sort of typed Exception though so you can see the issue.

PR #1561, opened 2022-12-19
Add transcripts to YouTube pages

jacksonh:


packages/content-handler/src/websites/youtube-handler.ts
```diff
@@ -66,11 +106,12 @@ export class YoutubeHandler extends ContentHandler {
       <iframe width="${width}" height="${height}" src="https://www.youtube.com/embed/${videoId}" title="${title}" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
         <p><a href="${url}" target="_blank">${title}</a></p>
         <p itemscope="" itemprop="author" itemtype="http://schema.org/Person">By <a href="${oembed.author_url}" target="_blank">${authorName}</a></p>
+        <p class='omnivore-youtube-transcript'>
```
jacksonh, line 109: One issue is what happens if this is null? I guess just an empty string wouldn't cause any issues here.

packages/content-handler/src/websites/youtube-handler.ts
```diff
@@ -53,6 +84,15 @@ export class YoutubeHandler extends ContentHandler {
     const width = height * ratio
     const authorName = _.escape(oembed.author_name)
 
+    console.log('got video id', videoId)
+    const defaultLanguageCode = await this.getDefaultLanguageCode(videoId)
+
+    const response = await YoutubeTranscript.fetchTranscript(videoId, {
```
jacksonh, line 90: Do we know if this can throw?

sywhb: Added a try-catch block here

packages/content-handler/src/websites/youtube-handler.ts
sywhb: Default empty string

jacksonh:
nice, the one thing I'm a bit worried about is the regex to parse the JSON.

sywhb:
> nice, the one thing I'm a bit worried about is the regex to parse the JSON.

Yeah, but I'm not sure about the JSON data structure of the response.
Let me find one

jacksonh:
I think this might fail with this URL but need to test again: https://www.youtube.com/watch?v=KdQHAeAnHmw&t=377s

jacksonh:
Nevermind, it just failed with my code to fetch bookmarks too

SilentVoid13:
Hey there @jacksonh, any update on this? That's a nice feature that I would love to see in Omnivore

hientt53:
I'm really looking forward to this feature from you

hanislm:
Kant wait.

Highlight during read aloud of an article or while watching the the transcription flow of youtube or podcast would also be *chefs kiss*.

rilaby:
Highly awaited feature. 
Are there any particular help needed to be merged in?

jacksonh:
@rilaby I'm going to close this one as I have a more full featured branch i am working on now. Some other upgrades allowed us to simplify this work

rilaby:
May I follow that branch, if you'd be able to share it?

PR #1575, opened 2022-12-27
Add cloud function for readability

sywhb:


jacksonh:
NIce, I'm thinking of how this would fit into the flow of slow processing. How would API know to call this one vs doing inline?

sywhb:
> NIce, I'm thinking of how this would fit into the flow of slow processing. How would API know to call this one vs doing inline?

Could we pass a param like `isInline=false` to createArticle API?

jacksonh:
Yeah, that makes sense, but I also wonder, could we put this inside of the `pupeteer-parse` service too? Or maybe wondering, when this finishes readability can we call the API back with the content?

sywhb:
@jacksonh Good idea. I Agree we should put this inside the `puppeteer-parse` service

sywhb:
We could put inline readability in the `puppeteer-parse` service so that after the page is scraped, we could parse the content with inline readability and call the API with parsed content @jacksonh

jacksonh:
Yeah if we ran in puppeteer-parse, then maybe we could send the originalContent and content in the savepage API request. I think we would have to modify it for that though.

sywhb:
I will close this PR and reopen a new one @jacksonh

Issue #1503, opened 2022-12-03
Highlight messes up whitespaces for Stanford Encyclopedia of Philosophy website


cleanclothes-pi:
## Problem
Highlighting texts causes some whitespaces to disappear, consequently mixing words together.

## How to reproduce
1) Save an article from the [Stanford Encyclopedia of Philosophy](https://plato.stanford.edu/index.html) website, 
2) Highlight some text. Notice all the words in that paragraphs get shifted and mixed, _even if not highlighted_.
3) Copy-pasted the text into other appliation or synced highlight via Logseq-omnivore. See that whitespaces getting messed up in both cases.

If we delete the highlight and refresh the page, the whitespaces return to where they should be. I tried but cannot reproduced this with other websites.

## Screenshots
Before highlighting, whitespaces are the same as original.
<img width="811" alt="Screen Shot 2022-12-03 at 15 53 13" src="https://user-images.githubusercontent.com/58788103/205462421-e9e45c1e-a639-4f02-aaff-91c362120bcb.png">

After highlighting, whitespaces disappear.
<img width="799" alt="Screen Shot 2022-12-03 at 15 53 03" src="https://user-images.githubusercontent.com/58788103/205462425-bd377bd0-1fcc-40c2-a09b-a457fdf7f906.png">

jacksonh:
Thanks, taking a look at this one. I suspect some hidden spans that were treated as non-whitespace.

jacksonh:
@cleanclothes-pi this should be fixed now, let me know if you still see the issue.

cleanclothes-pi:
I checked and this works now. Thanks!

PR #1576, opened 2022-12-27
Fix issue with spaces dissapearing when highlights are created

jacksonh:
Use the raw text when creating a highlight node instead of the
escaped text. This means \n spaces will be preserved in the
node and prevent spaces from being removed.


jacksonh:
Fixes: https://github.com/omnivore-app/omnivore/issues/1503

PR #1583, opened 2022-12-29
iOS clubs 

jacksonh:
Still a WIP, adding clubs, community modal, syncing fixes


pkg/admin/src/db.ts
```diff
@@ -8,25 +8,25 @@ import {
   JoinColumn,
```
jacksonh, line 8: These changes should really be separate but I think fine to mix this in. Was mostly just a model change so i could update a user's name for them.

PR #1610, opened 2023-01-04
Remove some places we are logging too much data causing sentry failures

jacksonh:


packages/api/src/routers/svc/newsletters.ts
```diff
@@ -96,7 +96,12 @@ export function newsletterServiceRouter() {
 
       const result = await saveNewsletterEmail(data)
       if (!result) {
-        console.log('Error creating newsletter link from data', data)
+        console.log(
+          'Error creating newsletter link from data',
+          data.email,
```
sywhb, line 101: This causes some lint error

sywhb: I guess we need to define the data structure

jacksonh: yeah, kind of annoying because this is right above 

```
if (
        !('email' in data) ||
        !('content' in data) ||
        !('title' in data) ||
        !('author' in data)
      ) {
```

sywhb (approved):
Awesome, thanks! I always wanted to do this 👍

jacksonh:
I hope it works, i have a feeling there are more places but its very hard to track down in our logging.

sywhb:
Yeah, I guess we could merge this one first and see if there is any more sentry report coming

PR #1612, opened 2023-01-04
Send email to user after a bulk import completes

jacksonh:


packages/import-handler/src/task.ts
```diff
@@ -10,7 +10,10 @@ type TaskPayload = {
 
 const cloudTask = new CloudTasksClient()
 
-export const createCloudTask = async (payload: TaskPayload) => {
+export const createCloudTask = async (
+  payload: unknown,
```
sywhb, line 14: I guess we could remove type `TaskPayload` since it's not used anymore

packages/import-handler/src/index.ts
```diff
@@ -50,6 +56,30 @@ const importURL = async (
...
+  const exp = Math.floor(Date.now() / 1000) + 60 * 60 * 24 // 1 day
+  const authToken = (await signToken(
+    { uid: userId, exp },
+    process.env.JWT_SECRET
+  )) as string
+  const headers = {
+    Authorization: `auth=${authToken}`,
+  }
+
+  return createCloudTask(
```
sywhb, line 73: Should we put the task handler URL in the param?

packages/import-handler/src/index.ts
```diff
@@ -79,21 +109,30 @@ export const importHandler: EventFunction = async (event, context) => {
...
-        const userId = [...groups][1]
         const result = await importURL(userId, url, 'csv-importer')
         console.log('import url result', result)
+        countImported = countImported + 1
       } catch (err) {
         console.log('error importing url', err)
       }
     })
+
+    await importCompletedTask(userId, countImported)
```
sywhb, line 136: Should we send a different email if there is no import?

sywhb (commented):
Looks good. I just post a few comments

packages/import-handler/src/index.ts
sywhb: In this case, it's the email sending api

packages/import-handler/src/index.ts
```diff
@@ -50,6 +56,30 @@ const importURL = async (
...
+    throw 'Envrionment not setup correctly'
+  }
+
+  const exp = Math.floor(Date.now() / 1000) + 60 * 60 * 24 // 1 day
+  const authToken = (await signToken(
+    { uid: userId, exp },
+    process.env.JWT_SECRET
+  )) as string
+  const headers = {
+    Authorization: `auth=${authToken}`,
```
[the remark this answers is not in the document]
jacksonh: Oh you are right, i guess the `auth=` is for when using a cookie

packages/import-handler/src/task.ts
jacksonh: RIP types

packages/import-handler/src/index.ts
sywhb: Something like "Sorry but your import failed"

packages/import-handler/src/index.ts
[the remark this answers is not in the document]
sywhb: Yeah

packages/import-handler/src/task.ts
sywhb: 🛌

packages/import-handler/src/index.ts
jacksonh: Yeah, unfortunately the main issue is pulling the userId out of the file, so some cases we can't handle well :-/

PR #1616, opened 2023-01-05
html to markdown

sywhb:
- Convert HTML to markdown if specified
- Fetch markdown content if specified in search API


jacksonh (approved):
Looks good, so the format is undefined by default so it will return html

sywhb:
> Looks good, so the format is undefined by default so it will return html

Yeah

PR #1620, opened 2023-01-06
Log email content if Sendgrid api key not set

sywhb:


packages/api/src/utils/sendEmail.ts
```diff
@@ -23,7 +23,8 @@ const asSendGridError = (error: any): SendGridError | undefined => {
 export const sendEmail = async (msg: MailDataRequired): Promise<boolean> => {
   const client = new MailService()
   if (!process.env.SENDGRID_MSGS_API_KEY) {
```
jacksonh, line 25: Do you know if in this context `env.dev.isLocal` is true, we could check that too so we throw. But maybe we throw on startup anyways if in prod and this isn't set.

sywhb: Yeah, we will throw in prod.

Also added checking `env.dev.isLocal` to make it more verbal

jacksonh (approved):

PR #1624, opened 2023-01-06
Add demo user in docker-compose migrate service

sywhb:
- Add demo user in docker-compose migrate service
- Add comment


jacksonh (approved):

jacksonh:
Only concern is if people ever try to deploy the docker-compose but I think saying its for dev only is fair.

PR #1628, opened 2023-01-09
Remove the popular subscriptions table

jacksonh:
We probably need a standalone page for this in the future, but
removing this now allows people to use this page to see their
subscriptions on web.


sywhb (approved):
LGTM

PR #1633, opened 2023-01-09
Add subscriptionCount and createdAt in NewsletterEmail type

sywhb:


packages/db/migrations/0105.undo.add_foreign_key_to_subscription.sql
```diff
@@ -0,0 +1,20 @@
+-- Type: UNDO
+-- Name: add_foreign_key_to_subscription
+-- Description: Add newsletter_email_id as foreign key to the subscription table
+
+BEGIN;
+
+-- remove old column
+ALTER TABLE omnivore.subscriptions
```
jacksonh, line 8: Oh, I thought this one was already in there. Sorry, more work than I expected.

sywhb: Actually good catch so I can add the foreign key here!

jacksonh (approved):

PR #1635, opened 2023-01-09
Fix creation of content display report

jacksonh:
Fixes this sentry issue: https://sentry.io/organizations/omnivore/issues/3810736646


sywhb (approved):
👍

jacksonh:
Added a few more things to this one @sywhb

packages/db/migrations/0105.do.remove_content_report_constraints.sql
```diff
@@ -0,0 +1,13 @@
+-- Type: DO
+-- Name: remove_content_report_constraints
+-- Description: Remove unneeded constraints from the content_report table
+
+BEGIN;
+
+ALTER TABLE omnivore.content_display_report
```
sywhb, line 7: Nice

packages/db/migrations/0105.do.remove_content_report_constraints.sql
```diff
@@ -0,0 +1,13 @@
...
+    DROP CONSTRAINT content_display_report_page_id_fkey;
+
+ALTER TABLE omnivore.content_display_report
+    DROP CONSTRAINT content_display_report_user_id_fkey;
```
sywhb, line 11: Why do we drop this foreign key?

packages/api/test/services/create_content_display_report.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  let user: User
+  let page: Page
+
+  before(async () => {
+    user = await createTestUser('fakeContentUser')
+    page = await createTestElasticPage(user.id)
+  })
+
+  after(async () => {
+    await getRepository(ContentDisplayReport).delete({
```
sywhb, line 41: We need to use the superuser here to delete `ContentDisplayReport` because it has the Row level security enabled

sywhb: I think if we delete `user`, it will also delete the `ContentDisplayReport` since it's `onDeleteCascade`

sywhb (changes requested):

jacksonh:
Ah RLS explains why I was getting such odd results. Thank you!!

packages/db/migrations/0105.do.remove_content_report_constraints.sql
jacksonh: @sywhb I think we don't want reports to delete if a user is deleted by default. It makes the tests more complicated, but I think we want to keep those reports.

sywhb: Yeah, you are right. I didn't think about that

jacksonh:
@sywhb mind taking another look at this one? We don't have delete permission on the reports table, which i think is what we want from a data governance point of view, but does mean we can't clean up after tests.

sywhb:
Hey @jacksonh, I'm thinking if it is okay just to leave the test data there so we do not clean up the content display report.

Or we might mock the database so we don't need to clean

jacksonh:
Yeah I think it is fine to leave there since really the testing database should be re-created

sywhb (approved):

PR #1639, opened 2023-01-10
Use uslug instead of slugify to better support unicode slugs

jacksonh:


jacksonh:
ok this library is very out of date, need something else.

PR #1646, opened 2023-01-11
Update matter import handler to use archives instead of just history files

jacksonh:


packages/import-handler/src/index.ts
```diff
@@ -86,26 +110,94 @@ const sendImportCompletedEmail = async (
...
+    return Promise.reject()
+  }
+  return Promise.resolve()
+}
+
+const handleEvent = async (data: StorageEvent) => {
+  if (shouldHandle(data)) {
     console.log('handling csv data', data)
 
     const stream = storage
```
sywhb, line 203: We could create read stream after the below checking

packages/import-handler/src/csv.ts
```diff
@@ -5,24 +5,19 @@
...
+export const importCsv = async (ctx: ImportContext, stream: Stream) => {
   const parser = parse()
   stream.pipe(parser)
-  let count = 0
   for await (const row of parser) {
     try {
       const url = new URL(row[0])
-      await handler(url)
+      await ctx.urlHandler(ctx, url)
+      ctx.countImported += 1
```
sywhb, line 17: I think this is good for now. Probably in the future, if we need to access this, we could save it in the memory (a map) or file

packages/import-handler/src/index.ts
```diff
@@ -1,41 +1,65 @@
...
 
 const signToken = promisify(jwt.sign)
 
 const storage = new Storage()
 
-interface StorageEventData {
-  bucket: string
+const CONTENT_TYPES = ['text/csv', 'application/zip']
+
+export type UrlHandler = (ctx: ImportContext, url: URL) => Promise<void>
```
sywhb, line 27: Sorry, I don't quite understand this part. Could you help me to understand the need to define the type here and the param `ctx` containing `urlHandler` itself?

packages/import-handler/src/matterHistory.ts
```diff
@@ -5,28 +5,234 @@
...
+      await ctx.urlHandler(ctx, url)
+      ctx.countImported += 1
     } catch (error) {
       console.log('invalid url', row, error)
+      ctx.countFailed += 1
+    }
+  }
+}
+
+const DOM_PURIFY_CONFIG = {
```
sywhb, line 45: I'm trying to find a way to move all these duplicate code in one place so we could reuse

sywhb (commented):
This looks good. A few small things come to my mind:
1. Maybe we could add a .dockerignore file
2. If a large import failed, how could we recover or continue at the failure point
3. Is there a need to split a very big import task into smaller ones so to accelerate the process?

packages/import-handler/src/index.ts
```diff
@@ -1,41 +1,65 @@
...
-type importHandlerFunc = (
-  stream: Stream,
-  handler: UrlHandler
-) => Promise<number>
-
-const shouldHandle = (data: StorageEventData, ctx: CloudFunctionsContext) => {
-  console.log('deciding to handle', ctx, data)
-  if (ctx.eventType !== 'google.storage.object.finalize') {
-    return false
+function isStorageEvent(event: any): event is StorageEvent {
```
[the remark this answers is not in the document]
jacksonh: eek yeah

packages/import-handler/src/matterHistory.ts
jacksonh: Yeah, would be nice, part of the issue is the other code is JS not typescript

packages/import-handler/src/index.ts
jacksonh: Oh yeah, so we have a few different types of import we can do: a CSV import or two types of Matter file import. So we have one function that figures out which type to use, and that returns the 'handler' function that should be used. 

Mostly this makes it a bit easier to test, i can decouple the detection of the file type from the execution.

packages/import-handler/src/index.ts
jacksonh: nice one

PR #1648, opened 2023-01-12
fix/puppeteer parse

sywhb:
- Fix readability not run in puppeteer-parse
- Fix last two tweets not saved in a long thread by waiting for document loaded after scrolling

### DIFF
--- a/packages/content-handler/src/websites/twitter-handler.ts
+++ b/packages/content-handler/src/websites/twitter-handler.ts
@@ -274,7 +274,7 @@ const getTweetIds = async (
         }
 
         window.scrollBy(0, distance)
-        await waitFor(100)
+        await waitFor(500)
         currentHeight += distance
       }
 
@@ -364,6 +364,7 @@ export class TwitterHandler extends ContentHandler {
     `
 
     const content = `
+<html>
     <head>
       <meta property="og:image" content="${authorImage}" />
       <meta property="og:image:secure_url" content="${authorImage}" />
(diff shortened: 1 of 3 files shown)


packages/api/src/services/subscriptions.ts
```diff
@@ -50,7 +50,7 @@ export const saveSubscription = async ({
   const result = await getRepository(Subscription).upsert(
     {
       name,
-      newsletterEmail,
+      newsletterEmail: { id: newsletterEmail.id },
```
jacksonh, line 53: Is this a different issue?

packages/content-handler/src/websites/twitter-handler.ts
```diff
@@ -274,7 +274,7 @@ const getTweetIds = async (
         }
 
         window.scrollBy(0, distance)
-        await waitFor(100)
+        await waitFor(500)
```
jacksonh, line 277: :-(

jacksonh (approved):

packages/api/src/services/subscriptions.ts
sywhb: Yeah, I think this is not causing any issue just precaution

packages/content-handler/src/websites/twitter-handler.ts
sywhb: Yeah, I know...

The issue is 100ms is not enough to load some tweets with big content so I have to increase it a bit

PR #1667, opened 2023-01-17
recent emails

sywhb:
- Add received_emails db table
- Add received_email entity
- Add saveReceivedEmail service
- Save received email
- Add list recent emails api
- Add mark a recent email as library item api


packages/api/src/services/received_emails.ts
```diff
@@ -0,0 +1,22 @@
...
+import { getRepository } from '../entity/utils'
+
+export const saveReceivedEmail = async (
+  from: string,
+  to: string,
+  subject: string,
+  text: string,
+  html: string,
+  userId: string,
+  type: 'article' | 'non-article' = 'non-article'
```
jacksonh, line 11: Does non-article include PDF attachment messages? I don't think we have to handle those specially, just wondering out loud.

packages/db/migrations/0107.do.received_emails.sql
```diff
@@ -0,0 +1,44 @@
...
+    updated_at timestamptz NOT NULL DEFAULT current_timestamp
+);
+
+CREATE TRIGGER received_emails_modtime BEFORE UPDATE ON omnivore.received_emails
+    FOR EACH ROW EXECUTE PROCEDURE update_updated_at_column();
+
+GRANT SELECT, INSERT, UPDATE, DELETE ON omnivore.received_emails TO omnivore_user;
+
+-- Create a trigger to keep the most recent 20 emails for each user
+CREATE OR REPLACE FUNCTION omnivore.delete_old_received_emails()
```
jacksonh, line 26: wild, I was wondering if this was possible

packages/db/migrations/0107.do.received_emails.sql
```diff
@@ -0,0 +1,44 @@
+-- Type: DO
+-- Name: received_emails
+-- Description: Create a table for received emails
+
+BEGIN;
+
+CREATE TABLE omnivore.received_emails (
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v1mc(),
+    user_id uuid NOT NULL REFERENCES omnivore.user ON DELETE CASCADE,
```
jacksonh, line 9: nice, this is good.

jacksonh:
Schema and API make sense to me. One thing I'd say is we could probably use one or two tests

packages/api/src/services/received_emails.ts
sywhb: Yeah, it should include PDF attachment

packages/db/migrations/0107.do.received_emails.sql
sywhb: I have tested it and this works.

Yeah, this might not be the best way if there are too many concurrent requests but since we don't need to keep exactly 20 emails per user, this might be good enough

jacksonh (approved):
Looks good besides some CI failure.

sywhb:
Hopefully, this time it will work. Really takes me some time because this touched quite some places.

@jacksonh Please let me know your thought. Thanks!

jacksonh:
Going to merge this one and build on top of it.

PR #1675, opened 2023-01-18
Fix error of getting subscriptions without newsletter email by inner join them

sywhb:


jacksonh (approved):

jacksonh:
Merging and will try to deploy after validating on demo

PR #1676, opened 2023-01-18
Do not escape the title of youtube video or tweet which is saved in elastic

sywhb:


packages/content-handler/src/websites/youtube-handler.ts
```diff
@@ -18,6 +18,10 @@ export const getYoutubeVideoId = (url: string) => {
   return videoId
 }
 
+export const escapeTitle = (title: string) => {
```
jacksonh, line 21: Not sure we have a utils file in this package, but maybe it could go there.

jacksonh (approved):

PR #1691, opened 2023-01-20
Improve iOS share extension and syncing

jacksonh:
- Proper case-insensitive compare on the labels in the share extension
- Fix search in filter by labels view
- Use the ApplyLabels view when users want to search for labels
- Use custom fonts for text chips in the share extension
- Cut down on loading refreshes


apple/OmnivoreKit/Sources/App/AppExtensions/Share/ShareExtensionScene.swift
```diff
@@ -1,8 +1,11 @@
 import SwiftUI
 import Utils
+import Views
 
 public extension PlatformViewController {
   static func makeShareExtensionController(extensionContext: NSExtensionContext?) -> PlatformViewController {
+    registerFonts()
```
satindar, line 7: Oh good catch here!

apple/OmnivoreKit/Sources/Views/Resources/en.lproj/Localizable.strings
```diff
@@ -10,8 +10,8 @@
 
 "labels.generic" = "Labels";
 "emails.generic" = "Emails";
-"subscriptions.generic" = "Text to Speech";
```
satindar, line 13: This is what happens when I stop drinking real coffee 😬

satindar (approved):
Looks awesome! I'll do some more testing but will merge now.

apple/OmnivoreKit/Sources/Views/Resources/en.lproj/Localizable.strings
jacksonh: Hah, amazingly a user reported within an hour of testflight being published

PR #1706, opened 2023-01-24
Localizations References - iOS

satindar:
- udpates more chinese strings
- add additional english strings to .strings file
- rearrange some strings
- rearrange more strings
- convert from dot syntax to camelcase in english strings file
- convert from dot syntax to camelcase in chinese strings file
- create static props for local text
- update localtext refs
- update some local text refs


satindar:
@jacksonh Added a bunch of strings but still need to update the references to them in the views. Will do that in a follow up PR. 

Also updated the string keys to use camel case rather than the dot notation. That way the static props match the keys and we can probably use them for Android and web too.

jacksonh (approved):

PR #1713, opened 2023-01-25
String updates + unit test for syntax

satindar:
- move comment in line with translation
- add missing semiclon in strings file
- add unit tests to default ios test plan
- add unit test to make sure Localizable.strings files are syntactically correct
- modify language unit tests to work on all locales


jacksonh (approved):

holton-jiang:
Good work, toast to the next version number. 🍻

PR #1752, opened 2023-01-31
Fix for handling multiple language keyboards

jacksonh:
By specifying alphabet swift wont let users switch to their
preferred language keyboard on these views.

### DIFF
--- a/apple/OmnivoreKit/Sources/App/Views/Labels/LabelsView.swift
+++ b/apple/OmnivoreKit/Sources/App/Views/Labels/LabelsView.swift
@@ -114,10 +114,7 @@ struct CreateLabelView: View {
       .padding(.bottom, 8)
 
       TextField("Label Name", text: $newLabelName)
-      #if os(iOS)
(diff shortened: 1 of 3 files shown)


apple/OmnivoreKit/Sources/App/Views/Labels/LabelsView.swift
```diff
@@ -114,10 +114,7 @@ struct CreateLabelView: View {
       .padding(.bottom, 8)
 
       TextField("Label Name", text: $newLabelName)
-      #if os(iOS)
-        .keyboardType(.alphabet)
-      #endif
-      .textFieldStyle(StandardTextFieldStyle())
+        .textFieldStyle(StandardTextFieldStyle())
```
sywhb, line 117: Do we need to explicitly use `.default` keyboard type?

jacksonh: @sywhb my understanding is we do not need to be explicit, unless you saw something in docs.

PR #1757, opened 2023-02-01
fix/substack email not forwarded

sywhb:
- Add default empty value to the text field in received_emails table
- Accept null value for text in an email

### DIFF
--- a/packages/api/src/routers/svc/emails.ts
+++ b/packages/api/src/routers/svc/emails.ts
@@ -36,13 +36,7 @@ interface EmailMessage {
 }
 
 function isEmailMessage(data: any): data is EmailMessage {
-  return (
-    'from' in data &&
-    'to' in data &&
-    'subject' in data &&
-    'html' in data &&
-    'text' in data
-  )
+  return 'from' in data && 'to' in data && 'subject' in data
 }
 
 const logger = buildLogger('app.dispatch')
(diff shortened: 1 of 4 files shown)


packages/api/src/routers/svc/emails.ts
```diff
@@ -36,13 +36,7 @@ interface EmailMessage {
...
 
 function isEmailMessage(data: any): data is EmailMessage {
-  return (
-    'from' in data &&
-    'to' in data &&
-    'subject' in data &&
-    'html' in data &&
-    'text' in data
-  )
+  return 'from' in data && 'to' in data && 'subject' in data
```
jacksonh, line 39: I wonder if subject is even required :-/

packages/api/test/routers/emails.test.ts
```diff
@@ -140,12 +141,26 @@ describe('Emails Router', () => {
         subject,
       }
       const res = await request
-        .post('/svc/pubsub/emails/save')
```
jacksonh, line 143: nice

jacksonh (approved):

packages/api/src/routers/svc/emails.ts
sywhb: It should be optional too. Let me change that

PR #1762, opened 2023-02-02
Update Localizable.strings

holton-jiang:


satindar:
Hi @holton-jiang thanks for this PR! I'll need to update the references to these strings but will merge this and then do that.

satindar (approved):

PR #1781, opened 2023-02-06
Core Data Context Fixes

satindar:
- use lookup rather than object cache when deleting highlight
- avoid using coredata cache in data service network calls handlers


jacksonh:
I am going to merge this, then rebase my branch on it.

jacksonh:
Actually I merged this into my branch, so I will close it now

PR #1795, opened 2023-02-09
Add the original URL as a fallback when creating URL proxys

jacksonh:


jacksonh:
@sywhb does this look OK to you? Separately I will add a handler on the frontend, where if fetching the regular `src` fails we use the `omnivore-original-src` image.

packages/readabilityjs/Readability.js
```diff
@@ -545,6 +545,7 @@ Readability.prototype = {
 
           const proxySrc = this.createImageProxyUrl(absoluteSrc, width, height);
           image.setAttribute('src', proxySrc);
+          image.setAttribute('data-omnivore-original-src', absoluteSrc)
```
sywhb, line 548: 👍

sywhb (approved):
@jacksonh LGTM

PR #1799, opened 2023-02-10
generate static html for readaibility

sywhb:
- Update python script to write content into index.html
- Add GitHub action to generate static html for readability and distiller test pages
- Fix bug in action


.github/workflows/static.yml
```diff
@@ -0,0 +1,76 @@
...
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Generate static html
+        run: |
+          cd packages/readabilityjs/test
+          python generate_static.py
+  deploy:
```
jacksonh, line 60: Cool trick, I didn't know you can do this in an action

sywhb: I'm not sure if it will work because the workflow failed somewhere before

jacksonh (approved):

PR #1800, opened 2023-02-10
Add a button to render the article using distiller

jacksonh:


jacksonh:
I will close this PR but come back to it, adding an article action button that allows switching.

PR #1805, opened 2023-02-13
fix/content parsing

sywhb:
- Add readability test for habr.com
- Allow image and media to be loaded in puppeteer
- Add readability test for yuyue.com
- Fix not timeout if scroll more than 5 seconds


packages/api/src/services/create_page_save_request.ts
```diff
@@ -93,7 +93,7 @@ export const createPageSaveRequest = async (
     url: normalizedUrl,
   })
   if (page) {
-    console.log('Page already exists', page)
+    console.log('Page already exists', page.id, page.url)
```
jacksonh, line 96: 🙏 🙏 🙏

packages/api/src/utils/parser.ts
```diff
@@ -239,7 +239,9 @@ export const parsePreparedContent = async (
     // to the handlers, and have some concept of postHandle
     if (article?.content) {
       const articleDom = parseHTML(article.content).document
-      const codeBlocks = articleDom.querySelectorAll('code')
+      const codeBlocks = articleDom.querySelectorAll(
```
jacksonh, line 242: nice

jacksonh (approved):
This all looks good. A couple scary parts with changing the JS executed on fetch and changing `document` to `preparedDocument`

packages/puppeteer-parse/index.js
```diff
@@ -409,32 +409,6 @@ function getUrl(req) {
...
-    'connect.facebook.com',
-    'connect.facebook.net',
-    'sp.analytics.yahoo.com',
-  ]
-
-  await client.send('Network.setBlockedURLs', { urls: blockedResources });
-}
-
 async function retrievePage(url, logRecord, functionStartTime) {
   validateUrlString(url);
```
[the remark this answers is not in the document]
sywhb: Yeah, but I think the `adblocker` would take care of it so I dropped this line. It should be okay

PR #1817, opened 2023-02-16
Library Filters - Android

satindar:
- track scrolling from webview rather than webview container
- use web scroll changes to change top app bar height
- multiply x and y tap coordinates by density to get correct bar positioning
- make sure nav bar is always shown when page first loads
- bump android build to 0.20
- use clientx/y in highlight layer
- load web reader via activity rather than composable
- avoid calling web view load twice
- animate webview alpha changes
- load bg colro of webview based on theme
- clear room db when logging out on android
- update isRefreshing value after sync op is complete
- add refresh button to library nav ar
- Use Instant for storing sync time as this should always be in UTC
- remove refresh buttin from library nav bar
- gst
- add context menu to show under saved item filter dropdown
- update filter in library view model when user changes it


jacksonh:
wow monster PR

jacksonh (approved):

PR #1822, opened 2023-02-17
New iOS developer experience

sixten:
A couple of small changes to make it easier to check out and build and run the iOS app.


satindar (approved):
Thanks for the PR! This will make it easier for the next dev that builds the iOS app 👍🏽

PR #1825, opened 2023-02-18
Snooze calculation

sixten:
The snooze tests were failing for me. Upon closer inspection, this appears to be because the strings in the test cases specified a UTC offset of `-08:00`, which is not the time zone I'm in.

To fix this, I've refactored the method that calculates the target dates to take a date and a calendar as arguments. The app code continues to use `Calendar.current`, while the test code now specifically passes a Gregorian calendar with a matching time zone. While still nowhere near exhaustive, I've added a couple more test cases in other time zones.

The test changes are more clear if you look at the commits separately, rather than the branch-wise diff.

### DIFF
--- a/apple/OmnivoreKit/Sources/Views/SnoozeView.swift
+++ b/apple/OmnivoreKit/Sources/Views/SnoozeView.swift
@@ -21,15 +21,15 @@ public struct SnoozeView: View {
       Spacer()
 
       HStack {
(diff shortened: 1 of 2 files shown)


jacksonh:
hah, thank you. Classic unit testing bug.

jacksonh (approved):

Issue #1094, opened 2022-08-18
Delete MP3 and PDF files when clearing cache on iOS


jacksonh:
When the user chooses to clear their cache we should delete the MP3 and PDF files that have been downloaded.

sixten:
Does this include both the temporary files in the caches directory, and the more permanent copies that are written to their Documents directory?

PR #1844, opened 2023-02-27
Clear local files when reseting local storage

sixten:
This PR implements #1094, removing PDF and audio files from the caches and Documents directories when reseting the app's local storage.

There are two other changes here:

1. I've added a wrapper around the new `URL` directory APIs in iOS 16/macOS 13, and refactored some existing file path code to use it.
1. I've reworked `DataService.lastItemSyncTime` not to use the `@AppStorage` attribute, so that the serialization logic for the date is now hidden in the property implementation. It doesn't appear that anything in the app was depending on the property publishing its value, so I don't think this breaks anything. But it would be straightforward to revert the change.

### DIFF
--- a/apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
+++ b/apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
@@ -6,8 +6,6 @@ import Utils
 import Views
 
 @MainActor final class HomeFeedViewModel: NSObject, ObservableObject {
-  let dateFormatter = DateFormatter.formatterISO8601
-
   var currentDetailViewModel: LinkItemDetailViewModel?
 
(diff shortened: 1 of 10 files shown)


apple/OmnivoreKit/Sources/App/Views/Home/HomeFeedViewModel.swift
```diff
@@ -110,7 +108,7 @@ import Views
 
   func syncItems(dataService: DataService) async {
     let syncStart = Date.now
-    let lastSyncDate = dateFormatter.date(from: dataService.lastItemSyncTime) ?? Date(timeIntervalSinceReferenceDate: 0)
+    let lastSyncDate = dataService.lastItemSyncTime
```
satindar, line 111: :nice:

satindar (approved):
Looks good! Much cleaner at the call sites too. Thanks for taking care of the issue 🙌🏽

PR #1852, opened 2023-03-02
Add publishedAt to the updatePageInput

sywhb:


packages/api/src/resolvers/update/index.ts
```diff
@@ -41,6 +41,7 @@ export const updatePageResolver = authorized<
     description: input.description ?? undefined,
     author: input.byline ?? undefined,
     savedAt: input.savedAt ? new Date(input.savedAt) : undefined,
+    publishedAt: input.publishedAt ? new Date(input.publishedAt) : undefined,
```
jacksonh, line 44: 👍

jacksonh (approved):

PR #1855, opened 2023-03-06
improve urlParse for puppeteer-parse

kristol07:
**Background**: When use Android app, for links shared from some apps, links are not pure url but with additional text such as article title.
Example: "The Missing Semester of Your CS Education - https://new.ycombinator.com/item?id=34934216"

**Solution**: To improve user experience, instead of just returning parse error, it could be better to try to parse the urlStr first and check whether any valid url. So I add a check with regex before that, the code change is rather simple and understandable.

Thanks for any comments, this is my first PR.

### DIFF
--- a/packages/puppeteer-parse/index.js
+++ b/packages/puppeteer-parse/index.js
@@ -402,13 +402,13 @@ function tryParseUrl(urlStr) {
     return null;
   }
   
-  // a regular expression to match URLs
-  var regex = /(https?:\/\/[^\s]+)/g;
+  // a regular expression to match all URLs
+  const regex = /(https?:\/\/[^\s]+)/g;
   
-  var match = regex.exec(urlStr);
+  const matches = urlStr.match(regex);
   
-  if (match) {
-    return match[1];
+  if (matches) {
+    return matches[0]; // only return first match
   } else {
     return null;
   }
--- a/packages/puppeteer-parse/index.js
+++ b/packages/puppeteer-parse/index.js
@@ -397,15 +397,33 @@ function validateUrlString(url) {
   }
 }
 
+function tryParseUrl(urlStr) {
+  if (!urlStr) {
+    return null;
+  }
+  
+  // a regular expression to match URLs
+  var regex = /(https?:\/\/[^\s]+)/g;
+  
+  var match = regex.exec(urlStr);
+  
+  if (match) {
+    return match[1];
+  } else {
+    return null;
+  }
+}
+
 function getUrl(req) {
   const urlStr = (req.query ? req.query.url : undefined) || (req.body ? req.body.url : undefined);
-  if (!urlStr) {
+  const url = tryParseUrl(urlStr)
+  if (!url) {
     throw new Error('No URL specified');
   }
 
-  validateUrlString(urlStr);
+  validateUrlString(url);
 
-  const parsed = Url.parse(urlStr);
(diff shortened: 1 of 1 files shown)


jacksonh:
Ah thank you this looks good, we have something similar in iOS 🚀 

I'll wait on @satindar to merge as he has a big branch going right now.

satindar:
> Ah thank you this looks good, we have something similar in iOS 🚀
> 
> I'll wait on @satindar to merge as he has a big branch going right now.

I can easily rebase my branch after this is in 👍🏽

satindar:
Thanks @kristol07 for bringing this to our attention and for this PR. We'll try to fix this client side as well. Will let @sywhb take a look here before merging.

kristol07:
> Thanks @kristol07 for bringing this to our attention and for this PR. We'll try to fix this client side as well. Will let @sywhb take a look here before merging.

Sure, I just found one small bug by myself, sorry about that even though I actually double checked it ==

@jacksonh @satindar However, this brings another question. I didn't run it before commit, because I couldn't...
When try to build the dev environment, following steps in the current doc just doesn't work. Setting environment variables, removing some strange configuration in docker compose file... and when I finally build the frontend, I don't know how to create account and log in, so just give up for now. It could be great if we can have more detailed documentation on this.
I asked one question in discord before, but maybe buried by too many other messages. If I could run in my local machine, I'd like to contribute more as I just like it and use it everyday ;-) so here thank you again for your great effort!

packages/puppeteer-parse/index.js
```diff
@@ -397,15 +397,33 @@ function validateUrlString(url) {
   }
 }
 
+function tryParseUrl(urlStr) {
+  if (!urlStr) {
+    return null;
+  }
+  
+  // a regular expression to match URLs
+  var regex = /(https?:\/\/[^\s]+)/g;
```
sywhb, line 406: We might need to remove `g` in the regex as it won't include capture groups in matches

packages/puppeteer-parse/index.js
```diff
@@ -397,15 +397,33 @@ function validateUrlString(url) {
...
+  
+  var match = regex.exec(urlStr);
+  
+  if (match) {
+    return match[1];
```
sywhb, line 411: We probably should use `String.prototype.match()` method since we want to fetch all the matches at one go

sywhb (changes requested):

packages/puppeteer-parse/index.js
kristol07: Thanks for comment! yes, no need to get details of each match with loop, I updated it to use `match` method, and only return first match, what do you think?

packages/puppeteer-parse/index.js
kristol07: Thanks for comment, I prefer to keep `g` here, see different regex expr with different results below: 

image

I think if we need any change in the future, this expression is less likely to be updated.

kristol07: `var` is also changed to `const`

sywhb: Yeah, good point!

packages/puppeteer-parse/index.js
sywhb: No problem at all. I think this good to 🚢

sywhb (approved):
LGTM. Really appreciate @kristol07 !

PR #1859, opened 2023-03-06
Switch express-http-context dependency to express-http-context2

artberri:
`express-http-context2` is a drop-in replacement for `express-http-context` that uses the more performant `async_hooks` Node.js native module instead of the (nowadays unnecessary) `cls-hooked` community provided CLS library. ([More info](https://itnext.io/one-node-js-cls-api-to-rule-them-all-1670ac66a9e8)).

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -55,7 +55,7 @@
     "dot-case": "^3.0.4",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
-    "express-http-context": "^1.2.4",
+    "express-http-context2": "^1.0.0",
     "express-rate-limit": "^6.3.0",
     "firebase-admin": "^10.0.2",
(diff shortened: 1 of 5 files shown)


sywhb:
Hey @artberri, nice PR. @jacksonh I don't recall why I used `express-http-context` lib in the first place because we already have the Apollo Express Context. This looks like something we should put in the Apollo Context too

jacksonh:
Maybe we need it for some of the REST methods? I know the mobile auth methods are REST based and might need the client info. Maybe for adding the iOS specific onboarding articles.

sywhb (approved):
LGTM

sywhb:
I will rebase and merge this

sywhb:
Close this one because of too many conflicts and created a new one here: #2028

PR #1863, opened 2023-03-07
Forward Gmail verification email to the user

sywhb:


packages/inbound-email-handler/src/index.ts
```diff
@@ -141,7 +141,6 @@ export const inboundEmailHandler = Sentry.GCPFunction.wrapHttpFunction(
         if (isConfirmationEmail(from, subject)) {
           console.log('handleConfirmation', from)
           await handleConfirmation(to, subject)
-          return res.send('ok')
```
sywhb, line 144: Removed this line so it will continue to forward the email

jacksonh (approved):

PR #1864, opened 2023-03-07
Add LABEL_ADDED as event type in webhooks

sywhb:


jacksonh:
Hey @sywhb sorry I can't tell from this, but this will be fired when a label is added to a page?

sywhb:
@jacksonh, yeah, it should, because we publish `LABEL_CREATED` event to the pubsub queue when there is a label added to some page.

It's the best to test it in demo first

jacksonh:
Ah thanks, i wasn't quite sure if `LABEL_CREATED` was when a label was first created.

sywhb:
> Ah thanks, i wasn't quite sure if `LABEL_CREATED` was when a label was first created.

Haha, sounds weird right? I, at that time, thought there should be such event for `label_added` instead of `label_created` but probably we should have both 👍

jacksonh (approved):
LETS GO!

Issue #1854, opened 2023-03-03
Webhook event type for labels


anseljh:
This is discussed a bit in passing here: https://github.com/omnivore-app/omnivore/issues/690#issuecomment-1135369409

> I don't know how this would work, because labels are added later, but I could see some workflow where you might not have something sent until a label is added. Like say I add an Omnivore label to a page, then the web hook would be triggered.

The idea is to have a type of webhook that will fire if a specified label is added to an article. For example, I might want to write something that sends an article to my Kindle if an article gets a "Send to Kindle" label.

UI wise, I guess you'd add a `LABEL_ADDED` event type to this list:

![image](https://user-images.githubusercontent.com/2075772/222629643-c01e6031-7674-4f5d-845d-03065e114a67.png)

...and then you'd need to have a label selector.

jacksonh:
Hey, thanks, actually this fits in pretty well with how we implemented our rules engine under the hood it uses a search filter. I think we could do the same thing here, so you'd add an optional search filter like `event.created label:"<my label>"`

sywhb:
Yeah, we actually have the `LABEL_ADDED` event type available in the code

jacksonh:
Oh yes, we could just use that instead of using a search filter

sywhb:
#1864

sywhb:
Hey @anseljh, we now made the `LABEL_ADDED` event type available in the webhooks page

anseljh:
I got it working. Rad! Now to figure out the GraphQL API...

PR #1882, opened 2023-03-09
Replace LABEL_CREATED with LABEL_ADDED in webhooks page

sywhb:
- Change LABEL_CREATED to LABEL_ADDED in webhook events label
- Remove unique constraint of user_id and event_types on webhooks table
- Allow webhooks with same event types
- Add checking for the emptiness of event types


sywhb:
#1854

jacksonh:
hah nice!

Issue #1837, opened 2023-02-22
Date_published is a bit late


StrollStars:
The date_published of saved article is dated one day after the date_published of original.
![2023-02-22_215225](https://user-images.githubusercontent.com/10942225/220641720-b3c7f591-8f77-4e9d-bef9-d62b8a05ac11.png)
![2023-02-22_215139](https://user-images.githubusercontent.com/10942225/220641745-13500e83-1f10-4dc2-91b2-002bc0b20b51.png)

jacksonh:
I wonder if this is the UTC date, not your local time (assuming you are in Asia time, so sometimes ahead of UTC by a day). We should format in local time in the reader for sure

StrollStars:
@jacksonh I find that the website in the screenshot uses the local time. Most of our websites here use this time. The local time is UTC+8, and the difference is exactly 8 hours, so can it be adjusted?

sywhb:
Hey @StrollStars, this has been fixed and you should see the published and saved date in your local timezone now (maybe need to clear cache)

StrollStars:
@sywhb ,thank you for responding to me.
I checked the existing articles and found that the dates of the most recent articles have been changed to local dates, but the dates of the earlier articles are still UTC dates.
I bookmarked an article through the Chrome extension, and found that it is still a UTC date.
BTW, I have cleared the cache.

sywhb:
#1969

jacksonh:
Hey @StrollStars except in extreme cases we don't override historical data when making updates as we don't want to "mess" with people's libraries, so this change will only take effect on newly saved items.

StrollStars:
@jacksonh Thank you for your reply, I will manually modify the saved articles first, hoping to fix this problem as soon as possible.

StrollStars:
@jacksonh @sywhb The latest testing has discovered that the issue of the publishing date being eight hours late on the Android client has been resolved, but it still persists on the Chrome extension. However, the problems of duplicated titles and missing publishing dates and times continue to exist on both the Android client and the Chrome extension.

sywhb:
Hey @StrollStars , sorry for the delay.

We have fixed "the problems of duplicated titles and missing publishing dates and times" in the backend so you can now get them right on your "Android client and Chrome extension".

For "the issue of the publishing date being eight hours late" on the Chrome extension, although we have fixed it, we are still waiting for the new release of the extension.

I shall let you know once it's released. Thanks for the patience!

[...]

PR #1902, opened 2023-03-13
return published date if the class name is omnivore-published-date which we added when we scraped the article

sywhb:


sywhb:
Hey @jacksonh, I created a special `<div>` to contain the publication date of WeChat articles and handled it in Readability because we the current one we parsed does not have the correct timezone referring to #1837

jacksonh:
Nice hack. Just wondering, would it be easier to return `create_date` if not null in `_checkPublishedDate` instead of inserting the div?

sywhb:
@jacksonh because I'm not sure if the `create_time` could be accessed from Readability

jacksonh (approved):

sywhb:
You mean using `<script>` tag to call the javascript?

PR #1905, opened 2023-03-13
Add type to the highlight in the elasticsearch and three types are HIGHLIGHTS, REDACTION and NOTE

sywhb:


jacksonh (commented):
Question on the consequences of this: if a user adds a few redactions now, and then syncs with Logseq (or an old version of the iOS app), would the REDACTIONS be included in the list of highlights?

sywhb:
> Question on the consequences of this: if a user adds a few redactions now, and then syncs with Logseq (or an old version of the iOS app), would the REDACTIONS be included in the list of highlights?

Yes, but i don't know if it matters because the `quote` in redactions is `null`

jacksonh (approved):

Issue #1911, opened 2023-03-14
trim whitespace text from user email addresses


jacksonh:
When a user creates an account, its possible for them to create it with a space after their name. We could probably prevent this on the frontend, but also might want to on the backend.

sywhb:
This is fixed in Demo env

PR #1912, opened 2023-03-14
Trim whitespace text from user email address

sywhb:


sywhb:
#1911

packages/api/src/routers/auth/auth_router.ts
```diff
@@ -373,18 +374,27 @@ export function authRouter() {
...
         )
       }
-
+      const { email, password } = req.body
       try {
-        const models = initModels(kx, false)
-        const user = await models.user.getWhere({
-          email,
+        const user = await getRepository(User).findOneBy({
+          email: ILike(email.trim()), // case insensitive
```
jacksonh, line 397: Do we need to use ILike here? I worry that maybe it could do partial matches?

packages/api/src/routers/auth/auth_router.ts
```diff
@@ -437,25 +446,43 @@ export function authRouter() {
...
+      if (!isValidSignupRequest(req.body)) {
         return res.redirect(
           `${env.client.url}/auth/email-signup?errorCodes=INVALID_CREDENTIALS`
         )
       }
-      const lowerCasedUsername = username.toLowerCase()
-
+      const { email, password, name, username, bio, pictureUrl } = req.body
+      // trim whitespace in email address
+      const trimmedEmail = email.trim()
```
jacksonh, line 476: nice

packages/api/src/routers/auth/auth_router.ts
sywhb: Yeah, partial match could be a problem.

I'm trying to do a case insensitive match of the email address. Probably having`lower(col) = lower(value)` in the query is better

@jacksonh

jacksonh: Oh yes, i think using lower is kind of the standard for this.

jacksonh (approved):

Issue #1908, opened 2023-03-14
Allow saving HTML with highlights


jacksonh:
When we save a highlight we should be able to include an optional html attribute, the backend shouldn't need to do any special handling on this. We should also expose it as part of the `Highlight` response type.

sywhb:
Should we make the `html` searchable or it is just used to display?

jacksonh:
@sywhb i think it would just be used for presentation, the main use case I can think of is it there is useful markup like a list or image that the user would want embedded, maybe in the logseq plugin (I know this could be hard) or just display in Omnivore highlight cards.

jacksonh:
I'll move this over to a web task now

PR #1914, opened 2023-03-14
save html with highlights

sywhb:
- Add html to the highlight document in elasticsearch
- Save html along with highlight and return


sywhb:
#1908

jacksonh (approved):

sywhb:
> /giphy waynes world "we're not worthy"

Issue #1907, opened 2023-03-14
Allow scoping searches to `site:`


jacksonh:
Ideally we should allow both domain and sitename in this scope, such as:

`site:cnn.com` 
`site:CNN` (the siteName returned by their metadata)

sywhb:
<img width="411" alt="Screenshot 2023-03-14 at 6 28 43 PM" src="https://user-images.githubusercontent.com/12963965/224972563-0405d840-d8c0-49cf-aeb6-3c4be408413b.png">
<img width="446" alt="Screenshot 2023-03-14 at 6 28 37 PM" src="https://user-images.githubusercontent.com/12963965/224972589-931eccfe-356d-4a7e-974f-3d4fc002c3c9.png">
<img width="737" alt="Screenshot 2023-03-14 at 6 28 28 PM" src="https://user-images.githubusercontent.com/12963965/224972596-04d624b9-6903-4d14-abe3-52d6bdb66c72.png">
<img width="731" alt="Screenshot 2023-03-14 at 6 28 21 PM" src="https://user-images.githubusercontent.com/12963965/224972604-2e0803cc-5cea-42e2-aeb5-6b28c165447c.png">

Some samples from my local machine

sywhb:
<img width="807" alt="Screenshot 2023-03-15 at 7 59 30 PM" src="https://user-images.githubusercontent.com/12963965/225302797-c15299e7-a7ff-44b0-98c0-4f9d13ed12a2.png">

Chinese in the site name

jacksonh:
that is fantastic. Will be really useful when combined with Saved Searches

sywhb:
Speaking of saved searches, we have the category API available if we want to add that to the UI

jacksonh:
Yeah once we have a more scalable system for forms we should add it. Getting close on that stuff.

PR #1915, opened 2023-03-14
Allow scoping searches to site:

sywhb:


sywhb:
#1907

sywhb:
Because `url` is of `keyword` type in elasticsearch so we could not search for the domain and do a `match` query against it.
I'm thinking about adding a `domain` or `hostname` of `text` type in the `page` document in elasticsearch to make it searchable so we could do a `multi_match` against both `siteName` and `domain`

sywhb:
> Because `url` is of `keyword` type in elasticsearch so we could not search for the domain (url is not tokenized) and do a `match` query against it.
> 
> I'm thinking about adding a `domain` or `hostname` of `text` type in the `page` document in elasticsearch to make it searchable so we could do a `multi_match` against both `siteName` and `domain`

Resolved this with `wildcard` query:
```
{
    should: [
      {
        match: {
          siteName,
        },
      },
      {
        wildcard: {
          // siteName is a domain name, so we need to wildcard the end
          url: `*${siteName}*`,
        },
      },
    ],
    minimum_should_match: 1,
  }```

**Comment by sywhb** [comment]
This is ready

**Comment by sywhb** [comment]
Fallback to hostname if site name is `null`

File: packages/api/.eslintignore [hunk 1]
```diff
@@ -3,3 +3,4 @@ dist/
 readabilityjs/
 src/generated/
 test/resolvers/
+apollo.config.js
```
sywhb, line 6: For Apollo extension to work in VSCode

jacksonh: nice. It only works in dev mode, right?

jacksonh (commented):
One question on this: i saw the `siteName` added to tests, was this needed because something would fail if siteName is null?

jacksonh (approved):

packages/api/.eslintignore
sywhb: yeah, but actually i don't really need to add this line because `apollo.config.js` is a Javascript file

sywhb:
> One question on this: i saw the `siteName` added to tests, was this needed because something would fail if siteName is null?

No, i think it might be better to have the host or domain name as the default `siteName` and these tests would fail because their `siteName` are `null` from the previous generation

PR #1921, opened 2023-03-17
Return correct page slug and get saving request by url

sywhb:
- Return existing slug if page exists
- Get saving request by url
- Add test case
- Frontend changes

### DIFF
--- a/packages/api/src/generated/graphql.ts
+++ b/packages/api/src/generated/graphql.ts
@@ -168,6 +168,7 @@ export type ArticleSavingRequest = {
   slug: Scalars['String'];
   status: ArticleSavingRequestStatus;
   updatedAt: Scalars['Date'];
+  url: Scalars['String'];
   user: User;
   /** @deprecated userId has been replaced with user */
   userId: Scalars['ID'];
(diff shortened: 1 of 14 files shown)


packages/api/src/generated/graphql.ts
```diff
@@ -1751,7 +1752,7 @@ export type QueryArticleArgs = {
 
 
 export type QueryArticleSavingRequestArgs = {
-  id: Scalars['ID'];
+  url: Scalars['String'];
```
jacksonh, line 1755: Could we allow both and make them optional? I'm just trying to think of how we'd have to roll this out. If only the web calls this, i guess we can deploy web at the right time.

jacksonh (approved):

jacksonh:
I think only web uses this, I'm verifying that now. If so, lets just plan to update web at the right moment.

packages/api/src/generated/graphql.ts
sywhb: Yes, so the frontend won't be messed up.

I will make the change

sywhb:
@jacksonh I have made the changes to the Web too.

Could we deploy the backend first but not split the traffic until the web is deployed?

jacksonh:
I think the traffic split might just be more complicated than needed, as the `/sr/` redirect doesn't work great right now anyways, we can probably time the changes to minimize the "down" time of this page.

PR #1932, opened 2023-03-20
use url in save request

sywhb:
- Return url in save API
- Return article's url in saveUrl api
- Use url in save request

### DIFF
--- a/packages/api/src/services/save_url.ts
+++ b/packages/api/src/services/save_url.ts
@@ -1,6 +1,5 @@
 import { PubsubClient } from '../datalayer/pubsub'
 import { UserData } from '../datalayer/user/model'
-import { homePageURL } from '../env'
 import { SaveErrorCode, SaveResult, SaveUrlInput } from '../generated/graphql'
 import { DataModels } from '../resolvers/types'
 import { createPageSaveRequest } from './create_page_save_request'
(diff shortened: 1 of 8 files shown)


packages/api/src/services/save_url.ts
```diff
@@ -26,9 +25,7 @@ export const saveUrl = async (
 
     return {
       clientRequestId: pageSaveRequest.id,
-      url: `${homePageURL()}/${saver.profile.username}/links/${
```
jacksonh, line 29: My above comment was actually for this line

jacksonh (changes requested):
I think this is good except for the `save_url.ts` change which I don't think we want.

packages/web/components/templates/homeFeed/AddLinkModal.tsx
```diff
@@ -23,9 +20,7 @@ export function AddLinkModal(props: AddLinkModalProps): JSX.Element {
   const handleLinkSubmission = useCallback(
     async (link: string) => {
       const result = await saveUrlMutation(link)
-      // const result = await saveUrlMutation(link)
-      if (result && result.jobId) {
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      if (result && result.url) {
```
jacksonh, line 23: oh i guess this should always use link too

PR #1939, opened 2023-03-21
fix/deleted article saving request

sywhb:
- Add BAD_DATA to the saving request error code
- Allow array values in page filters
- Do not fetch deleted article in saving request
- Do not check for deleted article in saving request in web
- Reset state to processing when creating articles


packages/api/src/resolvers/article/index.ts
```diff
@@ -649,24 +645,18 @@ export const setBookmarkArticleResolver = authorized<
     { input: { articleID, bookmark } },
     { claims: { uid }, log, pubsub }
   ) => {
-    const page = await getPageById(articleID)
+    const page = await getPageByParam({
```
jacksonh, line 648: 👍

jacksonh (approved):

PR #1946, opened 2023-03-23
Use receipient address in the envelope field

sywhb:


sywhb:
Reference from https://campus.barracuda.com/product/emailgatewaydefense/doc/96022980/understanding-the-message-header-and-envelope/#:~:text=To%20understand%20how%20mail%20is,the%20mail%20is%20being%20delivered.

packages/inbound-email-handler/src/index.ts
```diff
@@ -68,6 +68,19 @@ const saveReceivedEmail = async (
...
 }
 
+export const parsedTo = (parsed: Record<string, string>): string => {
+  // envelope to contains the real recipient email address
+  try {
+    const envelope = JSON.parse(parsed.envelope) as Record<
+      string,
+      string | string[]
+    >
+    return envelope.to.toString().split(',')[0] || parsed.to
```
jacksonh, line 78: What is the split for here, looking at the sample envelope I saw it was like this `{"to":["jacksonh-XXXXXXXX@inbox.omnivore.app"]`

sywhb: yeah, could be removed. i was considering the case where `to` is a string instead of an array

jacksonh: Oh, I see, so there can be multiple? Do you know if sendgrid always handles that?

sywhb: Screenshot 2023-03-23 at 6 31 01 PM

sywhb: yeah

sywhb: I updated the code to use the first element in the array

jacksonh (approved):

PR #1969, opened 2023-04-03
Get published date of wechat article in extension

sywhb:


pkg/extension/src/scripts/content/prepare-content.js
```diff
@@ -136,7 +136,15 @@
     document.documentElement.appendChild(contentCopyEl);
 
     Array.from(contentCopyEl.getElementsByTagName('*')).forEach(prepareContentPostItem);
-
+    // check if create_time is defined
```
jacksonh, line 139: We probably should do this in an exception handler, just in case the value is defined to something that isn't a date.

sywhb: I wrapped the function in an exception handler @jacksonh

jacksonh:
Will have to do another extension update in the stores. Might try to add `Add Note` functionality first.

jacksonh (approved):

sywhb:
@jacksonh I created another PR to fix the infinite scrolling issue for news websites: #1983 

We could include that one in the new extension release too

PR #1980, opened 2023-04-05
Stop inserting search history to the DB in the search API

sywhb:


jacksonh:
Will search history API still work?

sywhb:
> Will search history API still work?

No, it won't work but I don't think we are calling the search history API in the client now

jacksonh:
Yeah I suppose its fine for the clients to cache this locally for search history. It means it isn't synced across devices, but I don't know if that is a big concern

jacksonh (approved):

sywhb:
I agree with you @jacksonh

The local search history works great so far

sywhb:
@jacksonh I set a limit on the search query size to 255 characters. Is that enough?

packages/api/src/schema.ts
```diff
@@ -2542,7 +2542,7 @@ const schema = gql`
     search(
       after: String
       first: Int
-      query: String
+      query: String @sanitize(maxLength: 255)
```
sywhb, line 2545: In case you missed the change here @jacksonh

PR #1985, opened 2023-04-05
When deleting highlights finish core data sync before triggering API sync

jacksonh:


satindar (approved):
This looks like it should fix any threading issues. I probably need to do something similar on android too. I think we also have an issue on Android where we get a 429 error from the server but haven't been able to confirm that yet. One user said only the first highlight they create gets synced after going online which might happen if we get a 429.

I'll do some testing to confirm this change works.

PR #1998, opened 2023-04-06
iOS Reader Settings display updates and theme updates

jacksonh:
- WIP: update the display settings view on iOS
- Implement margins and iPad reader preferences options
- Faster audio playback rates
- More speeds
- Add handler for justify text, fix dependency issue with auto highlights
- Get slider closer to design


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/1999

Issue #2012, opened 2023-04-10
"Edit Labels" button not working for PDF


sywhb:
<img width="380" alt="Screenshot 2023-04-10 at 4 05 24 PM" src="https://user-images.githubusercontent.com/12963965/230856955-6d8cfde9-8159-423c-9a87-bb81e3be7d78.png">

It doesn't respond when the button is clicked

Only for PDF

jacksonh:
@sywhb you mean not working for non-PDF, right?

sywhb:
> @sywhb you mean not working for non-PDF, right?

@jacksonh I meant for PDF. We don't have the "Edit Labels" button for non-PDF

jacksonh:
oh, on smaller screens it will show up for Web. I think I have fixed for both now.

sywhb:
For instance

<img width="1727" alt="Screenshot 2023-04-10 at 6 55 24 PM" src="https://user-images.githubusercontent.com/12963965/230888827-3f33a564-cc86-406c-864b-5e451f41b361.png">

sywhb:
Thanks. I just tested and close this issue now

PR #2013, opened 2023-04-10
Fix edit labels drop down action

jacksonh:


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/2012

jacksonh:
Some extra junk in this. Will cleanup

PR #2017, opened 2023-04-11
If a lazy image src is specified use that instead of the default src

jacksonh:


jacksonh:
blah prettier made a mess of this one. Going to close and re-open.

PR #2025, opened 2023-04-12
Add TWEET, VIDEO and IMAGE pageType

sywhb:
- Add tweet, video, picture pageType
- Replace picture with image


packages/api/src/utils/parser.ts
```diff
@@ -118,12 +118,20 @@ const parseOriginalContent = (document: Document): PageType => {
         return PageType.Profile
       case 'website':
         return PageType.Website
+      case 'tweet':
```
jacksonh, line 121: Are these defined by opengraph? (Don't think it matters if they aren't, it will just default to unknown and then the user can set)

sywhb: No, I set these in the content_handlers @jacksonh

jacksonh: ohhh smart. Nice work!

Issue #2040, opened 2023-04-13
[Feature] Add more keyboard shortucts


Bad3r:
Please consider adding shortcuts to open the popups from the left menu:
![image](https://user-images.githubusercontent.com/25513724/231814038-a0bacbb8-2fa5-4205-a436-44ed3d7956ed.png)

Example:
- `Alt` `L` opens the labels popup menu

PR #2043, opened 2023-04-14
Clean up keyboard commands functionalities and modal, add tooltips

jacksonh:


jacksonh:
https://github.com/omnivore-app/omnivore/issues/2040

PR #2063, opened 2023-04-17
Upgrade natural nodejs module and use a new sentence tokenizer based on parsing

sywhb:


jacksonh:
@sywhb so do you think we can use different libraries based on language here?

sywhb:
> @sywhb so do you think we can use different libraries based on language here?

@jacksonh i haven't found a nice tokenizer library for CJK languages yet. 

I think we can choose the library based on language specified in the SSML if we found a good library for CJK too as this one is pretty good for English, German and English-like languages

Issue #2007, opened 2023-04-10
Return ContentReader.EPub for epub files


jacksonh:
The search, updates, and article APIs have a `ContentReader` property. Right now we only support Article and PDF. We need to add an epub option, and then update the APIs to return epub for those upload types. 

We might need to track this as contentType on pages in elastic to rely on using PageType.BOOK for the epub content reader.

Issue #2006, opened 2023-04-10
Support for uploading epubs


jacksonh:
The first step in handling epubs is upload support. The `uploadFileRequestResolver` makes a few assumptions about content type, setting everything to `PDF`. There is a contentType value passed to the resolver, but a few places assume that the uploaded content is a .pdf file (just search the source file for PDF).

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

Issue #2008, opened 2023-04-10
Figure out how to do search indexing on epubs


jacksonh:
Right now a background process detects uploaded PDF content, and triggers our pdf-handler, this parses the PDF for content and calls an API in our backend so it can be added to the search index.

We will need to do the same to support epubs.

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #2069, opened 2023-04-18
Support additional content reader types

jacksonh:
- Implement a reader for epubs
- Add epub container

### DIFF
--- a/packages/api/src/services/speech.ts
+++ b/packages/api/src/services/speech.ts
@@ -1,4 +1,6 @@
 import { Page, PageType } from '../elastic/types'
+import { ContentReader } from '../generated/graphql'
+import { contentReaderForPageType } from '../utils/uploads'
 import { FeatureName, isOptedIn } from './features'
 
 /*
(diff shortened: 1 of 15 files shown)


packages/api/src/services/speech.ts
```diff
@@ -8,7 +10,10 @@ export const shouldSynthesize = async (
   userId: string,
   page: Page
 ): Promise<boolean> => {
-  if (page.pageType === PageType.File || !page.content) {
+  if (
+    contentReaderForPageType(page.pageType) === ContentReader.Web ||
```
sywhb, line 14: I think this should be `!==`

sywhb (commented):
I think only the email attachment change is missing. I can add it to your branch too

packages/api/src/services/speech.ts
jacksonh: OMG that would have been hard to catch in testing too

jacksonh:
> I think only the email attachment change is missing. I can add it to your branch too

yeah, we could probably do separately also, I think we will also need to create a `epub-handler` to do the search indexing

sywhb:
@jacksonh yeah, `epub-handler` is another one.

Cool. LGTM

sywhb (approved):

jacksonh:
Tagged this ticket which kind of covers an epub handler: https://github.com/omnivore-app/omnivore/issues/2008

jacksonh:
Closes https://github.com/omnivore-app/omnivore/issues/2007 
Closes https://github.com/omnivore-app/omnivore/issues/2006

PR #2072, opened 2023-04-18
Fix labels not being displayed on the reader

jacksonh:


jacksonh:
Fixes: https://github.com/omnivore-app/omnivore/issues/2041

Issue #2081, opened 2023-04-19
Allow users to delete a Note


sywhb:
Note is empty
![](https://cdn.discordapp.com/attachments/1084693944861982791/1097996251116601456/Screen_Shot_2023-04-18_at_20.33.55.png)

The count of highlights is still 1
![](https://cdn.discordapp.com/attachments/1084693944861982791/1097996250906898612/Screen_Shot_2023-04-19_at_00.22.34.png)

Maybe what we can do is to add another button for the deletion or delete the note when the content is empty

jacksonh:
Oh actually we have this in the dropdown but maybe setting to empty string should delete it too. (From the client, would be weird for the backend to do that I think.) 

<img width="496" alt="Screen Shot 2023-04-19 at 09 51 57" src="https://user-images.githubusercontent.com/75189/232946074-eb5d0eb7-31a7-4f11-bff3-fe4231a845aa.png">

sywhb:
Oh I didn't notice. The button seems not working too @jacksonh

jacksonh:
Fixed

PR #2082, opened 2023-04-19
Fix delete article note not being displayed

jacksonh:


jacksonh:
Fixes: https://github.com/omnivore-app/omnivore/issues/2081

PR #2088, opened 2023-04-20
Remove build script because cloud-sdk will call it when deploying cloud functions

sywhb:


jacksonh (approved):

sywhb:
Close this PR as `lerna` also uses `build` script

i will build the docker image instead in cloud build and deploy the docker image in cloud run service

@jacksonh

Issue #1829, opened 2023-02-20
Add URL via email


sorenpeter:
It would be nice if you just send an URL to your omnivore email that it would work just like adding an URL from the app. 
The URL should be in the subject of the mail so it does not get confused.

jacksonh:
Good idea. I guess we can just check the subject and see if its a parsable URL, if so queue a SaveURL operation.

sorenpeter:
Yes that was exactly how I imagined would be the way to implement this.

sywhb:
Hey @sorenpeter, we have implemented this feature and you could save an URL by sending an email to your [Omnivore inbox's email address](https://omnivore.app/settings/emails) and use the URL as the subject.

sorenpeter:
Hi, That sounds awesome and it seem to work as expected too. Than you!

sywhb:
Thanks @sorenpeter for the verification. I will close this for now

PR #2101, opened 2023-04-24
feat/add url via email

sywhb:
- Eagerly load profile of the user
- Save url if the email subject is a parsable url


sywhb:
fixes: #1829

PR #2109, opened 2023-04-25
Better tweet titles

jacksonh:
- Better formatting of tweet titles
- Remove aabs from git

### DIFF
--- a/packages/content-handler/src/websites/twitter-handler.ts
+++ b/packages/content-handler/src/websites/twitter-handler.ts
@@ -2,6 +2,7 @@ import { ContentHandler, PreHandleResult } from '../content-handler'
 import axios from 'axios'
 import { DateTime } from 'luxon'
 import _ from 'underscore'
+import { truncate } from 'lodash'
 import { Browser, BrowserContext } from 'puppeteer-core'
 
 interface TweetIncludes {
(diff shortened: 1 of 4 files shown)


jacksonh:
This gives us titles like "Jackson Harper on Twitter: ...." the UI will ellipses the text. Also we remove URLs from the text, because usually tweets with pictures or embedded content will have URLs in the text.

packages/content-handler/src/websites/twitter-handler.ts
```diff
@@ -134,8 +134,8 @@ const getTweetsByIds = async (ids: string[]): Promise<Tweets> => {
   return response.data
 }
 
-const titleForAuthor = (author: { name: string }) => {
-  return `${author.name} on Twitter`
+const titleForTweet = (author: { name: string }, text: string) => {
+  return `${author.name} on Twitter: ${text.replace(/http\S+/, '')}`
```
sywhb, line 138: I think we should truncate the `text` if it is too long, i.e. a long thread; and probably put an ellipsis to the end

sywhb: Something like: "Jackson Harper on Twitter: one big thread, content..."

jacksonh: hey, yeah the reason I didn't here is because my first attempt created a string like: `Jackson Harper on Twitter: this is a link to an htt...`, but if I put the full string in, the browser ellipsis's it more nicely on the word boundary. 

I wonder if there is a good library that works across languages for ellipsis.

sywhb: I think saving very long titles for the tweets would bring us some trouble showing them in the Reader as well as saving them in the integrations and so on.

We can truncate the string first and then check the word boundary of the last word

jacksonh: Yeah I was thinking tweets are only 140 chars, but I guess they can be much longer now.

sywhb: Yeah, it is 280 now and 4000 For Twitter Blue Subscribers 😮

sywhb (approved):
LGTM

PR #2112, opened 2023-04-25
Do not enlarge small images, fallback to original src if proxy URL fails

jacksonh:


jacksonh:
Would like this one to sit on demo for a bit.

PR #2128, opened 2023-04-27
fix/trim user email sign up

sywhb:
- trim user email in mobile sign-up router
- trim user email in mobile sign-in router
- trim user email before saving in db

### DIFF
--- a/packages/api/src/routers/auth/mobile/sign_up.ts
+++ b/packages/api/src/routers/auth/mobile/sign_up.ts
@@ -1,16 +1,16 @@
 /* eslint-disable @typescript-eslint/restrict-template-expressions */
+import UserModel from '../../../datalayer/user'
+import { createUser } from '../../../services/create_user'
+import { hashPassword } from '../../../utils/auth'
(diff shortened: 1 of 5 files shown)


jacksonh (commented):
Thinking maybe we have one issue now, I think we check for empty string before the trim, so if a user set their email to `  ` then the empty string check would pass and we'd later create their email as '' because we trim

packages/api/src/routers/auth/mobile/sign_up.ts
```diff
@@ -45,24 +46,24 @@ export async function createMobileSignUpResponse(
...
 export async function createMobileEmailSignUpResponse(
-  email?: string,
-  password?: string,
-  username?: string,
-  name?: string
+  requestBody: any
 ): Promise<JsonResponsePayload> {
   try {
-    if (!email || !password || !username || !name) {
+    if (!isValidSignupRequest(requestBody)) {
```
sywhb, line 52: @jacksonh I added the validation here to check the length of the user email address before trimming

jacksonh (approved):

PR #2157, opened 2023-05-02
Spanish translation

josemas79:
I think it's ok, however there could be some improvement later when I see everything in the context of the app ;)


jacksonh:
Looks good to me. Will do a build and test it out

Issue #2118, opened 2023-04-26
Support iOS's native tap-to-scroll-to-top


ArttuLai:
Not sure if this is even a bug, but I noticed that the native feature where you tap the status bar (clock, battery, connectivity indicator bar) to scroll the current page to top doesn't work. I think it would great if this was supported on iOS.

jacksonh:
Thank you, yes I think there was some gestures that collided with our hiding navigation bar but we will take another look at this one.

jacksonh:
OK, this is updated in TestFlight beta now. If you tap the status bar when the nav bar is visible it will scroll to top. If the nav bar isn't visible, first we will show it, the next tap will then scroll to top.

PR #2160, opened 2023-05-02
Set the iOS caret colour based on theme

jacksonh:


jacksonh:
Fixes: https://github.com/omnivore-app/omnivore/issues/2118

Issue #2057, opened 2023-04-15
Feature Request: Sort labels by name and color


david-vaclavek:
Hello.

To improve the user experience, it would be nice to optionally sort labels not only alphabetically, but also "group" them by colors.

It would prevent such mash-up.

<img width="402" alt="image" src="https://user-images.githubusercontent.com/49518842/232209482-12ad68c5-4e8c-4d7d-aafa-7de75f7ee9f4.png">

Thanks for considering.

jacksonh:
Interesting, just to clarify: 

1. this is on the cards, right? Not in the label lists. 

2. In this screenshot are there three tags that have been grouped into one because they all have the same colour?

david-vaclavek:
Hi @jacksonh.

1.
Yes, this is on the cards.

2.
No. I may have not explained my FR correctly.
There are three tags:
- AI (orange)
- OmnivoreToObsidian:Technology.AI (violet)
- Technology (orange)

What I'd like it to have it:
- AI (orange)
- Technology (orange)
- OmnivoreToObsidian:Technology.AI (violet)

So not sorted only by tag name A-Z, but optionally also by color.

If multiple color groups (semantic groups) would be used, it would help me to orientate in the UI better.

jacksonh:
Thanks @david-vaclavek and would the sort order just be a user setting or is it something you'd want to change while looking at your library?

david-vaclavek:
@jacksonh To me personally, this would be just a matter of user setting (one timer).

But I can definitely imagine use cases where a user would like to switch this while browsing through the library.

Any of the solutions would work for me.

jacksonh:
OK, so here is how I've implemented the first version of this after playing a bit: We group by colour, then alphabetically sort the items in each group, then sort the groups by the first element's name. I think this gives a pretty nice effect of things being grouped, but also being in a fairly logical order:


<img width="362" alt="Screen Shot 2023-05-05 at 12 33 57" src="https://user-images.githubusercontent.com/75189/236378400-130963c9-e051-48cc-8020-ca3a8a9a04f9.png">

jacksonh:
This is done on the cards now. I haven't updated the reader view yet so I will keep this open.

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #2170, opened 2023-05-05
Fix issue where incorrect username validation message was displayed during signup, add label sorting to cards

jacksonh:


jacksonh:
Also implements https://github.com/omnivore-app/omnivore/issues/2057

PR #2183, opened 2023-05-11
Reduce the rate limit for unauthenticated users

jacksonh:


packages/api/src/server.ts
```diff
@@ -100,7 +101,14 @@ export const createApp = (): {
...
     const apiLimiter = rateLimit({
       windowMs: 60 * 1000, // 1 minute
-      max: 50, // Limit each IP to 10 requests per `window` (here, per minute)
+      max: (req) => {
+        // 50 RPM for an authenticated request, 5 for a non-authenticated request
+        const token = getClaimsByToken(
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+          req.header('authorization') ?? req.cookies['auth']
+        )
+        return !!token ? 50 : 10
```
sywhb, line 110: I guess we don't need double negation here

jacksonh: Oh, that was actually hiding a real issue, since getClaimsByToken is async.

sywhb (approved):
LGTM

PR #2217, opened 2023-05-18
fix/highlighted markdown

sywhb:
- fix: `&npsp;` breaks highlight
- fix: wrap link with `==`
- fix: sentry report on unable to find highlight

### DIFF
--- a/packages/api/src/utils/highlightGenerator.ts
+++ b/packages/api/src/utils/highlightGenerator.ts
@@ -422,7 +422,7 @@ export function makeHighlightNodeAttributes(
           isParagraphStart && !i && quote && (quote += '\n')
           quote += text
         }
-        console.log('quote', quote)
+
         const newHighlightSpan = document.createElement('span')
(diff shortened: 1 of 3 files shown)


packages/api/src/utils/highlightGenerator.ts
```diff
@@ -419,7 +422,7 @@ export function makeHighlightNodeAttributes(
           isParagraphStart && !i && quote && (quote += '\n')
           quote += text
         }
-
+        console.log('quote', quote)
```
jacksonh, line 425: not sure if you need this anymore

sywhb: haha, i'm removing this and then i saw your comment @jacksonh

jacksonh (approved):
I have a feeling we need to start building up test cases for this as there are going to be a lot of edge cases.

jacksonh:
LGTM

PR #2221, opened 2023-05-19
Change rate limiting rules

jacksonh:


packages/api/src/server.ts
```diff
@@ -102,12 +102,12 @@ export const createApp = (): {
     const apiLimiter = rateLimit({
       windowMs: 60 * 1000, // 1 minute
       max: async (req) => {
```
sywhb, line 104: @jacksonh probably we could remove `async` here too

sywhb (approved):
LGTM

packages/api/src/server.ts
jacksonh: 👍

PR #2245, opened 2023-05-24
Handle empty state columns in the import CSV files

jacksonh:


sywhb (approved):
Nice!

jacksonh:
> Nice!

Thanks, just added a test for labels too, but looks like no issues with that.

PR #2249, opened 2023-05-25
Update to the web menu and header sizes + header interaction

jacksonh:
- Modify the sizes of the left menu and the header
- Change the xxl min-width break point for the library grid


jacksonh:
Will do some testing of this one on demo

Issue #2265, opened 2023-05-26
Filter by Label dropdown on iOS could be sort by alphabetical order


gabdutra:

jacksonh:
This is fixed in TestFlight build

PR #2271, opened 2023-05-29
iOS label sorting

jacksonh:
- Sort the filter labels on iOS
- Use the same sorting of labels on iOS as Web


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/2265

PR #2298, opened 2023-06-01
Update label styles on iOS and MacOS 

jacksonh:
- Library colour and font size tweaks
- Use new style labels in the library
- Update label styles on iOS and MacOS


jacksonh:
Going to try and merge this one so I can rebase some of the Mac stuff.

PR #2308, opened 2023-06-05
Prevent width scrollbar on windows in primary layout

jacksonh:
- Remove unused tap handler
- Dont overlfow X in our primary layout


jacksonh:
Handled in a different PR

Issue #2284, opened 2023-05-31
Fetch metadata


sywhb:
- create a queue to fetch metadata when starting to save an article
- use the largest and squarest image as the thumbnail if there isn't one and pre-cache all the images
- update the metadata in the subscriber
- remove saving metadata in puppeteer
- modify updateMetadata api to allows changes for image

PR #2311, opened 2023-06-06
Feature: Set thumbnail and pre-cache all the images in the content

sywhb:
- feat: update api to include preview image
- feat: add thumbnail handler
- feat: start a cloud task to set thumbnail and pre-cache all the images in the content


sywhb:
Fix #2284

jacksonh (approved):
Seems good, will just need to do some testing in demo to figure out what happens.

jacksonh (approved):
Seems good, will just need to do some testing in demo to figure out what happens.

PR #2315, opened 2023-06-06
fix: parse unsubscribe mailto email address and set subject

sywhb:


packages/api/src/resolvers/subscriptions/index.ts
```diff
@@ -90,10 +90,13 @@ export const unsubscribeResolver = authorized<
...
 
-    const subscription = await getRepository(Subscription).findOne({
-      where: { name: ILike(name), user: { id: uid } },
-      relations: ['newsletterEmail'],
-    })
+    const subscription = await getRepository(Subscription)
+      .createQueryBuilder('subscription')
+      .innerJoinAndSelect('subscription.newsletterEmail', 'newsletterEmail')
+      .where({ user: { id: uid } })
+      .andWhere('LOWER(name) = LOWER(:name)', { name }) // case insensitive
```
jacksonh, line 97: nice

jacksonh (approved):
nice, and tests too!

Issue #2103, opened 2023-04-24
bulk api


sywhb:
something like: `changePages(pageIds: string[], action: 'ARCHIVE' | 'DELETE' | 'MARK_READ' | 'ADD_LABELS', arg -- labelIds: [])`

jacksonh:
Maybe to match some other APIs `update` is better than `change`

sywhb:
@jacksonh Agree!

sywhb:
I'm just writing down this and we can finalize the API later

PR #2329, opened 2023-06-08
update bulk api to allow mark as read and add labels

sywhb:
- feat: add bulk action api schema
- allow mark as read and add labels bulk actions

### DIFF
--- a/packages/api/src/resolvers/article/index.ts
+++ b/packages/api/src/resolvers/article/index.ts
@@ -31,6 +31,7 @@ import {
   BulkActionError,
   BulkActionErrorCode,
   BulkActionSuccess,
+  BulkActionType,
   ContentReader,
   CreateArticleError,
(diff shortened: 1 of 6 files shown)


packages/api/src/resolvers/article/index.ts
```diff
@@ -1097,26 +1098,43 @@ export const bulkActionResolver = authorized<
...
-    log.error('bulkActionResolver', { error: 'Unauthorized' })
-    return { errorCodes: [BulkActionErrorCode.Unauthorized] }
-  }
-
   analytics.track({
     userId: uid,
     event: 'BulkAction',
     properties: {
       env: env.server.apiEnv,
+      action,
```
jacksonh, line 1109: thank you!

jacksonh (approved):

sywhb:
Fix #2103

Issue #1234, opened 2022-09-25
“Favorites” section


xplosionmind:
As there are the “Archive”, the “to read”, the “inbox” section, etc, it would be nice to have a “**Favorites**” section, with all of the user’s most favorite articles—duh.

This would be a nice thing also to show favorite articles as “pinned” in a user’s profile, when social functionalities will be implemented (#409, #419).

---

<img width="690" alt="Screen Shot 2022-05-12 at 12 21 58" src="https://user-images.githubusercontent.com/75189/168152449-962b3aaf-f28d-4c69-9b31-051e9ac94e5d.png">

_Originally posted by @jacksonh in https://github.com/omnivore-app/omnivore/issues/409#issuecomment-1125343300_

jacksonh:
I think with this one, we can probably use a hidden label, and then a special query for `Favorites`, what would be handy here is when syncing to other platforms, you might only want to sync favorites, for example, maybe I only want my Favorites imported into Logseq.

jacksonh:
Commenting to send this one to the top. /cc @sywhb -- let me know if you think we should use a magic label or a new field here.

sywhb:
@jacksonh Yeah, I think we can use a magic label for it and make it non-editable

sywhb:
Created a SetFavorite API to add the magic label `favorites` to the article

PR #2332, opened 2023-06-09
feature/favorites

sywhb:
- feat: make newsletters and favorites internal labels
- feat: add set favorite article api which adds the favorites internal label to it


jacksonh (approved):

sywhb:
Fix #1234 

For the client, we need to call the `setFavoriteArticle` API to mark an article as "favorite" which would add a "Favorites" label to it.

To fetch a list of "favorites", we could search for `label:favorites`

PR #2336, opened 2023-06-12
fix/thumbnailer task queue

sywhb:
- use a separate queue for thumbnailer
- better handling error in puppeteer

### DIFF
--- a/packages/api/src/utils/createTask.ts
+++ b/packages/api/src/utils/createTask.ts
@@ -56,7 +56,9 @@ const createHttpTaskWithToken = async ({
   }
 
   // Construct the fully qualified queue name.
-  priority === 'low' && (queue = `${queue}-low`)
+  if (priority === 'low') {
+    queue = `${queue}-low`
+  }
 
   const parent = client.queuePath(project, location, queue)
   // Convert message to buffer.
@@ -534,6 +536,7 @@ export const enqueueThumbnailTask = async (
     payload,
     taskHandlerUrl: env.queue.thumbnailTaskHandlerUrl,
     requestHeaders: headers,
+    queue: 'omnivore-thumbnail-queue',
   })
 
   if (!createdTasks || !createdTasks[0].name) {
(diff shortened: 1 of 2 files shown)


packages/api/src/utils/createTask.ts
```diff
@@ -56,7 +56,9 @@ const createHttpTaskWithToken = async ({
   }
 
   // Construct the fully qualified queue name.
-  priority === 'low' && (queue = `${queue}-low`)
+  if (priority === 'low') {
```
jacksonh, line 59: lol so much easier to understand.

packages/api/src/utils/createTask.ts
```diff
@@ -534,6 +536,7 @@ export const enqueueThumbnailTask = async (
     payload,
     taskHandlerUrl: env.queue.thumbnailTaskHandlerUrl,
     requestHeaders: headers,
+    queue: 'omnivore-thumbnail-queue',
```
jacksonh, line 539: Guess we just need to make sure we create it in prod.

packages/import-handler/src/index.ts
```diff
@@ -223,15 +223,29 @@ const sendSavePageMutation = async (userId: string, input: unknown) => {
...
+  try {
+    const response = await axios.post(
+      `${REST_BACKEND_ENDPOINT}/graphql`,
+      data,
+      {
+        headers: {
+          Cookie: `auth=${auth};`,
+          'Content-Type': 'application/json',
+        },
+        timeout: 30000, // 30s
```
jacksonh, line 235: Seems very high but I guess the side effects aren't too bad, it just means the thumbnailer instances stay alive longer and probably slows down thumbnailing and saves the API if things are getting slow.

jacksonh (approved):

packages/import-handler/src/index.ts
sywhb: yeah, that's exactly my thought

PR #2340, opened 2023-06-13
fix update pages

sywhb:
- refresh index if update is not async
- refresh index after archiving page

### DIFF
--- a/packages/api/src/elastic/pages.ts
+++ b/packages/api/src/elastic/pages.ts
@@ -696,7 +696,7 @@ export const searchAsYouType = async (
 }
 
 export const updatePages = async (
-  userId: string,
+  ctx: PageContext,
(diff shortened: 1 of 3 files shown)


packages/api/src/elastic/pages.ts
```diff
@@ -789,6 +790,8 @@ export const updatePages = async (
       return null
     }
 
+    // TODO: publish entityUpdated events for each page
```
jacksonh, line 793: oh good TODO.

jacksonh (approved):

packages/api/src/elastic/pages.ts
sywhb: haha, need to wrap up a new way to publish bulk events

PR #2344, opened 2023-06-13
Use useCallback on performAction dependencies

jacksonh:


jacksonh:
Will handle this separately

Issue #2338, opened 2023-06-12
Enter does not save edited changes


rajivharlalka:
When Info about a saved url is edited and `Enter` is pressed on keyboard, instead of saving changes, the dialog box just closes discarding the changes

jacksonh:
Hey thank you for the report, this one is fixed now @rajivharlalka -- let me know if you still notice issues.

PR #2345, opened 2023-06-13
Fix issue with enter not properly submitting the edit info modal

jacksonh:


jacksonh:
Fixes https://github.com/omnivore-app/omnivore/issues/2338

PR #2358, opened 2023-06-14
feat/api reduce events

jacksonh:
- Remove some debugging
- Remove backend events that we dont need to watch


sywhb (approved):
Thanks to cover me up while I was stuck with the csv importer

jacksonh:
haha no worries!

PR #2359, opened 2023-06-14
fix/pocket import

sywhb:
- stringify csv before uploading
- add debug logs
- use a separate queue for sending emails
- allow more labels format in the csv


jacksonh (approved):
Is the stringify needed because some data wasn't strings?

sywhb:
> Is the stringify needed because some data wasn't strings?

there could be comma in the url

PR #2360, opened 2023-06-14
WIP: use native colour picker

jacksonh:


jacksonh:
Will close this as I had to incorporate as part of the label picker changes.

PR #2363, opened 2023-06-14
create another rate limiter for auth api with 5 rpm

sywhb:


packages/api/src/server.ts
```diff
@@ -97,32 +98,30 @@ export const createApp = (): {
...
+    keyGenerator: (req) => {
+      return (
+        req.header('authorization') ||
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        (req.cookies['auth'] as string) ||
+        req.ip
+      )
+    },
+    // skip preflight requests and test requests
+    skip: (req) => req.method === 'OPTIONS' || env.dev.isLocal,
```
jacksonh, line 120: oh nice!

packages/api/src/server.ts
```diff
@@ -137,11 +136,19 @@ export const createApp = (): {
   // respond healthy to auto-scaler.
   app.get('/_ah/health', (req, res) => res.sendStatus(200))
 
-  app.use('/api/auth', authRouter())
+  // 5 RPM for auth requests
+  const authLimiter = rateLimit({
+    windowMs: 60 * 1000, // 1 minute
+    max: 5,
```
jacksonh, line 142: I wonder if during account creation this will be slightly too low. I'm just going to go through the flow now and see how many times its called.

sywhb: yeah, one concern is "login" could be called many times if the user is unsure about the password but probably we should also limit the number of attempts

jacksonh (approved):

PR #2368, opened 2023-06-15
Add robots.txt to the list of reserved names

jacksonh:


packages/api/src/utils/usernamePolicy.ts
```diff
@@ -116,6 +116,7 @@ const RESERVED_NAMES = new Set([
   'register',
   'remove',
   'replies',
+  'robots.txt',
```
sywhb, line 119: May I know how will this affect us?

sywhb (approved):

PR #2371, opened 2023-06-16
Use normalized URLs when querying article saving requests

jacksonh:
When we save pages we save the normalized URL, so if a user saves
a URL like https://foo.com/ in elastic it will be saved as
https://foo.com but the browser will redirect to this url:

https://omnivore.app/article?url=https://foo.com/ and the URL
lookup will fail.


sywhb (approved):
LGTM!

PR #2389, opened 2023-06-20
Add The Atlantic handler to avoid paywall and correctly format

Podginator:
Added a handler to parse the Atlantic content, and fixed a small issue where the line-height would not be correctly considered. 

I believe the actual issue is in a CSS property, but I didn't know whether or not that was used elsewhere and so instead I changed the article to a div.


jacksonh:
I'll run `yarn lint --fix` on this in a bit, i think that should clean up the linting errors.

Podginator:
Made a few changes to remove stuff unrelated to the specific article. I find that they break up the flow of the article when reading on a Read-it-later app.

jacksonh:
Looks like I need to lint again, should get this merged and deployed today though.

jacksonh:
NIce, should be in prod in a few hours. Thank you!

PR #2393, opened 2023-06-20
fix/scrape twitter

sywhb:
- change the page viewport when scraping twitter thread
- feat: add nitter handler to scrape tweets from twitter.com and nitter.net


packages/content-handler/src/index.ts
```diff
@@ -64,7 +64,6 @@ const contentHandlers: ContentHandler[] = [
   new PdfHandler(),
   new ScrapingBeeHandler(),
   new TDotCoHandler(),
-  new TwitterHandler(),
```
jacksonh, line 67: So the nitter handler does both now?

sywhb: yeah

jacksonh (approved):

PR #2399, opened 2023-06-21
fix/subscription state

sywhb:
- fix: mark failed to unsubscribe newsletters as unsubscribed state
- do not subscribe a newsletter if subscription already exists and is unsubscribed

### DIFF
--- a/packages/api/src/routers/svc/newsletters.ts
+++ b/packages/api/src/routers/svc/newsletters.ts
@@ -3,6 +3,7 @@ import {
   createPubSubClient,
   readPushSubscription,
 } from '../../datalayer/pubsub'
+import { SubscriptionStatus } from '../../generated/graphql'
 import {
   getNewsletterEmail,
   updateConfirmationCode,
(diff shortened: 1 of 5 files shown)


packages/api/src/routers/svc/newsletters.ts
```diff
@@ -128,6 +130,16 @@ export function newsletterServiceRouter() {
           return res.status(500).send('Error saving url from email')
         }
       } else {
+        // do not subscribe if subscription already exists and is unsubscribed
+        const existingSubscription = await getSubscriptionByNameAndUserId(
+          data.author,
+          newsletterEmail.user.id
+        )
+        if (existingSubscription?.status === SubscriptionStatus.Unsubscribed) {
```
jacksonh, line 138: nice, so these wont get put into recents, right?

jacksonh (approved):

packages/api/src/routers/svc/newsletters.ts
sywhb: @jacksonh i choose to put these into recent emails so that a user could still subscribe a `Unsubscribed` newsletter by marking the email as an article

let me know if there is a better way to handle it

jacksonh: Yeah I would say still going to recents makes sense, maybe it could have a status like `unsubscribed` or something in the future.

sywhb: yeah, that is a good idea!

PR #2400, opened 2023-06-21
Add support for multi select Add Label

jacksonh:
- Add support for multi select labels
- Update labels on cards when setting from library


sywhb (approved):
LGTM

Issue #2267, opened 2023-05-27
Feature suggestion: Add a copy button to highlights


karlemilnikka:
When sentences are highlighted, they become harder to mark and copy. I suggest adding a copy button to the existing action bar that’s displayed when clicking on a highlight (i.e., the bar with “delete”, “labels” and “note” buttons).

jacksonh:
Thanks @karlemilnikka this is on web, right?

karlemilnikka:
@jacksonh Yes, exactly.

jacksonh:
From the community: https://github.com/omnivore-app/omnivore/pull/2414 -- hopefully in production soon

jacksonh:
Done!

PR #2414, opened 2023-06-22
Add copy button to highlight bar

340rahul:
Addresses issue [2267](https://github.com/omnivore-app/omnivore/issues/2267)

- Adds a copy button to the highlight bar to allow user to copy selected/highlighted text to clipboard
- Clean up of `HighlightBar.tsx` to remove dup code

### DIFF
--- a/packages/web/components/templates/article/HighlightsLayer.tsx
+++ b/packages/web/components/templates/article/HighlightsLayer.tsx
@@ -20,7 +20,7 @@ import { removeHighlights } from '../../../lib/highlights/deleteHighlight'
 import { createHighlight } from '../../../lib/highlights/createHighlight'
 import { HighlightNoteModal } from './HighlightNoteModal'
 import { NotebookModal } from './NotebookModal'
-import { showErrorToast } from '../../../lib/toastHelpers'
+import { showErrorToast, showSuccessToast } from '../../../lib/toastHelpers'
 import { ArticleMutations } from '../../../lib/articleActions'
 import { isTouchScreenDevice } from '../../../lib/deviceType'
 import { UserBasicData } from '../../../lib/networking/queries/useGetViewerQuery'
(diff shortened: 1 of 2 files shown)


packages/web/components/templates/article/HighlightsLayer.tsx
```diff
@@ -454,6 +455,30 @@ export function HighlightsLayer(props: HighlightsLayerProps): JSX.Element {
...
+              focusedHighlight.id
+            )
+            highlightedElements.forEach(
+              (element) => (textToCopy += element.textContent)
+            )
+          } else if (userSelectionText) {
+            textToCopy = userSelectionText
+          }
+
+          if (textToCopy) await navigator.clipboard.writeText(textToCopy)
```
jacksonh, line 476: We might want to add a success toast

340rahul: I guess we will want to use the `showSuccessToast` from `toastHelpers.tsx` for this? Is "Copied Successfully!" a good message? Might be a corner case (if user has the clipboard access disabled) but do we want to show an error toast as well?

jacksonh: Yeah we can do an error toast too, something like:

```
if (success) {
   showSuccessToast('Highlight copied', {
            position: 'bottom-right',
          })
   } else {
   showErrorToast('Error copying highlight, permission denied.',  {
            position: 'bottom-right',
          })
```
   }

340rahul: Added the toasts :)

jacksonh:
Thanks! Took on the scariest file in our code base too. Going to get this into our staging environment and hopefully on to production later today or tomorrow (we have a few web things stacked up on staging right now that I want to test more)

jacksonh:
Will have to tweak this a bit for mobile

IMG_5CB83E977E46-1

340rahul:
> Will have to tweak this a bit for mobile
> 
> IMG_5CB83E977E46-1

Oops! Sorry, I missed this. Working on a fix.

PR #2428, opened 2023-06-27
fix: failed to save tweet

sywhb:
- fix: tweet not saved correctly when using share button on iOS
- save normalized url
- remove tracking params from tweet url


jacksonh (commented):
Looks good, only concern is making sure when we save the same URL multiple times it still deduplicates.

PR #2445, opened 2023-06-30
feat: create labels if not exist when setting labels

sywhb:
- feat: create labels if not exist when setting labels in the page
- feat: create labels if not exist when setting labels in the highlight


sywhb:
I merged this so it will be deployed in demo and it doesn't break the existing set labels feature

Issue #2436, opened 2023-06-27
Improve full-text search result


sywhb:
* sort by score, savedAt
* boost title score probably by three `title^3`
* tie-breaker

sywhb:
ref: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-multi-match-query.html#field-boost

jacksonh:
Linking this somewhat related item here as well: https://github.com/omnivore-app/omnivore/issues/2129

Issue #2129, opened 2023-04-27
Search does not look into manually added note


jakemkc:
Hi all,

Thanks for creating the great read-it-later app Omnivore for the community. I use it almost daily and it's becoming an important part of my routine.

I usually create notes for the article saved. However, I discovered that the search function won't look for text in my manually typed notes. 

I wonder is it a design feature not to search into notes but only text from the original page? 

Best,
Jake

jacksonh:
Thanks @jakemkc yeah we definitely need this. Perhaps as a starting point a scope like `note:<the text>`.

it was definitely not intentional, just search was implemented before we really had notes.

rochakagrawal:
just to add - i would love this feature as well

jacksonh:
Linking these two https://github.com/omnivore-app/omnivore/issues/1802

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #2447, opened 2023-06-30
fix/full text search

sywhb:
- fix: improve full-text search result
- feat: allow searching by `note:${some note}`
- feat: show soft deleted pages within 14 days by searching `in:trash`
- feat: allow recovering a soft deleted item through update page api

### DIFF
--- a/packages/api/src/elastic/pages.ts
+++ b/packages/api/src/elastic/pages.ts
@@ -92,6 +92,17 @@ const appendInFilter = (builder: ESBuilder, filter: InFilter): ESBuilder => {
       return builder.query('exists', { field: 'archivedAt' })
     case InFilter.INBOX:
       return builder.notQuery('exists', { field: 'archivedAt' })
+    case InFilter.TRASH:
+      // return only deleted pages within 14 days
+      return builder
+        .query('term', {
+          state: ArticleSavingRequestStatus.Deleted,
+        })
+        .andQuery('range', {
+          updatedAt: {
+            gte: 'now-14d',
+          },
+        })
   }
   return builder
 }
@@ -515,7 +526,7 @@ const buildSearchBody = (userId: string, args: PageSearchArgs) => {
       state: ArticleSavingRequestStatus.Processing,
     })
   }
-  if (!args.includeDeleted) {
+  if (!args.includeDeleted && inFilter !== InFilter.TRASH) {
     builder = builder.notQuery('term', {
       state: ArticleSavingRequestStatus.Deleted,
     })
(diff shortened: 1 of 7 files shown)


sywhb:
Fix #2436 #2129

packages/api/src/elastic/pages.ts
```diff
@@ -28,14 +28,21 @@ import {
 } from './types'
 
 const appendQuery = (builder: ESBuilder, query: string): ESBuilder => {
-  const fields = ['title', 'content', 'author', 'description', 'siteName']
+  const fields = [
+    { field: 'title', boost: 3 },
+    { field: 'content', boost: 1 },
+    { field: 'author', boost: 1 },
+    { field: 'description', boost: 1 },
+    { field: 'siteName', boost: 1 },
```
jacksonh, line 36: We can experiment, but I suspect we want to boost title even more and maybe siteName a little

jacksonh:
This looks great, I think I should just merge it so we can use on demo.

packages/api/src/elastic/pages.ts
```diff
@@ -28,14 +28,21 @@ import {
 } from './types'
 
 const appendQuery = (builder: ESBuilder, query: string): ESBuilder => {
-  const fields = ['title', 'content', 'author', 'description', 'siteName']
+  const fields = [
```
jacksonh, line 31: One thing here is I think we should include the note field if possible. So we have scoped note search, but also do text search of the notes when someone does a query, since I think in most cases this content would be relevant to the user.

sywhb: If we include `note` field in the query, the `boost` score and `tie_breaker` might not work very well because `note` is actually stored as a nested field `highlights.annotation` and the `multi_match` query probably does not work with nested fields

sywhb: Let me try if it breaks the results

packages/api/src/elastic/pages.ts
sywhb: I have tuned a bit the boost score and the title boost score seems good enough.

I now boost the siteName score by 2 and it looks good.
https://github.com/omnivore-app/omnivore/pull/2457#issuecomment-1617799805

packages/api/src/elastic/pages.ts
sywhb: It looks okay to add `note` field in the basic search: https://github.com/omnivore-app/omnivore/pull/2457#issuecomment-1617799805

Issue #1993, opened 2023-04-06
Sign Up Error handling unclear: 'wrong username or password' when full name is left empty.


Lionstiger:
## Steps to reproduce:
- Navigate to Email Sign Up: https://omnivore.app/auth/email-signup
- Enter Email, Username and Password but not Full Name.
- Click Sign Up

## Encountered Behavior:
Error Message is "Invalid email or password".

## Expected Behavior:
Error Message showing that Full Name is not optional.
Ideally there would be a visible indication that all fields are mandatory.

jacksonh:
Thank you, linking these two issues together, https://github.com/omnivore-app/omnivore/issues/1968

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

Issue #1968, opened 2023-04-02
Error messages on sign up page aren't very good


jacksonh:
- we require a lowercase letter for usernames but no message about this
- username and full name need to be filled in but no error message for this

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #2459, opened 2023-07-03
#1993, #1968 - Make sign up form fields required

340rahul:
For #1993 and #1968. With this MR, it won't be possible to submit the sign up form without the required fields filled in. 

## Changes in this MR:
- Makes sign up form fields required
- Adds '*' to the labels corresponding to the required fields
- Minor fix - use `defaultValue` instead of `value`

### DIFF
--- a/packages/web/components/elements/FormElements.tsx
+++ b/packages/web/components/elements/FormElements.tsx
@@ -41,6 +41,10 @@ export const FormInput = styled('input', {
 export const FormLabel = styled('label', {
   fontSize: '16px',
   color: '$omnivoreGray',
+  '&.required:after': {
+    content: ' *',
+    color: 'red',
+  },
 })
 
 export const BorderedFormInput = styled(FormInput, {
--- a/packages/web/components/templates/auth/EmailSignup.tsx
+++ b/packages/web/components/templates/auth/EmailSignup.tsx
@@ -14,16 +14,14 @@ import Link from 'next/link'
 
 export function EmailSignup(): JSX.Element {
   const router = useRouter()
-  const [email, setEmail] = useState<string | undefined>(undefined)
-  const [password, setPassword] = useState<string | undefined>(undefined)
-  const [fullname, setFullname] = useState<string | undefined>(undefined)
-  const [username, setUsername] = useState<string | undefined>(undefined)
+  const [email, setEmail] = useState<string | undefined>()
+  const [password, setPassword] = useState<string | undefined>()
+  const [fullname, setFullname] = useState<string | undefined>()
+  const [username, setUsername] = useState<string | undefined>()
   const [debouncedUsername, setDebouncedUsername] = useState<
     string | undefined
-  >(undefined)
-  const [errorMessage, setErrorMessage] = useState<string | undefined>(
-    undefined
-  )
+  >()
+  const [errorMessage, setErrorMessage] = useState<string | undefined>()
 
   useEffect(() => {
     if (!router.isReady) return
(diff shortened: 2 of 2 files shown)


340rahul:
A question about #1968 which mentions,
> we require a lowercase letter for usernames but no message about this

I am guessing this means all letters in username should be lowercase? If so, I was wondering about how do we want to handle the user input; we can either not allow the user to input any uppercase letters and show a toast when the user attempts to do so, or we can allow the user to input uppercase letters but prevent form submission and show a toast when submission is attempted. 

Either of the change should be minor and I can include this as part of this PR.

jacksonh:
Hey thank you. Looking at this one now!

jacksonh:
> 

I think the first option might be best. 

> A question about #1968 which mentions,
> 
> > we require a lowercase letter for usernames but no message about this
> 
> I am guessing this means all letters in username should be lowercase? If so, I was wondering about how do we want to handle the user input; we can either not allow the user to input any uppercase letters and show a toast when the user attempts to do so, or we can allow the user to input uppercase letters but prevent form submission and show a toast when submission is attempted.
> 
> Either of the change should be minor and I can include this as part of this PR.

I think I made a mistake in that issue. Here's the validation function and the regex:

```
const REGEX_POLICY = /^[a-z0-9][a-z0-9_]+[a-z0-9]$/


export const validateUsername = (username: string): boolean => {
  const lowerCasedUsername = username.toLowerCase()
  if (RESERVED_NAMES.has(lowerCasedUsername)) {
    return false
  }

  if (lowerCasedUsername.length < 4 || lowerCasedUsername.length > 15) {
    return false
  }

  if (
    lowerCasedUsername.includes('omnivore') ||
    lowerCasedUsername.includes('admin')
  ) {
    return false
  }

  return REGEX_POLICY.test(lowerCasedUsername)
}


```

jacksonh:
Hey @340rahul I will merge this one as-is. Its a big improvement over what we have!

jacksonh (approved):

340rahul:
> > 
> 
> I think the first option might be best.
> 
> > A question about #1968 which mentions,
> > > we require a lowercase letter for usernames but no message about this
> > 
> > 
> > I am guessing this means all letters in username should be lowercase? If so, I was wondering about how do we want to handle the user input; we can either not allow the user to input any uppercase letters and show a toast when the user attempts to do so, or we can allow the user to input uppercase letters but prevent form submission and show a toast when submission is attempted.
> > Either of the change should be minor and I can include this as part of this PR.
> 
> I think I made a mistake in that issue. Here's the validation function and the regex:
> 
> ```
> const REGEX_POLICY = /^[a-z0-9][a-z0-9_]+[a-z0-9]$/
> 
> 
> export const validateUsername = (username: string): boolean => {
>   const lowerCasedUsername = username.toLowerCase()
>   if (RESERVED_NAMES.has(lowerCasedUsername)) {
>     return false
>   }
> 
>   if (lowerCasedUsername.length < 4 || lowerCasedUsername.length > 15) {
>     return false
>   }
> 
>   if (
>     lowerCasedUsername.includes('omnivore') ||
>     lowerCasedUsername.includes('admin')
>   ) {
>     return false
>   }
> 
>   return REGEX_POLICY.test(lowerCasedUsername)
> }
> ```

Hey, sorry it took me a little longer to respond to this. I will create a follow up PR and incorporate the changes we discussed. 

Thanks for the validation snippet as well. I agree with it except the part where we're checking for `admin`, `omnivore`, and `RESERVED_NAMES` user. I think it's not very secure to have this in code like this. IMHO these should be stored in an environment variable and the check should happen on the BE. I realize that this is a long list and probably not very suited for being an environment variable. Maybe this can be a JSON file which we don't commit and we only commit a template JSON. What do you think about this? Is this an overkill?

Also, thanks for merging this :)

jacksonh:
Thanks! Actually that code is on the backend, there's a `validateUsername` call that is made that calls into that function on the backend. Its done while the user types but then also once the form is submitted the same function is validated on the backend.

340rahul:
Should we still avoid keeping the list of reserve users in code even on the BE? I am aware this is probably out of scope of this PR or the main issues we're addressing.

PR #2462, opened 2023-07-04
fix/wechat published date

sywhb:
- fix: missing published date in some wechat articles
- fix: * published/saved date not editable on web * these two dates are not fully shown


sywhb:
Fix #2425

Issue #2029, opened 2023-04-12
Feed/rss support


avelino:
Bring news (blogpost) automatically from pre-registered feed

daviddavo:
In the meantime, I've been using an RSS to e-mail service as a workaround

I7T5:
UI/UX and RSS are my only concerns for moving from NetNewsWire.

thisisjasonmurphy:
Being able to collate newspapers is a major advantage for me over Readwise Reader at the moment. Integrating RSS would seal the deal.

cs-mshah:
An RSS integration would make this a killer reading app.

ThinkSalat:
Would bring omnivore one step closer to feature parity with Reader

kkiani:
bump!

jacksonh:
Starting testing on this one soon!

FetchFast:
Has it been removed recently? I added 2 RSS feeds. Now I want to remove one, but can't find it anywhere in the interface

davidrsim-p:
> Has it been removed recently? I added 2 RSS feeds. Now I want to remove one, but can't find it anywhere in the interface

Just in case anyone else is looking: https://omnivore.app/settings/feeds

ugurbolat:
Would be cool to have an import functionality from OPML files :)

cori:
> Would be cool to have an import functionality from OPML files :)

c.f. https://github.com/omnivore-app/omnivore/issues/2870

anarcat:
this seems to have been implemented in https://omnivore.app/settings/feeds ?

daviddavo:
Yeah this issue should be closed @sywhb

PR #2465, opened 2023-07-04
feat: rss

sywhb:


sywhb:
Fix #2029

packages/api/src/schema.ts
```diff
@@ -1619,16 +1620,24 @@ const schema = gql`
...
+  enum SubscriptionType {
+    RSS
+    NEWSLETTER
+  }
+
   type Subscription {
     id: ID!
     name: String!
-    newsletterEmail: String!
+    newsletterEmail: String
```
jacksonh, line 1631: I might need to update the clients for this change but shouldn't really be breaking since it would be set for all existing.

packages/api/src/services/subscriptions.ts
```diff
@@ -96,42 +98,63 @@ export const saveSubscription = async ({
...
+    let unsubscribed = false
+
+    if (subscription.unsubscribeMailTo && subscription.newsletterEmail) {
+      // unsubscribe by sending email
+      unsubscribed = await sendUnsubscribeEmail(
+        subscription.unsubscribeMailTo,
+        subscription.newsletterEmail.address
+      )
+    }
+    // TODO: find a good way to unsubscribe by url if email fails or not provided
```
jacksonh, line 145: I think in this case we might need to return the URL to the client, that's at least what I have seen GMail do

packages/api/src/routers/svc/rss_feed.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
+import express from 'express'
+import { readPushSubscription } from '../../datalayer/pubsub'
+import { Subscription } from '../../entity/subscription'
+import { getRepository } from '../../entity/utils'
+import { SubscriptionStatus, SubscriptionType } from '../../generated/graphql'
+import { enqueueRssFeedFetch } from '../../utils/createTask'
+
+export function rssFeedRouter() {
```
jacksonh, line 9: So this is triggered by the cron and its job is to create one cloud task per RSS subscription, right? Then the rss-handler does the actual fetching of the RSS feeds?

packages/api/src/services/subscriptions.ts
sywhb: yeah, and we can open the URL on the client side

packages/api/src/routers/svc/rss_feed.ts
sywhb: yes, exactly

jacksonh (approved):

PR #2482, opened 2023-07-12
fix save page api request invalid parameter

sywhb:


jacksonh (approved):
Can you set this to a random uuid? Or does it not matter?

sywhb:
> Can you set this to a random uuid? Or does it not matter?

I think it does not matter here because we are creating a new page here

PR #2488, opened 2023-07-14
fix/rss

sywhb:
- skip rss items which are published before 48 hours
- set savedAt and publishedAt timestamp for rss item

### DIFF
--- a/packages/rss-handler/src/index.ts
+++ b/packages/rss-handler/src/index.ts
@@ -114,16 +114,15 @@ export const rssHandler = Sentry.GCPFunction.wrapHttpFunction(
 
       // save each item in the feed
       for await (const item of feed.items) {
-        const publishedAtString = item.pubDate || item.isoDate
-        console.log('Processing feed item', item.link, publishedAtString)
+        console.log('Processing feed item', item.link, item.isoDate)
 
-        if (!item.link || !publishedAtString) {
+        if (!item.link || !item.isoDate) {
           console.log('Invalid feed item', item)
           continue
         }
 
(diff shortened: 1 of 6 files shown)


packages/rss-handler/src/index.ts
```diff
@@ -114,15 +114,20 @@ export const rssHandler = Sentry.GCPFunction.wrapHttpFunction(
 
       // save each item in the feed
       for await (const item of feed.items) {
-        const publishedAt = item.pubDate || item.isoDate
-        console.log('Processing feed item', item.link, publishedAt)
+        const publishedAtString = item.pubDate || item.isoDate
```
jacksonh, line 117: Do you know the format of this `pubDate`? Not sure if the RSS library parses it for us but i noticed its this format in my one testing feed `<pubDate>Wed, 12 Jul 2023 04:00:00 -0000</pubDate>`

jacksonh (approved):
Looks good, I am surprised that RSS doesn't force ISO format on date strings

packages/rss-handler/src/index.ts
sywhb: i think the `pubDate` format is not fixed. I could use the `isoDate` which is in ISO format that the RSS node library parsed from `pubDate`

jacksonh: interesting, yeah i noticed the feed i have doesn't have isoDate, does the library add that for us?

sywhb: yeah, the `isoDate` is added by the library getting from `pubDate`

PR #2489, opened 2023-07-14
add in:subscription and in:library in search scope

sywhb:


packages/api/src/elastic/pages.ts
```diff
@@ -148,6 +148,50 @@ const appendInFilter = (builder: ESBuilder, filter: InFilter): ESBuilder => {
             gte: 'now-14d',
           },
         })
+    case InFilter.LIBRARY:
+      return builder.query('nested', {
+        path: 'labels',
+        query: {
+          bool: {
+            should: [
```
jacksonh, line 156: Does this mean it would need the `library` label to return true?

sywhb: @jacksonh not really. it will be true if it has `library` label or it does not have either `newsletter` or `rss` label

jacksonh: great, thanks! Looks good!

jacksonh (approved):

PR #2497, opened 2023-07-17
make library an internal label

sywhb:


jacksonh (approved):

jacksonh:
One thing I guess we should do is check to see if anyone has this label. I think its probably OK though, since it should be most people's intent, trying to do something similar to this.

sywhb:
> One thing I guess we should do is check to see if anyone has this label. I think its probably OK though, since it should be most people's intent, trying to do something similar to this.

yeah, i think it's okay since it won't affect them.

@jacksonh probably i should also do a db migration to update the existing "library" labels and mark them as "internal"?

jacksonh:
> > One thing I guess we should do is check to see if anyone has this label. I think its probably OK though, since it should be most people's intent, trying to do something similar to this.
> 
> yeah, i think it's okay since it won't affect them.
> 
> @jacksonh probably i should also do a db migration to update the existing "library" labels and mark them as "internal"?

yeah, or at least we can just get an idea of how much its used already first.

sywhb:
@jacksonh i just checked in kibana and we have 94 items labelled with `library` in total

PR #2502, opened 2023-07-18
do not include archived items in library or subscription

sywhb:


jacksonh (approved):

sywhb:
> eek, surprised i didn't notice

yeah me too. i think on web it firstly "removed" the item from "read later" and if you refresh, it will be back

PR #2518, opened 2023-07-20
Update email inputs to type email

MagsMagnoli:
Updated email inputs from `text` to `email` which helps with autocomplete and accessibility


jacksonh:
nice. Thank you!

But @MagsMagnoli how do I get you to change this 😆 


Screen Shot 2023-07-24 at 12 06 24

PR #2519, opened 2023-07-20
Add node version

MagsMagnoli:
This enables node version managers like [fnm](https://github.com/Schniz/fnm) to select the appropriate version of node


jacksonh:
FYI, we will switch over to node 18 this week, but I'll update this as well.

PR #2541, opened 2023-07-25
Control article margins on small viewports

karlprieb:
This change is only for devices with viewport < 678px. Anything bigger like tablets and desktops will keep the current behavior.

Basically I had to find a width percentage that would behave similarly as the fixed widths used on desktop. Knowing that the margin selector goes from 200 to 560 in steps of 45 I came up the following calc:
For mobile the max width will be 120% - (10% of margin selector value)
10% seems to be the percentage that best fit the current behavior on desktop.

I also removed a unused margin.

Phone
Kapture 2023-07-25 at 18 27 37

Small Tablet
Kapture 2023-07-25 at 18 30 33


jacksonh (approved):

jacksonh:
Going to merge this one and play with it in our demo environment.

PR #2555, opened 2023-07-28
Vendor the mathjax and highlightjs code to avoid CSP loading

jacksonh:


jacksonh:
Need to test this one on demo, i think there will still be some CSP issues.

PR #2563, opened 2023-07-28
Fix CSP for intercom

jacksonh:


jacksonh:
Seems like this doesn't work at all and the issue is related to something else.

PR #2575, opened 2023-08-02
Added handler to Bypass Wired Paywall.

Podginator:
Also did some cleaning up of the content to remove mid-fill stuff like related content from the article itself.


jacksonh:
tests 🤤

PR #2592, opened 2023-08-04
Rename rss to feeds in web (redirect from rss to feeds)

jacksonh:


sywhb:
Do we need to change the title and button too?

PR #2608, opened 2023-08-04
Ars Technica Multipage handling

Podginator:
ArsTechnica occasionally does long form articles that span multiple pages. 

When handled in the default path, these get ignored. I have added some code to follow the links, and append it all to the same Document.


sywhb (approved):
Thanks! LGTM

PR #2617, opened 2023-08-07
[Feat] Add Links to Search, and show "PROCESSING" Links in the Card/List Views. 

Podginator:
Hey - I've been working on something. 

So when I'm on desktop I tend to add links manually - copy and paste the URL and add it that way - I know there's a chrome plugin, but I always forget to use it. And I'm using a new browser that I'm trying to get used to. I find myself copying and pasting links into the Search Bar and then going "Oh right..." 

I also find it really jarring that when you do hit enter the link isn't populated in your list of links. I always expect it to show up, and it makes it feel a little... undynamic. 

So I worked on two things:

-   **Contextual search bar;** when adding a URL to the search bar the action changes from search to add link. You can see in the video there's a + to indicate this. I am also going to have that be a drop down so you can turn it back to search if necessary.  Though I suspect few people will search with a full URL including the https:// 

-   **Loading Cards**: Then when a link is added, I added a card view for the article, which shows that it is currently processing the link. 

I'm uploading a couple of video to show this, one with a CardView, one with the ListView, and one Showing the flow from the "AddLink" modal.

Videos of features: 
https://discord.com/channels/844965259462311966/1084693944861982791/1137721672603021382

### DIFF
--- a/packages/web/components/elements/LoadingBar.tsx
+++ b/packages/web/components/elements/LoadingBar.tsx
@@ -1,0 +1,78 @@
+import { Box } from './../elements/LayoutPrimitives'
+import { Dispatch, SetStateAction, useEffect, useState } from "react"
+
+type LoadingBarProps = {
+  fillColor: string
+  backgroundColor: string
+  borderRadius: string
+  percentFill?: number
+}
+
+type AnimationStatus = {
+  position: number,
+  transition: string
+}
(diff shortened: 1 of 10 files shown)


jacksonh:
nice! pulling this one to test out locally

jacksonh (approved):

jacksonh:
Going to deploy this one in our staging environment now

PR #2627, opened 2023-08-08
fix existing opted in users not being able to be granted with the feature after increasing the limit

sywhb:


jacksonh:
Nice, lets upgrade to 1500 now too

sywhb:
> Nice, lets upgrade to 1500 now too

upgraded~

jacksonh (approved):

PR #2644, opened 2023-08-11
Attempt to upgrade lerna

jacksonh:


jacksonh:
@sywhb can you try testing this one locally too? I think this is the first step to updating a bunch of other packages....hopefully.

sywhb (approved):
It works locally @jacksonh

PR #2646, opened 2023-08-11
Better handling of PDF downloads on Android

jacksonh:


jacksonh:
This one has been pulled in

PR #2647, opened 2023-08-11
feat/library highlights split

jacksonh:
- Improve the highlights view
- Add a key to the notebook content


jacksonh:
Allotment PR superscedes this one

Issue #2650, opened 2023-08-14
Better handling of zhihu.com


jacksonh:
Example URL: https://www.zhihu.com/question/470000216/answer/3161410353

jacksonh:
@sywhb would probably need your help with this one. Not sure if we have a good format for capturing formums/QA style pages though

jacksonh:
Going to close this one.

PR #2663, opened 2023-08-16
fix/zhihu

sywhb:
- generate test page for zhihu
- Add points for any commas within this paragraph
- do not remove QuestionHeader
- fix title not fetched correctly for some chinese websites
- speically handle zhihu.com


sywhb:
fix #2650

PR #2679, opened 2023-08-21
db migration

sywhb:
- add create library_item sql
- change page_type to enum
- add sql to create library_item_subscriptions db table
- add sql to create library_item_labels db table
- add sql for creating library_item_preview db table
- add libraryItem entity class
- add libraryItemPreview entity class


packages/api/src/entity/library_item_preview.ts
```diff
@@ -0,0 +1,43 @@
+import {
```
jacksonh, line 1: I think we could leave this one out for now. Since we are still kind of working out the social bits.

jacksonh: I also think this might cause some RLS issues, since user B wouldn't be able to access the library_items of user A.

sywhb:
Fix https://omnivore.sentry.io/issues/3992886684/?project=5591542&query=is%3Aunresolved&referrer=issue-stream&stream_index=3 too

jacksonh (approved):

PR #2697, opened 2023-08-28
Add feed items to subscriptions list, allow unsubscribing from feed items on the subscriptions page

jacksonh:
- Handle feed subscriptions on the subscriptions page
- Allow getting all subscritons regardless of type at the same time

### DIFF
--- a/packages/api/src/resolvers/subscriptions/index.ts
+++ b/packages/api/src/resolvers/subscriptions/index.ts
@@ -47,61 +47,60 @@ export const subscriptionsResolver = authorized<
   SubscriptionsSuccessPartial,
   SubscriptionsError,
   QuerySubscriptionsArgs
->(
-  async (
-    _obj,
-    { sort, type = SubscriptionType.Newsletter }, // default to newsletter
-    { claims: { uid }, log }
-  ) => {
-    log.info('subscriptionsResolver')
+>(async (_obj, { sort, type }, { claims: { uid }, log }) => {
+  log.info('subscriptionsResolver')
 
-    analytics.track({
-      userId: uid,
-      event: 'subscriptions',
-      properties: {
-        env: env.server.apiEnv,
-      },
-    })
+  analytics.track({
+    userId: uid,
+    event: 'subscriptions',
+    properties: {
+      env: env.server.apiEnv,
+    },
+  })
 
-    try {
-      const sortBy =
-        sort?.by === SortBy.UpdatedTime ? 'lastFetchedAt' : 'createdAt'
-      const sortOrder = sort?.order === SortOrder.Ascending ? 'ASC' : 'DESC'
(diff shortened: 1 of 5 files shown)


packages/api/src/resolvers/subscriptions/index.ts
```diff
@@ -47,61 +47,56 @@ export const subscriptionsResolver = authorized<
...
-      // only return active subscriptions for newsletter
-      if (type === SubscriptionType.Newsletter) {
-        queryBuilder.andWhere({ status: SubscriptionStatus.Active })
-      }
+    const queryBuilder = getRepository(Subscription)
+      .createQueryBuilder('subscription')
+      .leftJoinAndSelect('subscription.newsletterEmail', 'newsletterEmail')
+      .where({
+        user: { id: uid },
+        status: SubscriptionStatus.Active,
```
sywhb, line 77: We probably need add `status` to the request payload because we need to show `Paused` subscriptions on RSS page

So for newsletters, we can get subscriptions which `status = active` and for rss feeds, we don't put `status` in the query

sywhb (commented):
Most of the changes look good.

I think we also have to update the `getSubscriptions` GraphQL query to include `status` in `pages/settings/subscriptions.tsx`

I will also add a PubSub subscription for the new `PAGE_UPDATED` event in the webhooks

jacksonh:
> Most of the changes look good.
> 
> I think we also have to update the `getSubscriptions` GraphQL query to include `status` in `pages/settings/subscriptions.tsx`
> 
> I will also add a PubSub subscription for the new `PAGE_UPDATED` event in the webhooks

oh so we don't have PAGE_UPDATED now?

sywhb:
> > Most of the changes look good.
> > I think we also have to update the `getSubscriptions` GraphQL query to include `status` in `pages/settings/subscriptions.tsx`
> > I will also add a PubSub subscription for the new `PAGE_UPDATED` event in the webhooks
> 
> oh so we don't have PAGE_UPDATED now?

Yeah, sort of. The code is live but I didn't create a subscription to consume the PubSub event

packages/api/src/resolvers/subscriptions/index.ts
jacksonh: Ah ok, that makes it a little tricky, so i guess right now for feeds if its paused it sets it to UNSUBSCRIBED

jacksonh: Makes sense, I will add some tests for these cases.

sywhb:
@jacksonh I created the PubSub subscription in demo and prod so it should be safe to add the Webhook Event now

jacksonh:
@sywhb ok i think this works now. It returns all RSS regardless of status, and only active newsletters.

sywhb (approved):

PR #2707, opened 2023-08-30
Add Saved Searches,  the ability to add new searches, delete searches, and reorder searches. 

Podginator:
Added saving, editing, and deleting view UI elements and API components.


jacksonh:
man, this is fantastic. No idea how many people have asked for this.

jacksonh:
Looks like the one test failure is my fault, working on a fix, then we can rebase.

packages/db/migrations/0119.do.add_defaults_to_filters.sql
```diff
@@ -0,0 +1,59 @@
...
+
+CREATE OR REPLACE FUNCTION update_filter_position()
+    RETURNS TRIGGER AS $$
+DECLARE
+    new_position INTEGER;
+BEGIN
+    IF (TG_OP = 'DELETE') THEN
+        UPDATE omnivore.filters SET position = position - 1 WHERE user_id = OLD.user_id AND position > OLD.position;
+        RETURN OLD;
+    ELSIF (TG_OP = 'INSERT' and NEW.position is null) THEN
```
Podginator, line 19: Change to be defaultFilter is true rather than this.

jacksonh:
Going to check this one out again today. Works really well for me in testing.

packages/db/migrations/0119.do.add_defaults_to_filters.sql
```diff
@@ -0,0 +1,59 @@
...
+        SELECT COALESCE(MAX(position), 0) + 1 INTO new_position FROM omnivore.filters WHERE user_id = NEW.user_id AND name < NEW.name;
+        UPDATE omnivore.filters SET position = position + 1 WHERE user_id = NEW.user_id AND position >= new_position;
+        NEW.position = new_position;
+        RETURN NEW;
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE 'plpgsql';
+
+INSERT INTO omnivore.filters (user_id, category, name,  filter, position, default_filter)
```
jacksonh, line 29: Hey, so does this one do an insert for all users? I was thinking of handling that in the resolver just returning default values, but this is probably easier.

packages/api/src/resolvers/filters/index.ts
```diff
@@ -166,6 +176,113 @@ export const filtersResolver = authorized<FiltersSuccess, FiltersError>(
...
+  try {
+    const user = await getRepository(User).findOneBy({ id: uid })
+    if (!user) {
+      return {
+        errorCodes: [UpdateFilterErrorCode.Unauthorized],
+      }
+    }
+
+    const filter = await getRepository(Filter).findOne({
+      where: { id },
```
jacksonh, line 240: I think we should have user id here in the where clause too, otherwise users could "guess" uuids and query for other users filters I think.

packages/api/src/services/create_user.ts
```diff
@@ -85,6 +86,9 @@ export const createUser = async (input: {
           group: invite.group,
         })
       }
+
+      await createDefaultFiltersForUser(t)(user.id)
```
jacksonh, line 90: I suppose this could also be done in postgres with a transaction

jacksonh: I mean trigger.

jacksonh (commented):
Can we also add the "Oldest First" default filter `sort:saved-asc`

jacksonh (changes requested):
Just had one comment on a query, and then hope we can add in `Oldest First` plus need to think about whether we should use a trigger to create new users filters.

packages/api/src/services/create_user.ts
jacksonh: OK, I don't really have a preference here. So I think this approach works. The one thing I'd say is it should probably be wrapped in a try/catch. I wouldn't want the user creation to fail.

Podginator:
Probably won't be able to make these changes until tomorrow, but on it 👍

packages/db/migrations/0119.do.add_defaults_to_filters.sql
jacksonh: I think your approach is right, its quite complex to handle the null case I think when you consider partially set default searches.

jacksonh:
Nice, going to deploy this to our staging environment and do some testing to see how quickly the filters load in the left menu. I've been thinking we could easily cache some of these items with SWR, like labels/filters/subscriptions which rarely change.

PR #2715, opened 2023-08-31
Ignore link density checks in newsletters

jacksonh:


packages/readabilityjs/Readability.js
```diff
@@ -88,6 +88,7 @@ function Readability(doc, options) {
   this._articleDir = null;
   this._languageCode = null;
   this._attempts = [];
+  this._testLinkDensity = options.testLinkDensity || true
```
jacksonh, line 91: I think we need to change this to  ignoreLinkDensity

PR #2721, opened 2023-09-01
Fix Loading: Ensure that image is loaded when retrieving images

Podginator:
Small Fix: Ensure that when a user adds an item, it displays the newly loaded image, rather than the Omnivore fallback

### DIFF
--- a/packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
+++ b/packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
@@ -224,7 +224,7 @@ export function HomeFeedContainer(): JSX.Element {
 
           if (link && link.state != 'PROCESSING') {
             const updatedArticle = { ...item }
-            updatedArticle.node = { ...item.node, ...link, image: "https://cdn.arstechnica.net/wp-content/uploads/2023/09/cpglitch-800x450.jpg" }
+            updatedArticle.node = { ...item.node, ...link }
             updatedArticle.isLoading = false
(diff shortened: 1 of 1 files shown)


packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
```diff
@@ -224,7 +224,7 @@ export function HomeFeedContainer(): JSX.Element {
 
           if (link && link.state != 'PROCESSING') {
             const updatedArticle = { ...item }
-            updatedArticle.node = { ...item.node, ...link }
+            updatedArticle.node = { ...item.node, ...link, image: "https://cdn.arstechnica.net/wp-content/uploads/2023/09/cpglitch-800x450.jpg" }
```
Podginator: Blerg.

packages/web/components/templates/homeFeed/HomeFeedContainer.tsx
```diff
@@ -1229,7 +1229,7 @@ function LibraryItems(props: LibraryItemsProps): JSX.Element {
           data-testid="linkedItemCard"
           id={linkedItem.node.id}
           tabIndex={0}
-          key={linkedItem.node.id}
+          key={linkedItem.node.id + linkedItem.node.image}
```
Podginator, line 1232: The key was causing the element to not rerender after the article had fully loaded. By appending the image url it will now rerender with the correct image.

jacksonh (approved):

PR #2737, opened 2023-09-07
fix/api switch to posthog

jacksonh:
- Remove identify call
- Switch backend to send events directly to posthog


sywhb:
Close this PR as we opened a new one #3580

PR #2796, opened 2023-09-24
[Bug-Fix] Fix Editing Saved-Searches on Mobile View

Podginator:


packages/web/pages/settings/saved-searches.tsx
```diff
@@ -349,7 +352,7 @@ export default function SavedSearchesPage(): JSX.Element {
               <Box>
                 <StyledText style="fixedHeadline">Saved Searches </StyledText>
               </Box>
-              <InfoLink href="https://docs.omnivore.app/using/search.html" />
+              <InfoLink href="/help/search" />
```
jacksonh, line 355: I think maybe you need a rebase?

Podginator: Merged the wrong thing here I think, will fix

Issue #2801, opened 2023-09-25
Handle text only newsletters


jacksonh:
Some newsletters don't contain an HTML element, for these, if text is present we should create a text element using markdown to HTML

jacksonh:
This is where the error manifests: https://github.com/omnivore-app/omnivore/blob/4bd2ac2ac37fe5798741cdc8ce6921aed165c18c/packages/content-handler/src/content-handler.ts#L166

sywhb:
Thanks for the details. I'm working on it @jacksonh

PR #2802, opened 2023-09-25
convert plain text newsletter email content to html using markdown-to-html and save it if html is not available

sywhb:


sywhb:
Fix #2801

packages/inbound-email-handler/src/index.ts
```diff
@@ -163,19 +169,24 @@ export const inboundEmailHandler = Sentry.GCPFunction.wrapHttpFunction(
           await handleAttachments(to, subject, attachments, receivedEmailId)
           return res.send('ok')
         }
+
+        // convert text to html if html is not available
+        const content = html || plainTextToHtml(text)
```
jacksonh, line 174: nice trick

jacksonh (approved):

PR #2809, opened 2023-09-27
fix getting 403 when parsing some rss url by setting a user agent in the request

sywhb:


packages/rss-handler/src/index.ts
```diff
@@ -118,9 +118,16 @@ Sentry.GCPFunction.init({
 
 const signToken = promisify(jwt.sign)
 const parser = new Parser({
+  timeout: 60000, // 60 seconds
```
jacksonh, line 121: 😬

jacksonh (approved):

packages/rss-handler/src/index.ts
sywhb: @jacksonh the default timeout is 60 seconds. I just try to make it more clear here

jacksonh: Yeah sounds good. I think long time out instead of failing fast and repeating is probably the right thing for RSS so this is fine. I just hate to see timeouts > 4s. :)

Issue #2750, opened 2023-09-09
CSV import does not work for me with Firefox


Cito:
When trying to import via CSV with Firefox, I always get the error message "No upload URL available". On Chrome, it works.

jacksonh:
Interesting, it might have been a random API failure as I haven't been able to replicate this one. Can i ask your firefox version?

Cito:
I have tried to replicate this today, and found that it is caused by a strange behavior of Firefox that happens only on Windows, and maybe only if Excel is installed. That's probably why you couldn't replicate it.

Already submitted PR #2848 to fix this, adding a description with more info on the cause of this issue.

jacksonh:
Closing this one out as i think its all fixed.

PR #2848, opened 2023-10-03
Fix import with Firefox on Windows (#2750)

Cito:
This PR fixes the issue reported in #2750.

The problem here is that Firefox under Windows may report the content type of a CSV file as `application/vnd.ms-excel` instead of `text/csv` which will then be rejected by Omnivore.

Users on [Stackoverflow](https://stackoverflow.com/questions/7076042/what-mime-type-should-i-use-for-csv/28233618#28233618) and [Mozilla](https://support.mozilla.org/en-US/questions/1401889) confirm this behavior.

The PR adds a pragmatic workaround that considers `application/vnd.ms-excel` the same as `text/csv` if the file name ends with `.csv`. It also improves the error messages when the content type is not valid.

### DIFF
--- a/packages/web/components/templates/UploadModal.tsx
+++ b/packages/web/components/templates/UploadModal.tsx
@@ -122,7 +122,14 @@ export function UploadModal(props: UploadModalProps): JSX.Element {
   const uploadSignedUrlForFile = async (
     file: UploadingFile
   ): Promise<UploadInfo> => {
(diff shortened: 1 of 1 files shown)


jacksonh:
oh awesome, we would have never found this. Thank you!

jacksonh (approved):

PR #2860, opened 2023-10-05
fix demo user not able to login when self hosting

### LINKED ISSUE #2847: Cannot login with demo password
The README says [here](https://github.com/omnivore-app/omnivore#3-login-with-the-test-account) that the test account account is set up with the password "demo", but [the backend code here](https://github.com/omnivore-app/omnivore/blob/089fbe23a67c5fe9b488b27316b230f0fd342a28/packages/api/src/routers/auth/auth_router.ts#L420) does not accept passwords with a length smaller than 8, and in fact I cannot login with the password "demo" because of that. It only worked after removing that restriction. Am I doing something wrong?

sywhb:
change password to demo_password

Fix #2847

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,29 +56,6 @@ services:
     ports:
       - "9200:9200"
 
-  kibana:
-    image: docker.elastic.co/kibana/kibana:7.17.1
-    container_name: "omnivore-kibana"
-    environment:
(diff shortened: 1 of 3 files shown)


jacksonh (approved):

Cito:
Thanks for fixing this.

PR #2864, opened 2023-10-05
Android: Use XML resources for copy

remychantenay:
## Summary
Moving the copy to XML resources to later facilitate localization. 

## Note(s)
* I did not change any of the copy. It's merely a lift-and-shift operation :)
* I also did not try translating things into Spanish, even less so in Chinese :D (Happy to create issues for these, however)
* I am human, so a mistake could have sneaked in, but I double-checked and didn't see any.


jacksonh:
OMG, thank you

jacksonh (approved):

jacksonh:
This is great, thanks. FYI we locked main today as we are doing a big data migration, but I'll merge this ASAP. Probably before the weekend.

Issue #2820, opened 2023-09-30
Choose highlight color using Android app.  


theotheo:
Tell me please, is this possible to use another color that yellow to highlight text with Android app?

jacksonh:
Hi, right now its not possible to change the color in the native apps.

remychantenay:
@jacksonh out of curiosity, do you have designs for this? Or even a vague prototype? A drawing? 😄

jacksonh:
Hey, that's actually part of the hold up, but the Web design is probably really close to what we want. Two screenshots attached
<img width="790" alt="Screenshot 2023-10-05 at 20 56 42" src="https://github.com/omnivore-app/omnivore/assets/75189/95e8b953-e088-4f12-b287-4d97b89d0493">
<img width="841" alt="Screenshot 2023-10-05 at 20 56 52" src="https://github.com/omnivore-app/omnivore/assets/75189/99d58d4a-7d40-4143-8f4a-c4abd3effa01">

Lorite:
Hi. Are there any updates on this issue? 😊

remychantenay:
Hey @Lorite, 

I sort of put that one on the back burner and eventually forgot. I am pretty busy these days so it's hard for me to find time to contribute, but I will try to squeeze that in for a coming weekend.

From what I remember, I was almost there.

PR #2874, opened 2023-10-06
Android: Add Highlight color palette for web reader

remychantenay:
Hey 👋 

This goes toward #2820 

## Note(s)
- The colours match what's in the web version.
- Only added to the web reader for now.
- The default colour is yellow.
- The selected colour is bound to the article read (i.e., not persisted).
- Happy to discuss implementation/ui/ux details, I literally just went with the flow 😃 
- It only partially works (see below)

## Important
I am facing an issue where the highlight is first created with the default colour, but if leaving the article and opening it back again, then it appears with the colour manually selected (see attached screen record below.)

My guess is that the highlight is "pre-created" but I couldn't pinpoint where? There is something that escapes me in the flow 🤔 

Any pointers would be much appreciated!

https://github.com/omnivore-app/omnivore/assets/4558395/5eb5ad48-4773-4d38-a484-678d0318c46b

### DIFF
--- a/android/Omnivore/app/src/main/java/app/omnivore/omnivore/dataService/HighlightActionHandlers.kt
+++ b/android/Omnivore/app/src/main/java/app/omnivore/omnivore/dataService/HighlightActionHandlers.kt
@@ -12,7 +12,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import java.util.*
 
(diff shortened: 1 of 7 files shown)


jacksonh:
I'll merge this one then play around with it a bit. There's definitely a lot of back and forth between the webview and native code that I can't fully remember how it works.

Issue #2866, opened 2023-10-05
Failed to delete Feeds


Nexus-Gits:
when i click to delete any feed it only show paused and deleted from list. also the feed list also show in side bar.

Also after edit tags in android mobile browser its struck the page. it only work after hard refresh the page.

sywhb:
Thanks @Nexus-Gits, I know where it went wrong and fixing it now

PR #2878, opened 2023-10-07
delete rss subscription row from db when unsubscribe

sywhb:
@jacksonh I think this is a regression when a newsletter subscription is deleted, we mark it as unsub but for rss subscriptions, we should just delete it from the database.


sywhb:
Fix #2866

PR #2879, opened 2023-10-07
allow filter by wordsCount

### LINKED ISSUE #2696: Sort by estimated reading time
Hello! I've just switched from Pocket to Omnivore and love it. Thanks to all involved!

I would love to be able to sort the main inbox & library views by estimated reading time. There are a couple of use cases for this.

The first is that I've got a pretty big reading backlog, and sorting by reading time would let me read the shortest items first, thus letting me clear off a lot of items in one reading session, making me feel like I'm making good progress.

The second is that sometimes I have a rough idea of how much time I have, and I'd like to be able to read something, say, about 20 minutes long. Sorting by read time would make such a thing easier to find. I guess _filtering_ by time would also address this.

sywhb:
syntax: `wordsCount:${operator}${integer}`;
allowed operator: `>`, `>=`, `<`, `<=`;
example: `wordsCount:<=1000` searches for items with words count less or equal than 1000
we could also combine them like `wordsCount:>=1000 wordsCount<5000`

Fix #2696


jacksonh:
nice, would be awesome if we can do the same with readPosition

sywhb:
> nice, would be awesome if we can do the same with readPosition

Yeah, I will add it in this PR

packages/api/src/services/library_item.ts
```diff
@@ -117,27 +132,38 @@ const buildWhereClause = (
         break
       case InFilter.SUBSCRIPTION:
         queryBuilder
-          .andWhere('library_item.subscription IS NOT NULL')
           .andWhere("NOT ('library' ILIKE ANY (library_item.label_names))")
-          .andWhere('library_item.archived_at IS NULL')
+          .andWhere({
```
jacksonh, line 136: nice

jacksonh (approved):

Issue #2722, opened 2023-09-02
Feature Request: Adaptive, Themed Icon for Android 13+ Launcher APIs


raygervais:
Firstly,I want to say how much I love the app so far! It's quickly become one of my favorites and has made it's way onto my minimalist home screen. Upon adding it, I noticed the app doesn't-yet implement the [adaptive icon API](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive#add_your_adaptive_icon_to_your_app) for those who are on the Material You train -pixel users, really.

I'd love to help add this if no one else is interested, but don't have the design / graphics skills to produce a monochrome-variant of the current icon found under `/android/Omnivore/app/src/main/res/mipmap-*`. If provided that, I'd be happy to provide the additional XML changes required to support this.

Thanks!

jacksonh:
🔥

PR #2884, opened 2023-10-08
Add support for Android themed icon - #2722

prof18:
Hey 👋 

I started using themed icons on my device, I saw the open issue (#2722), and I couldn't resist :)

<div>

</div>

The SVG that I've used to generate the monochromatic version is the favicon that I found [here](https://github.com/omnivore-app/omnivore/blob/main/packages/web/public/favicon.svg)


jacksonh:
Wow, so many people have been asking for this.

PR #2893, opened 2023-10-09
Fix response url in the share target

jacksonh:


jacksonh:
ok i think this is all working now.

jacksonh:
Going to merge this and continue testing on demo.

Issue #1481, opened 2022-11-30
Display validation info for long label descriptions


jacksonh:
Looks like we have a max length of 100 but no client side validation, would be nice to show users errors on the client side here: https://sentry.io/organizations/omnivore/issues/3752841002/?project=5466498&query=is%3Aunresolved&referrer=issue-stream

PR #2901, opened 2023-10-10
iOS: Add length indicator to new label textfield

remychantenay:
## Summary
* Started working on this after seeing #1481 (although it refers to label description.)
* Adding a length indicator under the TextField for label creation.
* Adding guard rails to prevent users from even trying to create label names that are [too long](https://github.com/omnivore-app/omnivore/blob/29c878605709b7039f72cdbc7770dbc268395e3c/packages/api/src/schema.ts#L1455).

## Additional Note
I am aware that there are two schools of thought; some think validation should only be performed in one place, and others might not agree. In this case, I believe this makes for a better user experience (currently, creating a new label that is too long is a very strange experience; it looks like it works at first, but actually doesn't.)

Anyway, let me know what you think. If you think that's valuable, I am happy to add that to Android as well.

## Screen Record
https://github.com/omnivore-app/omnivore/assets/4558395/cf417026-86cf-49f0-ae89-96566aa18117

### DIFF
--- a/apple/OmnivoreKit/Sources/Views/Resources/zh-Hans.lproj/Localizable.strings
+++ b/apple/OmnivoreKit/Sources/Views/Resources/zh-Hans.lproj/Localizable.strings
@@ -20,6 +20,7 @@
 "labelsViewAssignNameColor" = "指定名称和颜色。";
 "createLabelMessage" = "创建新标签";
 "labelsPurposeDescription" = "使用标签创建精选的链接集合。";
+"labelNamePlaceholder" = "标签名称";
 
 // Manage Account View
 "manageAccountDelete" = "删除帐户";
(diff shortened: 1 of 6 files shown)


apple/OmnivoreKit/Sources/Views/Resources/zh-Hans.lproj/Localizable.strings
```diff
@@ -20,6 +20,7 @@
 "labelsViewAssignNameColor" = "指定名称和颜色。";
 "createLabelMessage" = "创建新标签";
 "labelsPurposeDescription" = "使用标签创建精选的链接集合。";
+"labelNamePlaceholder" = "标签名称";
```
remychantenay, line 23: Please, feel free to correct me here, I don't really know what I am doing 🤣

jacksonh: haha, asked and its right!

PR #2914, opened 2023-10-12
Allow nullable on the highlight positions

jacksonh:


packages/db/migrations/0133.do.remove_highlight_position_not_nulls.sql
```diff
@@ -4,4 +4,17 @@
 
 BEGIN;
 
+ALTER TABLE omnivore.highlight
+    ALTER COLUMN highlight_position_percent DROP NOT NULL;
```
sywhb, line 8: @jacksonh why do we drop `NOT NULL` here?

sywhb: We probably should set the `percent = 0` if we want to reset it

jacksonh:
Closing this one as @sywhb did a proper fix.

Issue #2641, opened 2023-08-10
run `trim()` on user input text when creating a new label


kquinsland:
This is a _super minor_ bug/issue that I've noticed.
I hit this "bug" on iOS but it probably applies to other platforms as well.

**EDIT:** Yep, same behavior with the [android app](https://play.google.com/store/apps/details?id=app.omnivore.omnivore) as well.

When creating a new label, the keyboard will try to predict a word. If you select the prediction, the text will be updated to the selected word **and an additional white space**.

When the user clicks the "create this label" button, consider trimming the trailing white space off.

To illustrate this, see attached image.

![omnivore-label-issue](https://github.com/omnivore-app/omnivore/assets/1808375/9dc13cde-6fb0-4ab1-9f94-753ef4ea8208)

**To reproduce:**

- Start the "create new label" flow
- Type `Examp` and by now your keyboard should suggest `Example`. Select the suggestion.
- Notice that the keyboard has helpfully added a space as it expects you to keep typing new words.

jacksonh:
- [ ] iOS: https://github.com/omnivore-app/omnivore/pull/2642
- [ ] Android:
- [ ] Web:

PR #2920, opened 2023-10-12
Android: Add label name length validation upon creation

remychantenay:
## Summary
Adding validation for the label name length. Towards #1481 

While similar to #2901 on iOS, the implementation is different due to the label creation works. I couldn't think of a better way to implement this check in terms of value-to-effort, but I'm happy to hear other ideas?

## Screen Record
https://www.loom.com/share/c691155bf0164a9e8b748b909707db56?sid=b9741b4b-ebf3-47ab-8e08-77a1acd2118f


jacksonh:
Thanks, yeah the label creation code on android is pretty messy too

jacksonh:
pulling this one to test out

remychantenay:
@jacksonh just FYI, since this is touching the exact same part of the code, I pushed another commit (https://github.com/omnivore-app/omnivore/pull/2920/commits/668485ba15b75b04a76b34942e707b5bcd31d881) that also fixes #2641 for Android.

That way we can close the issue :)

PR #2925, opened 2023-10-13
fix bulk action slow query

sywhb:


packages/api/src/resolvers/article/index.ts
```diff
@@ -792,6 +791,12 @@ export const bulkActionResolver = authorized<
       },
     })
 
+    // parse query
+    const searchQuery = parseSearchQuery(query)
+    if (searchQuery.ids.length > 100) {
```
jacksonh, line 796: 👍

packages/api/src/services/library_item.ts
```diff
@@ -133,37 +120,35 @@ const buildWhereClause = (
       case InFilter.SUBSCRIPTION:
         queryBuilder
           .andWhere("NOT ('library' ILIKE ANY (library_item.label_names))")
-          .andWhere({
-            subscription: Not(IsNull()),
-            archivedAt: IsNull(),
-          })
+          .andWhere('library_item.archived_at IS NULL')
+          .andWhere('library_item.subscription IS NOT NULL')
```
jacksonh, line 124: nice, seems like that ILIKE above was a bug.

jacksonh (approved):

packages/api/src/services/library_item.ts
sywhb: sql looks okay. let me test it in the db

sywhb: verified in database and it is not a bug

PR #2926, opened 2023-10-13
Add ILIKE queries to all field queries

jacksonh:
This adds OR ILIKE to our field queries which resolves things
like `site:omnivore.app` not matching correctly. The TSV match
doesn't work in this case because it will split `omnivore.app`
into word vectors that dont match well when given exact searches.


packages/api/src/services/library_item.ts
```diff
@@ -238,10 +239,16 @@ const buildWhereClause = (
     args.matchFilters.forEach((filter) => {
       const param = `match_${filter.field}`
       queryBuilder.andWhere(
-        `websearch_to_tsquery('english', :${param}) @@ library_item.${filter.field}_tsv`,
-        {
-          [param]: filter.value,
-        }
+        new Brackets((qb) => {
```
sywhb, line 242: nice

packages/api/src/services/library_item.ts
```diff
@@ -238,10 +239,16 @@ const buildWhereClause = (
...
+          qb.andWhere(
+            `websearch_to_tsquery('english', :${param}) @@ library_item.${filter.field}_tsv`,
+            {
+              [param]: filter.value,
+            }
+          ).orWhere(`${filter.field} ILIKE :value`, {
+            value: `%${filter.value}%`,
```
sywhb, line 249: i'm wondering if just prefix matching is enough for us like`${filter.value}%`?

packages/api/src/services/library_item.ts
```diff
@@ -238,10 +239,16 @@ const buildWhereClause = (
...
-        {
-          [param]: filter.value,
-        }
+        new Brackets((qb) => {
+          qb.andWhere(
+            `websearch_to_tsquery('english', :${param}) @@ library_item.${filter.field}_tsv`,
+            {
+              [param]: filter.value,
+            }
+          ).orWhere(`${filter.field} ILIKE :value`, {
```
sywhb, line 248: probably best to add alias here in case typeorm is stupid

sywhb (commented):
hey @jacksonh, LGTM. 
i made few comments

packages/api/src/services/library_item.ts
jacksonh: Oh i had a little trouble with this because fieldname is like `site_name` I couldn't do:

orWhere({
   site_name: value
})

because typeORM expects `siteName: value`

packages/api/src/services/library_item.ts
jacksonh: yeah i think you are right. Updating

jacksonh: oh actially, i wonder if prefix and suffix is needed for items like author `author:Harper` for `author:Jackson Harper`

jacksonh:
Still working on the tests a bit

packages/api/src/services/library_item.ts
sywhb: yeah, we'd best use raw sql here like `library_item.${filter.field} ILIKE :value` because typeorm is stupid and the update query for bulk action would fail if we do object literal 😢

packages/api/src/services/library_item.ts
sywhb: yeah, you are right. we should do both prefix and suffix.

thank you!

PR #2958, opened 2023-10-17
Fix password in README

dentropy:
Fix password in readme, password is demo_password not demo, see packages/db/setup.sh for reference

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Open <http://localhost:3000> and confirm Omnivore is running
 
 ### 3. Login with the test account
 
-During database setup docker compose creates an account `demo@omnivore.app`, password: `demo`.
+During database setup docker compose creates an account `demo@omnivore.app`, password: `demo_password`.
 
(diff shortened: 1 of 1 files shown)


jacksonh:
Thanks!

Hrishikesh156:
@jacksonh I'm unable to login with demo@omnivore.app, password: demo_password. Do I need to run any other command than docker compose up
image

dentropy:
@Hrishikesh156 You need to make sure the migrations ran for the postgres database, see the omnivore-migrate docker container that was run

image

dentropy:
@Hrishikesh156 you can check if the database migrations ran by running the following commands,

``` bash

docker exec -it omnivore-postgres bash
psql -d omnivore -U app_user
\dt
exit
psql -d omnivore -U postgres
\dt
select * from schemaversion;

```

image

Hrishikesh156:
@dentropy from docker ps -a, I guess the migrations ran for the postgres database

image

Apparently it couldn't find any relations in the omnivore database

image

PR #2959, opened 2023-10-18
Only use signed URLs for PDF attachments

jacksonh:


sywhb (approved):

packages/api/src/resolvers/upload_files/index.ts
```diff
@@ -142,7 +140,7 @@ export const uploadFileRequestResolver = authorized<
         const uploadFileId = uploadFileData.id
         const item = await createLibraryItem(
           {
-            originalUrl: isFileUrl(input.url) ? publicUrl : input.url,
+            originalUrl: input.url,
```
sywhb, line 143: @jacksonh we should change this line too
`originalUrl: isFileUrl(input.url) ? attachmentUrl : input.url`

sywhb (changes requested):

sywhb (approved):
🚢

PR #2962, opened 2023-10-18
Cache and check feed checksums to reduce fetching

jacksonh:


packages/rss-handler/src/index.ts
```diff
@@ -21,10 +23,35 @@ function isRssFeedRequest(body: any): body is RssFeedRequest {
   )
 }
 
+export const fetchAndChecksum = async (url: string) => {
+  try {
+    const response = await axios.get(url, {
```
sywhb, line 28: it's probably better to add timeout too

packages/rss-handler/src/index.ts
```diff
@@ -121,15 +149,12 @@ const parser = new Parser({
...
-    'User-Agent':
-      'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36',
-    Accept:
-      'application/rss+xml, application/rdf+xml;q=0.8, application/atom+xml;q=0.6, application/xml;q=0.4, text/xml;q=0.4',
+    item: [
+      ['link', 'links', { keepArray: true }],
+      'published',
+      'updated',
+      'created',
+    ],
```
sywhb, line 157: the custom fields for feed are missing
`feed: ['dc:date', 'lastBuildDate', 'pubDate']`

sywhb (commented):
@jacksonh i have made some comments

packages/rss-handler/src/index.ts
jacksonh: oh yes!

packages/rss-handler/src/index.ts
jacksonh: weird i wonder why this diff got all messed up, thanks for catching

sywhb (approved):
LGTM

PR #2966, opened 2023-10-18
WIP: Start looking at selfhosting with k8s and tailnet

lawrencegripper:
Started working on this as I wanted to host a version of this myself via Tailscale and k8s. 

Mostly this is me randomly trying stuff as it's been ages since I've done anything with node and ts ... feel free to ignore.

So far the hacking has me at/fixes:
- Use postgres with vector extension (why does this work in compose version without this?)
- Add CLIENT_URL to CORS accepted list
- Add Web UI to the self-host image
- [maybe] Skip email verification
- [maybe] Auto comfigure schema

Broken:
- Using the Firefox extension, doesn't seem to accept self-host details
- Login on iOS, not sure why, get "something went wrong"

### DIFF
--- a/self-hosting/deployment.yaml
+++ b/self-hosting/deployment.yaml
@@ -1,0 +1,143 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: omnivore
+  namespace: omnivore
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      app: omnivore
+  template:
+    metadata:
+      labels:
+        app: omnivore
+    spec:
+      containers:
+        - name: omnivore-web
+          command:
+            - /bin/sh
+            - -c
+            - "yarn workspace @omnivore/web start -p 3000"
+          image: ghcr.io/lawrencegripper/omnivore:17-10-23-11
+          imagePullPolicy: Always
+          ports:
+            - containerPort: 3000
+          env:
+            - name: NEXT_PUBLIC_APP_ENV
+              value: prod
+            - name: NEXT_PUBLIC_BASE_URL
+              value: http://omnivore-web.some-tailnet.ts.net
+            - name: NEXT_PUBLIC_SERVER_BASE_URL
+              value: http://omnivore-web.some-tailnet.ts.net:4000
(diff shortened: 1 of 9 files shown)


jacksonh:
wow this is fantastic, thanks

`Using the Firefox extension, doesn't seem to accept self-host details` -- most likely this is the content security policy. I'm not sure if these keep be updated automatically though :-/

jacksonh:
Just addressing a couple things:

```
Use postgres with vector extension (why does this work in compose version without this?)
```
Yeah I think we can get away without this now. It is something we will use in the future, but probably try to do it in a way thats easier to opt in/out of.

```
Add Web UI to the self-host image
```
Nice, do you need to do anything with routing? Or will you use a k8s ingress routing rule?
```

```
[maybe] Skip email verification
```
For this we should probably have an envar like DISABLE_EMAIL_VERIFICATION that needs to be opted into.

File: self-hosting/deployment.yaml [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+          env:
+            - name: ELASTIC_PASSWORD
+              value: some-postgres-passwordelastic
+            - name: discovery.type
+              value: single-node
+          volumeMounts:
+            - mountPath: /usr/share/elasticsearch/data
+              name: elastic-data
+            # - mountPath: /usr/share/elasticsearch/config
+            #   name: elastic-config
```
fr3fou, line 126: imo we shouldn't be deploying Elasticsearch as a sidecar to this Deployment (or let alone as a Deployment instead of StatefulSet). I think the way to go would be to make a stand-alone Omnivore Helm Chart which has a dependency on some Elasticsearch Helm Chart. Either that or we leave it up to the user to manage / have an Elasticsearch instance deployed.

grapemix: The same for pg too. Lost of self-hoster already installed cloudnative-pg/Zalando PG or something similar. Also, splitting web, api and contect-fetch is more k8s way. But this commit is a good start, good job! Thanks for your work.

lawrencegripper:
With workload and family stuff I'm not going to be able to finish this off.

Happy for someone to pickup, for now I'll close up

PR #2969, opened 2023-10-19
Group RSS subscriptions by feed URL and filter by scheduled time

sywhb:
- add scheduledAt to subscriptions
- add scheduledAt to the task payload
- Group RSS subscriptions by feed URL and create cloud task for the groups


packages/rss-handler/src/index.ts
```diff
@@ -185,162 +220,182 @@ const getLink = (links: RssFeedItemLink[]) => {
...
+      publishedAt < new Date(Date.now() - 24 * 60 * 60 * 1000)
+    ) {
+      console.log('Skipping old feed item', item.link)
+      continue
+    }
 
-      let lastItemFetchedAt: Date | null = null
-      let lastValidItem: Item | null = null
+    const created = await createSavingItemTask(userId, feedUrl, item)
+    if (!created) {
```
jacksonh, line 302: I noticed in the other spot this fails we throw an exception, but here we continue. I guess because need to keep processing the other items

packages/rss-handler/src/index.ts
```diff
@@ -185,162 +220,182 @@ const getLink = (links: RssFeedItemLink[]) => {
...
+    lastItemFetchedAt = lastValidItem.isoDate
+      ? new Date(lastValidItem.isoDate)
+      : new Date()
+  }
 
-        // Max limit per-feed update
-        if (itemCount > 99) {
-          continue
-        }
+  const updateFrequency = (feed['syn:updateFrequency'] ||
```
jacksonh, line 335: I have a feeling we might need to put this in a function and a try/catch block, because I suspect we get garbage in these values sometimes and we wouldn't want to completely fail if something weird was in updateFrequency

packages/api/test/resolvers/subscriptions.test.ts
```diff
@@ -347,4 +348,91 @@ describe('Subscriptions API', () => {
...
+
+  describe('Subscribe API', () => {
+    const query = (
+      name: string | null,
+      url: string | null,
+      subscriptionType: SubscriptionType | null
+    ) => `
+      mutation {
+        subscribe(input: {
+          name: ${name ? `"${name}"` : null}
```
jacksonh, line 360: its probably ok in some tests but I've noticed its not good to use string interop in the GQL as it can give really unexpected results sometimes like maybe an empty string for something that was a null value.

jacksonh (approved):

packages/rss-handler/src/index.ts
sywhb: Yeah I will put it in a try/catch block

packages/api/test/resolvers/subscriptions.test.ts
sywhb: Yeah, I was hesitated to write this 😞 Gonna change it for sure

packages/rss-handler/src/index.ts
```diff
@@ -155,10 +164,56 @@ const parser = new Parser({
...
+    return 1
+  }
+
+  const frequency = parseInt(updateFrequency, 10)
+  if (isNaN(frequency)) {
+    return 1
+  }
+
+  return frequency
+}
```
sywhb, line 193: @jacksonh I added a `isNaN` to check invalid input here

packages/api/test/resolvers/subscriptions.test.ts
```diff
@@ -347,4 +348,108 @@ describe('Subscriptions API', () => {
...
+        })
+
+        after(async () => {
+          await deleteSubscription(existingSubscription.id)
+        })
+
+        it('returns an error', async () => {
+          const res = await graphqlRequest(query, authToken, {
+            input: { url, subscriptionType },
+          }).expect(200)
```
sywhb, line 406: I will put query and variables separately in GraphQL

packages/api/src/resolvers/subscriptions/index.ts
```diff
@@ -203,49 +186,83 @@ export const subscribeResolver = authorized<
...
 
+      // create a cloud task to fetch rss feed item for resub subscription
+      await enqueueRssFeedFetch({
+        userIds: [uid],
+        url: input.url,
+        subscriptionIds: [updatedSubscription.id],
+        scheduledDates: [new Date()], // fetch immediately
+        fetchedDates: [updatedSubscription.lastFetchedAt || null],
+        checksums: [updatedSubscription.lastFetchedChecksum || null],
+      })
```
sywhb, line 216: Re-subscribe to a rss feed if its status is `UNSUB`

sywhb:
Hey @jacksonh, I have added some changes. Could you please take a look at this PR again?

Thank you

packages/api/src/generated/schema.graphql
```diff
@@ -2058,9 +2058,8 @@ enum SubscribeErrorCode {
 }
 
 input SubscribeInput {
-  name: String
   subscriptionType: SubscriptionType
-  url: String
+  url: String!
```
jacksonh, line 2062: I think this is probably safe but technically an API change. Is there any case right now where the client would be sending a null URL?

packages/api/src/resolvers/subscriptions/index.ts
```diff
@@ -203,49 +186,83 @@ export const subscribeResolver = authorized<
...
+    const existingSubscription = await getRepository(Subscription).findOneBy({
+      url: input.url,
+      user: { id: uid },
+      type: SubscriptionType.Rss,
+    })
+    if (existingSubscription) {
+      if (existingSubscription.status === SubscriptionStatus.Active) {
         return {
-          errorCodes: [SubscribeErrorCode.ExceededMaxSubscriptions],
+          errorCodes: [SubscribeErrorCode.AlreadySubscribed],
```
jacksonh, line 198: Will probably try to handle this in the client in the future to give a useful message.

packages/api/src/services/subscriptions.ts
```diff
@@ -199,5 +201,16 @@ export const createSubscription = async (
     unsubscribeMailTo,
     lastFetchedAt: new Date(),
     type: subscriptionType,
+    url,
   })
 }
+
+export const deleteSubscription = async (id: string): Promise<DeleteResult> => {
```
jacksonh, line 208: Is this new? I thought we were always soft deleting subscriptions now

jacksonh (commented):
nice work. Just one question about the deletion.

packages/api/src/generated/schema.graphql
sywhb: I think only the Web client would make this API call and it always sends a not-null `URL`

packages/api/src/resolvers/subscriptions/index.ts
sywhb: The Web client handles the error code and gives some meaningful error messages for subscriptions

packages/api/src/services/subscriptions.ts
sywhb: Yeah, this is new and I only use it in the tests for clean-up

jacksonh: ohhh haha, great

jacksonh (approved):

jacksonh (approved):

PR #2976, opened 2023-10-20
Add a longer rate limit window on createaccount/reset password, reduce api rate limit hits

jacksonh:


packages/api/src/routers/auth/auth_router.ts
```diff
@@ -80,6 +81,14 @@ export const isValidSignupRequest = (obj: any): obj is SignupRequest => {
   )
 }
 
+// The hourly limiter is used on the create account,
+// and reset password endpoints
+// this limits users to five operations per an hour
+const hourlyLimiter = rateLimit({
+  windowMs: 60 * 60 * 1000,
+  max: 5,
+})
```
sywhb, line 90: Perhaps we would like to skip preflight requests and test requests?
`skip: (req) => req.method === 'OPTIONS' || env.dev.isLocal`

sywhb: I'm not sure if 5 per hour is too low for `email-signup` endpoint because we do not have frontend validation for email address and password

sywhb (commented):
LGTM. I have a few questions though

packages/api/src/routers/auth/auth_router.ts
jacksonh: Hey, so I added a skip for dev environment. This endpoint wont actually have the preflight checks because its a form redirect not client -> server request. 

Regarding the 5 an hour limit, I guess we can monitor. The problem is, if they type their email wrong, they probably wont realize anyways, so they will be checking their email waiting for our verification email.

sywhb: Got it. This makes sense

sywhb (approved):

PR #2982, opened 2023-10-21
modified package.json

UDanV:


jacksonh:
I think this allows it to be listed in npm, but doesn't actually list it, right? I'm not sure if anyone would want to install omnivore packages this way, but we should make a client package for connecting to the API

jacksonh:
I had to revert this one actually as we got an error that workspaces are not available on public packages after merging.

PR #2992, opened 2023-10-23
perf: limit search item results to 100 a time and fetch labels and recommendations only needed

sywhb:


jacksonh (approved):
LGTM! Nice work

PR #2997, opened 2023-10-23
ignore position check if force = true in reading progress api payload

### LINKED ISSUE #2968: Add an optional force boolean parameter to the reading progress API
When force is set to true we will ignore the position checks (must be greater than previous) and set the value to the supplied value. This will be used in the app to allow users to reset the location and will be used by the PDF reader, which will also set to the most recently read page. Essentially moving some of the business logic to the client side.

sywhb:
Fix #2968


jacksonh:
There will be people carrying you around on their shoulders tonight.

jacksonh (approved):

jacksonh:
Going to merge this one so I can update client side GQL

PR #2998, opened 2023-10-23
Dockerfile optimization

PeterDaveHello:
Apply some best practices and also make the Docker image a little bit smaller.


jacksonh:
Thanks, there are a few PRs changing dockerfiles open right now. I am going to try to get them all merged today, but also they might cause conflicts and we have one high priority item we need to get live first.

jacksonh:
thanks!

PR #3013, opened 2023-10-25
Use a VStack instead of List in the splitbar view to avoid keyboard interaction issues

jacksonh:
I think the list is trying to avoid the keyboard by scrolling,
which can cause teh sidebar toggle button to be scrolled out of
view.

Some information here:

https://samwize.com/2015/12/02/the-buggy-uisplitviewcontroller/


jacksonh:
Not sure if this actually solves the issue, will close it out in favor of using SplitViewController in the next version

PR #3014, opened 2023-10-25
feat: create an api for the cronjob trigger which cleans up the soft deleted users

sywhb:
- add index for status and updated_at columns on user table
- add a pubsub service endpoint for cleanup soft deleted users
- add test cases
- create omnivore_admin role and allow omnivore_admin to delete all users

### DIFF
--- a/packages/api/src/routers/svc/user.ts
+++ b/packages/api/src/routers/svc/user.ts
@@ -34,41 +34,37 @@ const getCleanupMessage = (msgStr: string): CleanupMessage => {
 export function userServiceRouter() {
   const router = express.Router()
 
-  router.post(
-    '/cleanup',
-    cors<express.Request>(corsConfig),
-    async (req, res) => {
-      logger.info('cleanup soft deleted users')
+  router.post('/prune', cors<express.Request>(corsConfig), async (req, res) => {
+    logger.info('prune soft deleted users')
 
-      const { message: msgStr, expired } = readPushSubscription(req)
+    const { message: msgStr, expired } = readPushSubscription(req)
 
-      if (!msgStr) {
-        return res.status(200).send('Bad Request')
-      }
-
-      if (expired) {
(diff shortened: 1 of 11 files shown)


packages/api/src/routers/svc/user.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+export function userServiceRouter() {
+  const router = express.Router()
+
+  router.post('/prune', cors<express.Request>(corsConfig), async (req, res) => {
+    logger.info('prune soft deleted users')
+
+    const { message: msgStr, expired } = readPushSubscription(req)
```
jacksonh, line 40: This function also validates the token, right?

sywhb: Yeah

packages/api/src/routers/svc/user.ts
```diff
@@ -0,0 +1,70 @@
...
+  try {
+    const obj = JSON.parse(msgStr) as unknown
+    if (isCleanupMessage(obj)) {
+      return obj
+    }
+  } catch (err) {
+    console.log('error deserializing event: ', { msgStr, err })
+  }
+
+  return {
```
jacksonh, line 29: Because of this could we sometimes process invalid messages?

sywhb: Actually, the message here is just to sub the days from the deletedAt timestamp so we can prune users within a period

jacksonh (approved):
My only fear with this one is we could kind of lock the database up when it happens, imagine if there were many users with large libraries.

sywhb:
> My only fear with this one is we could kind of lock the database up when it happens, imagine if there were many users with large libraries.

Yeah, I have this concern too

sywhb:
I probably should:

1. Rewrite the sql to do batch deletion
2. run the cronjob at off-peak hours every day

PR #3024, opened 2023-10-26
improvement on integrations

sywhb:
- change source to pocket if item is imported from pocket
- create an integration handler for syncing with readwise
- create a cronjob to export data through integrations


jacksonh:
Hey @sywhb one thing that's come up is people importing their archive items. Would it be hard to add an option "Include my archived items" checkbox, or maybe another way of filtering what's imported? I think ideally we would not import archived by default, but would give users the option to import them too. 

I think maybe Pocket has multiple ways of archive items also, some people have mentioned an "archived folder"

sywhb:
Hey @jacksonh, I didn't find the other way to archive an item in Pocket like the mentioned "archived folder". Does it like a custom list in Pocket?

Except for that, I understand the need to have a filter on items to be imported. Pocket "retrieve" API has a `state` filter too but I think it is probably not enough for us so we will still need to filter them once we retrieved all items. 
```
state
unread = only return unread items (default)
archive = only return archived items
all = return both unread and archived items
```

The retrieved item has only two status: success or archived so we probably can only filter by this as well and skip archived by default

jacksonh (approved):

PR #3027, opened 2023-10-26
fix: site scoped search not working for domain and hostname

sywhb:
* alter site_name_tsv to include hostname and domain
* add test cases


jacksonh:
Thanks. Question on this one: if a site has siteName set, will querying for hostname still work?

sywhb:
> Thanks. Question on this one: if a site has siteName set, will querying for hostname still work?

Yeah, the `site_name_tsv` contains both site name and hostname and domain

jacksonh (approved):

jacksonh:
> > Thanks. Question on this one: if a site has siteName set, will querying for hostname still work?
> 
> Yeah, the `site_name_tsv` contains both site name and hostname and domain

nice. I do think sometimes the vector generation doesnt work for domain names like "omnivore.app" can get split into "omnivore" "app" in the vector and then searching for "omnivore.app" might not work. But lets do more testing once live.

PR #3035, opened 2023-10-27
fix: wrong node version in docker-file in rss-handler

sywhb:


jacksonh (approved):

jacksonh:
Actually looks like its always been the older version. Good to get them all updated though.

PR #3040, opened 2023-10-27
Improve developer environment onboarding

lily-mara:
I encountered a number of issues while trying to get my own local dev environment set up. This change adds missing information to the readme, updates the Postgres image to one that includes required module dependencies, and removes next.js rewrite rules (IS THIS SAFE?) which were no longer used and causing issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![GitHub Workflow Status](https://img.shields.io/github/workflow/status/omnivore-app/omnivore/Run%20tests)](https://github.com/omnivore-app/omnivore/actions/workflows/run-tests.yaml)
 [![Discord](https://img.shields.io/discord/844965259462311966?label=Join%20our%20Discord)](https://discord.gg/h2z5rppzz9)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -100,10 +99,22 @@ with docker compose and the frontend locally:
...
+cp .env.template .env.local
 yarn dev
 ```
 
+You will need to configure some values in the new `.env.local` file. These are
+the values for running the `web` service directly on your host machine and
+running `api` and `content-fetch` within docker:
+
+```sh
+NEXT_PUBLIC_BASE_URL=http://localhost:3000
```
lily-mara, line 111: these values work for me, I'm not sure if they're actually _correct_ for more complete testing.

packages/web/next.config.js
```diff
@@ -24,28 +24,6 @@ const moduleExports = {
       'proxy.omnivore-image-cache.app',
     ],
   },
-  rewrites: () => [
```
lily-mara, line 27: unsure if removing this is safe for production

jacksonh:
This is great, thank you!

The rewrite rules are these for some legacy clients. I have a feeling they wont be needed anymore, but its also quite strange that they cause problems, I am going to try and figure out why they don't for me. You'd think they wouldn't be checking these URLs unless there is a request to them

PR #3049, opened 2023-10-29
Update Obsidian logo

juanbretti:
Based on the new Obsidian logo published at https://obsidian.md/blog/new-obsidian-icon/
I am updating the file using the same filename.


jacksonh:
Bring on the arguments about which one is better.

jacksonh:
Should be in prod in a couple hours

PR #3051, opened 2023-10-30
Fixes #3050 - URLs sensitive to trailing slash

### LINKED ISSUE #3050: URLs sensitive to trailing slash
When saving URLs with a trailing slash, the server removes the trailing slash. But, I've found some URLs which are sensitive to the trailing slash and so most of the times, the content is failed to parse and even if parsed, the content is not the actual content that I wanted to save.

The following is an example URL:
https://www.india.com/webstories/health/yoga-for-digestion-10-asanas-to-try-after-meal-6298225/
Actual URL with trailing slash

https://www.india.com/webstories/health/yoga-for-digestion-10-asanas-to-try-after-meal-6298225
Trailing slash removed - leads to a default page with no content

Expected behavior:
The server should not remove the trailing slash.

naveennamani:
By passing false to `removeTrailingSlash` for normalize-url call, the URLs sensitive to trailing slash will be parsed correctly

### DIFF
--- a/packages/api/src/utils/helpers.ts
+++ b/packages/api/src/utils/helpers.ts
@@ -385,5 +385,6 @@ export const cleanUrl = (url: string) => {
     stripHash: true,
     stripWWW: false,
     removeQueryParameters: trackingParams,
+    removeTrailingSlash: false,
   })
 }


jacksonh:
Thanks, this seems sensible. @sywhb do you think it would cause any issues?

sywhb:
This is useful. My only concern is it might affect deduplication when saving by URL

naveennamani:
@sywhb I agree with you. Let us look at the possibilities where there may be issues.

| URL in database | URL requested to save | Existing scheme | Proposed scheme |
| -- | -- | -- | -- |
| without slash | without slash | no duplication | no duplication |
| without slash | with slash | no duplication | treated as different |
| with slash | without slash | no URLs with slash in existing scheme | treated as different |
| with slash | with slash | no URLs with slash in existing scheme | no duplication |

In the problematic two scenarios, there are two cases, one where the URL is sensitive to the slash and one where it is not. If the URL is sensitive, there will be two different contents and only one of them might be correct and user will delete another. Also, when sharing or copying this URL, the trailing slash will also be present (if required), so the possibility of this scenario is minimal and duplication is mostly removed.

In the remaining case, there shall be duplicate articles, but the situation cannot be avoided with the assumption that the URL is sensitive to the trailing slash. Only after parsing the content we will know if the two URLs are same or not.

jacksonh:
One thing I wonder is if our checksums could help too. I think duplication going forward isn't too much of an issue, i am trying to think if there are any issues with historical data, but I don't think so.

naveennamani:
> One thing I wonder is if our checksums could help too. I think duplication going forward isn't too much of an issue, i am trying to think if there are any issues with historical data, but I don't think so.

It all boils down to the question correctness vs duplication. In my opinion, it's better to have two correct articles then one wrong article which user have no control on it.

sywhb (approved):

PR #3054, opened 2023-10-31
feature: following

sywhb:
- create feed, feed_item and user_feed_item tables in db
- add feeds graphql api to the schema
- add feed entity
- add feeds graphql api for searching feed by title or url
- add following graphql api to fetch user following


packages/api/src/entity/library_item.ts
```diff
@@ -196,4 +196,28 @@ export class LibraryItem {
...
   recommenderNames?: string[] | null
+
+  @Column('timestamptz')
+  hiddenAt?: Date | null
+
+  @Column('timestamptz')
+  sharedAt?: Date | null
+
+  @Column('text')
+  sharedBy?: string | null
```
jacksonh, line 207: We could maybe leave out sharedAt and sharedBy right now.

packages/api/src/entity/library_item.ts
```diff
@@ -196,4 +196,28 @@ export class LibraryItem {
...
+
+  @Column('jsonb')
+  links?: any | null
+
+  @Column('text')
+  previewContent?: string | null
```
jacksonh, line 213: I wonder if we need a `previewContentType` or is it always HTML from feeds?

packages/api/src/entity/library_item.ts
```diff
@@ -196,4 +196,28 @@ export class LibraryItem {
...
+
+  @Column('timestamptz')
+  seenAt?: Date | null
+
+  @Column('boolean')
```
jacksonh, line 218: I think before we were going to do `addedToLibraryAt?: Date` ?

packages/api/src/entity/library_item.ts
```diff
@@ -196,4 +196,28 @@ export class LibraryItem {
...
+  @Column('text')
+  sharedBy?: string | null
+
+  @Column('jsonb')
+  links?: any | null
+
+  @Column('text')
+  previewContent?: string | null
+
+  @Column('timestamptz')
```
jacksonh, line 215: nice

packages/api/src/entity/subscription.ts
```diff
@@ -70,4 +70,10 @@ export class Subscription {
 
   @Column('timestamp', { nullable: true })
   scheduledAt?: Date | null
+
+  @Column('boolean')
+  isPublic?: boolean | null
+
+  @Column('boolean')
+  isFetchingContent?: boolean | null
```
jacksonh, line 78: I wonder if we should use a state enum instead of a bool

packages/api/src/generated/schema.graphql
```diff
@@ -1,5 +1,21 @@
 directive @sanitize(allowedTags: [String], maxLength: Int, minLength: Int, pattern: String) on INPUT_FIELD_DEFINITION
 
+type AddFollowingToLibraryError {
```
jacksonh, line 3: I wonder if this should be named something like `move` like `MoveFromFollowingToLibrary` -- but also not sure if it should move the item or copy the item.

packages/api/src/generated/schema.graphql
```diff
@@ -2071,6 +2129,8 @@ enum SubscribeErrorCode {
 }
 
 input SubscribeInput {
+  isFetchingContent: Boolean
```
jacksonh, line 2132: Is this a state, or is it like "automatically add to library"?

packages/api/src/repository/feed.ts
```diff
@@ -0,0 +1,33 @@
+import { ILike } from 'typeorm'
+import { appDataSource } from '../data_source'
+import { Feed } from '../entity/feed'
+
+export const feedRepository = appDataSource.getRepository(Feed).extend({
```
jacksonh, line 5: For this do we need to use user id to fetch their private feeds, or add a filter for public/private?

jacksonh (commented):
One part I'm not sure about is the relationship between feed and subscription and how we handle private feeds. If a user adds a private feed, does it get added to the feeds table?

packages/api/src/entity/library_item.ts
sywhb: Why do we want to leave out them?

I need `sharedAt` to know if this is in following and `sharedBy` is to store the feed url or followee's username.

packages/api/src/entity/library_item.ts
sywhb: I think we need `previewContentType` because I'm not sure if it is always HTML

packages/api/src/entity/library_item.ts
jacksonh: @sywhb oh i see. I was thinking it was for posting recommendations, not data needed by the feeds.

packages/api/src/entity/subscription.ts
sywhb: Ahh the name is misleading. It's actually a flag to indicate if we want to fetch content

packages/api/src/entity/library_item.ts
sywhb: Yeah, I will add it

packages/api/src/generated/schema.graphql
sywhb: From user perspective, it probably makes more sense to be named `copy` than `move`

packages/api/src/generated/schema.graphql
sywhb: I know it is confusing but I meant "automatically add to library"

packages/api/src/repository/feed.ts
sywhb: This table is actually used for searching a feed and we have `is_public` flag in the `subscription` table

sywhb:
> One part I'm not sure about is the relationship between feed and subscription and how we handle private feeds. If a user adds a private feed, does it get added to the feeds table?

I think we would not add it to the `feed` table but marks `is_public = false` in `subscription` table

packages/api/src/entity/library_item.ts
sywhb: I didn't come up with good variable names here

sywhb: @jacksonh maybe I should change `sharedAt` to `addedToFollowingAt` and `sharedBy` to `addedToFollowingBy`?

packages/api/src/entity/subscription.ts
sywhb: @jacksonh maybe change this one to `shouldFetchContent`?

packages/api/src/repository/feed.ts
jacksonh: I see, so when a user adds an RSS subscription is a feed row created?

jacksonh:
> > One part I'm not sure about is the relationship between feed and subscription and how we handle private feeds. If a user adds a private feed, does it get added to the feeds table?
> 
> I think we would not add it to the `feed` table but marks `is_public = false` in `subscription` table

Got it. I think that answers my previous question

packages/api/src/entity/library_item.ts
jacksonh: Yeah I think that would make it a little easier to understand

jacksonh:
Looks great, just trying to understand all the moving parts.

packages/api/src/repository/feed.ts
sywhb: Yeah and we need to migrate the existing data too

sywhb:
Hey @jacksonh I have changed the names accordingly and could you help to take a look again?

packages/api/src/generated/schema.graphql
```diff
@@ -1294,6 +1351,7 @@ type Query {
   article(format: String, slug: String!, username: String!): ArticleResult!
   articleSavingRequest(id: ID, url: String): ArticleSavingRequestResult!
   deviceTokens: DeviceTokensResult!
+  feeds(input: FeedsInput!): FeedsResult!
```
jacksonh, line 1354: Sorry, i realized maybe this should be named `following`

jacksonh (commented):
Sorry, i realize this is an annoying naming change, but I think we need to rename `feeds` and `FeedsResult` to `following` `FollowingResult`

packages/api/src/generated/schema.graphql
sywhb: @jacksonh, I think maybe I didn't make it very obvious but this API is to search an RSS feed by name or URL but to search following which we would use the same search API

sywhb: Some users want to search for an RSS feed by name so they won't need to memorise or fill in the complete URL when they subscribe to one

jacksonh (approved):

Issue #3044, opened 2023-10-28
Data sent to webhook url using `HIGHLIGHT_CREATED` contains vulnerable information


muhimu:
Highlight created data contains user data (email, encoded password). Those data are irrelevant to highlight and should not be sent to webhook url.
```
{
  "action": "created",
  "userId": "<REMOVED>",
  "highlight": {
    "type": "highlight",
    "userId": "<REMOVED>",
    "id": "ee89383f-bace-4a70-afc8-7ac2b5552876",
    "shortId": "-GqT60Ib",
    "quote": "Omnivore's advanced search syntax supports searching for multiple labels using AND and OR clauses. You can also negate a label search to find pages that do not have a certain label.",
    "prefix": " ",
    "suffix": " \n        ",
    "patch": "@@ -1739,32 +1739,52 @@\n abels %0A         \n+%3Comnivore_highlight%3E\n Omnivore's advan\n@@ -1948,16 +1948,37 @@\n n label.\n+%3C/omnivore_highlight%3E\n  %0A      \n",
    "annotation": null,
    "createdAt": "2023-10-28T10:35:07.331Z",
    "updatedAt": "2023-10-28T10:35:07.331Z",
    "sharedAt": null,
    "highlightPositionPercent": 89.107796,
    "highlightPositionAnchorIndex": 38,
    "highlightType": "HIGHLIGHT",
    "html": "Omnivore's advanced search syntax supports searching for multiple labels using AND and OR clauses. You can also negate a label search to find pages that do not have a certain label.",
    "color": "red",
    "user": {
      "id": "<REMOVED>",
      "name": "<REMOVED>",
      "source": "EMAIL",
      "email": "<REMOVED>",
      "sourceUserId": "<REMOVED>",
      "createdAt": "<REMOVED>",
      "updatedAt": "<REMOVED>",
      "password": "<REMOVED>",
      "status": "ACTIVE",
      "profile": {
        "id": "<REMOVED>",
        "username": "muhimu",
        "bio": null,
        "pictureUrl": null,
        "createdAt": "<REMOVED>",
        "updatedAt": null,
        "private": false
      }
    },
    "labels": [],
    "pageId": "2509fec2-6b1e-11ee-aef2-3f8a7ac28e92"
  }
}
```

jacksonh:
Thanks for the report, sorry just saw this one now.

PR #3057, opened 2023-11-01
fix/irrelevant information in webhook event

sywhb:
- do not publish events if the highlights are recommended by club users
- remove user from highlight webhook event
- remove unused fields in library_item entity so we exclude them from the webhook event


sywhb:
Fix #3044

jacksonh (approved):

PR #3093, opened 2023-11-08
Update voices

jacksonh:
- Add some voices


packages/text-to-speech/src/textToSpeech.ts
```diff
@@ -7,6 +7,7 @@ export interface TextToSpeechInput {
   rate?: string
   secondaryVoice?: string
   audioStream?: NodeJS.ReadWriteStream
+  isOpenAIVoice?: boolean
```
sywhb, line 10: Why do we need this flag? Is it used in the client?

sywhb (approved):
LGTM!

packages/text-to-speech/src/textToSpeech.ts
jacksonh: Yeah, just passed from the client to be more explicit. I think maybe in the future we could pass something more explicit like "provider" or just have a big list of names and figure it out based on name.

Issue #2640, opened 2023-08-10
Bump node-gyp for local development


pythoninthegrass:
Hi there.

When building the `docker-compose.yml` locally, it fails on the `api_builder` layer due to a deprecation [upstream in Python](https://github.com/nodejs/gyp-next/commit/d0504e6700ce48f44957a4d5891b142a60be946f) 3.11 raising:
```bash
ValueError: invalid mode: 'rU' while trying to load binding.gyp
```
Node's not my forte, so it's unclear where to bump the `"node-gyp": "^9.4.0",` dependency — although it seems the top-level `yarn.lock` via `package.json` would be the most likely spot (cf. https://github.com/omnivore-app/omnivore/blob/main/yarn.lock#L21123.)

Happy to test suggestions and open a PR after getting a working solution running.

**OS**: macOS 12.6.7 (21G651)
**Docker**: version 24.0.5, build ced0996

[node-gyp_traceback_api_builder.log](https://github.com/omnivore-app/omnivore/files/12316729/node-gyp_traceback_api_builder.log)

pythoninthegrass:
```bash
λ docker compose up
[+] Building 48.7s (17/35)
 => [api internal] load build definition from Dockerfile                                                           0.0s
 => => transferring dockerfile: 1.89kB                                                                             0.0s
 => [api internal] load .dockerignore                                                                              0.0s
 => => transferring context: 156B                                                                                  0.0s
 => [api internal] load metadata for docker.io/library/node:18.16-alpine                                           0.0s
 => [api internal] load build context                                                                              0.1s
 => => transferring context: 80.02kB                                                                               0.0s
 => [api builder  1/23] FROM docker.io/library/node:18.16-alpine                                                   0.0s
 => CACHED [api builder  2/23] WORKDIR /app                                                                        0.0s
 => CACHED [api builder  3/23] RUN apk add g++ make python3                                                        0.0s
 => CACHED [api builder  4/23] COPY package.json .                                                                 0.0s
 => CACHED [api builder  5/23] COPY yarn.lock .                                                                    0.0s
 => CACHED [api builder  6/23] COPY tsconfig.json .                                                                0.0s
 => CACHED [api builder  7/23] COPY .prettierrc .                                                                  0.0s
 => CACHED [api builder  8/23] COPY .eslintrc .                                                                    0.0s
 => CACHED [api builder  9/23] COPY /packages/readabilityjs/package.json ./packages/readabilityjs/package.json     0.0s
[...]

PR #3112, opened 2023-11-11
Update node and run-script

pythoninthegrass:
- Bump node docker image: 14.18 -> 18.16.1
- Bump run-script as workaround for `node-gyp` upstream deprecation (`ValueError: invalid mode: 'rU' while trying to load binding.gyp`) (cf. #2640)


jacksonh:
Hey it looks like this one needs to be rebased. Let me know if you don't have time and I'll pull in the other parts.

pythoninthegrass:
Hi @jacksonh,

Sorry about that! Looks like in the process of rebasing, I accidentally closed the PR. Believe it's fixed now, but please let me know if that's not the case or you need anything else.

self-hosting/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:18.16-alpine as builder
+FROM node:18.16.1-alpine as builder
```
jacksonh, line 1: Do you know if we need this?

pythoninthegrass: It's [best practice](https://hackernoon.com/docker-images-name-vs-tag-vs-digest) if you want images to work on machines other than yours. By pinning the minor version of the node image, it can still get security updates (vs. pinning manifest), but is guaranteed to work with the version of node Omnivore was tested with.

So for example, I can be fairly confident that we'll always get this
```bash
/app # node --version
v18.16.1
```
whereas if you only use the `node:18.16-alpine` tag, you can end up with `v18.6.2` or later (although it appears to [get updated every 5-6 months](https://hub.docker.com/layers/library/node/18.16-alpine/images/sha256-055cfc51875a99402319113cb69404f8cd5086e8f5b128bee71126c04990a97a?context=explore).)

It's your show, so understand if only specifying major/minor version is preferred.

jacksonh:
This one isn't needed anymore. Thanks!

Issue #2273, opened 2023-05-29
In dark mode on Android the scrollbar is not visible in the reader


thiswillbeyourgithub:
Maybe related to #184 

When reading, if I select the "black" reading mode the scrollbar is invisible (maybe just too faint?).

In the other less black theme it's still visible.

I'm on the latest release from may 24th (don't have the version number on me)

jacksonh:
hey, is this iOS?

thiswillbeyourgithub:
Sorry no, android

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #3116, opened 2023-11-13
Android: Fix reader scrollbar visibility

remychantenay:
Fixing #2273

## Important
Please note that with the current code, when changing the theme directly in the reader, the UI is not refreshed. This PR doesn't fix this.


jacksonh:
wow i could never figure this one out!

jacksonh (approved):

PR #3139, opened 2023-11-17
change archive and trash from folder to state

sywhb:
- add folder to the updates api params
- allow updating taskName in updateIntegration API
- update taskName to null when no data to import
- change archive and trash from folder to state
- add RSS label to the following feed items


jacksonh (approved):

packages/api/src/services/labels.ts
```diff
@@ -92,6 +93,10 @@ export const saveLabelsInLibraryItem = async (
     userId
   )
 
+  if (skipPubSub) {
```
jacksonh, line 96: smart

PR #3159, opened 2023-11-22
feat: wildcard search for labels and nested labels

sywhb:


jacksonh (approved):
Could probably use a test case on this one

PR #3175, opened 2023-11-27
detect language from html content

sywhb:
Tested with different libraries and this seems pretty fast and accurate: https://github.com/dachev/node-cld
It also supports HTML content

### DIFF
--- a/packages/readabilityjs/Readability.js
+++ b/packages/readabilityjs/Readability.js
@@ -3030,22 +3030,19 @@ Readability.prototype = {
     return false;
   },
 
-  _getLanguage: async function(locale, content, languageCode) {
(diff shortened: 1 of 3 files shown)


jacksonh:
I think we'd be best off if we stored the code like `en` instead of `English`. I know in the past we stored the full name, but I bet more tools would work with the codes.

sywhb:
@jacksonh Should we store ISO 639-1 or 639-3 language code?

We probably should also store a mapping of language code to ISO language name in the backend for searching.
A simple dropdown list of language selector could be added to the UI as well

sywhb:
I think probably we should use ISO 639-3 for differing languages like Simplified and Traditional Chinese

jacksonh (approved):

PR #3184, opened 2023-11-28
fix feed url in pdf file not saved

### LINKED ISSUE #3180: RSS links to PDFs do not properly set subscription information
If a link in RSS points to a PDF item, i believe puppeteer-parse detects that this is a PDF and then creates a separate download, so this is why items don't get their subscription information set.

sywhb:
Fix #3180

### DIFF
--- a/packages/api/src/services/labels.ts
+++ b/packages/api/src/services/labels.ts
@@ -3,7 +3,7 @@ import { QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity
 import { EntityLabel } from '../entity/entity_label'
 import { Label } from '../entity/label'
 import { LibraryItem } from '../entity/library_item'
-import { createPubSubClient, EntityType } from '../pubsub'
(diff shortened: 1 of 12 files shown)


sywhb:
> LGTM, one thing i can't tell from the PR though, will the `RSS` label also be added if feedUrl is non-null?

Hah, no but yeah we should. Let me add that part

packages/api/src/services/labels.ts
```diff
@@ -65,6 +65,33 @@ export const findOrCreateLabels = async (
...
+  userId: string,
+  labels?: CreateLabelInput[] | null,
+  rssFeedUrl?: string | null,
+  pubsub?: PubsubClient,
+  skipPubSub?: boolean
+) => {
+  if (rssFeedUrl) {
+    // add rss label to labels
+    labels = (labels || []).concat({ name: 'RSS' })
+  }
```
sywhb, line 79: @jacksonh This will add RSS label to the library item if feedUrl is not null

jacksonh (approved):

PR #3188, opened 2023-11-30
Abstract out fetching from view model so we can better handle multiple fetch folders

jacksonh:
Rename LinkedItem to LibraryItem

More on following

Add new fetcher

Tab bar


sywhb:
This is a HUGE one 🔥

PR #3198, opened 2023-12-04
fix moveToFolder API error by skipping fetching content again

sywhb:


sywhb:
> I wonder if we even need the articleSavingRequest as a response now.

Good question. I think it is not needed for now but would it be troublesome for you to update the client if I replaced it with a `success: boolean` as a response? @jacksonh

jacksonh:
Naw, would probably be simpler on the client. iOS would just be a small change so only the beta would break for you and I.

packages/api/src/schema.ts
```diff
@@ -2665,7 +2665,7 @@ const schema = gql`
   union MoveToFolderResult = MoveToFolderSuccess | MoveToFolderError
 
   type MoveToFolderSuccess {
-    articleSavingRequest: ArticleSavingRequest!
```
sywhb, line 2668: @jacksonh I have replaced the `articleSavingRequest` with `success` here as a response

PR #3214, opened 2023-12-06
fix some bugs with new search feature

sywhb:
- fix searching for multiple labels in iOS
- fix bulk action query limited to 1000 char
- add test cases

### DIFF
--- a/packages/api/src/resolvers/article/index.ts
+++ b/packages/api/src/resolvers/article/index.ts
@@ -822,8 +822,12 @@ export const bulkActionResolver = authorized<
         },
       })
 
-      // the query size is limited to 255 characters
-      if (!query || query.length > 255) {
+      // the query size is limited to 1000 characters
+      if (!query || query.length > 1000) {
+        log.error('bulkActionResolver error', {
+          error: 'QueryTooLong',
+          query,
+        })
         return { errorCodes: [BulkActionErrorCode.BadRequest] }
       }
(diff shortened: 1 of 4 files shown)


packages/api/src/resolvers/article/index.ts
```diff
@@ -822,8 +822,12 @@ export const bulkActionResolver = authorized<
         },
       })
 
-      // the query size is limited to 255 characters
-      if (!query || query.length > 255) {
+      // the query size is limited to 1000 characters
```
jacksonh, line 825: hah! I was wondering if it was something like that. I think we might need it even higher if we are allowing 100 uuids.

packages/api/src/resolvers/article/index.ts
```diff
@@ -822,8 +822,12 @@ export const bulkActionResolver = authorized<
         },
       })
 
-      // the query size is limited to 255 characters
-      if (!query || query.length > 255) {
+      // the query size is limited to 4000 characters to allow for 100 items
+      if (!query || query.length > 4000) {
```
sywhb, line 826: @jacksonh Changed to max 4000 characters here

jacksonh: 🙏

sywhb:
Okay. I will do a final quick test in demo and deploy backend in prod @jacksonh

PR #3223, opened 2023-12-08
wrap the where query in parenthesis because we also need to check user_id

sywhb:


jacksonh (approved):
I'd say maybe we should have a test

jacksonh (approved):

PR #3235, opened 2023-12-11
increase feed limit to 256

sywhb:
- get max feed limit from env
- max rss feed defaults to 256

### DIFF
--- a/packages/api/src/util.ts
+++ b/packages/api/src/util.ts
@@ -104,6 +104,11 @@ interface BackendEnv {
   pocket: {
     consumerKey: string
   }
+  subscription: {
+    feed: {
+      max: number
+    }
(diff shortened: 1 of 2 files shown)


jacksonh (approved):

packages/api/src/util.ts
```diff
@@ -303,6 +309,14 @@ export function getEnv(): BackendEnv {
     consumerKey: parse('POCKET_CONSUMER_KEY'),
   }
 
+  const subscription = {
```
jacksonh, line 312: thanks, we should probably do this default thing for more items.

sywhb: yeah, starting to add the default value to the env var

PR #3236, opened 2023-12-11
update scan feed api for the client

sywhb:
- validate and return rss feed if the url is an rss feed url
- return absolute feed url

### DIFF
--- a/packages/api/src/resolvers/subscriptions/index.ts
+++ b/packages/api/src/resolvers/subscriptions/index.ts
@@ -17,7 +17,6 @@ import {
   ScanFeedsError,
   ScanFeedsErrorCode,
   ScanFeedsSuccess,
-  ScanFeedsType,
   SortBy,
   SortOrder,
   SubscribeError,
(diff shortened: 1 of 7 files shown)


packages/api/src/resolvers/subscriptions/index.ts
```diff
@@ -434,37 +427,60 @@ export const scanFeedsResolver = authorized<
...
-        title: link.getAttribute('title') || '',
-        type: 'rss',
-      }))
-      .filter((feed) => feed.url)
+    // fetch page content and parse feeds
+    const response = await axios.get(url, RSS_PARSER_CONFIG)
+    const content = response.data as string
+    // check if the content is html or xml
+    const contentType = response.headers['content-type']
+    const isHtml = contentType?.includes('text/html')
```
jacksonh, line 443: I wonder if we will need more types here. Will find out with some testing I guess.

jacksonh (approved):
I think there's one more case we could consider: if the user passes a feed to an OPML. But we can do that separately.

jacksonh:
> > I think there's one more case we could consider: if the user passes a feed to an OPML. But we can do that separately.
> 
> that is a good one. probably we should rename it to XML to allow both RSS and OPML?

oh i think it would still make sense because we are passing a URL still.

PR #3247, opened 2023-12-14
feat: fetch content and folder settings for newsletters and subscriptions

sywhb:
- add fetch_content and folder column to the subscriptions table and folder column to newsletter_emails table
- allow setting fetchContent and folder when subscribe to a feed
- fetch following item content only when required
- allow updating folder and fetch content settings for subscriptions
- allow setting folder when creating a new email
- save newsletters to the right folder get from subscription and email settings
- move to folder api will create a task to fetch content if content not exists

### DIFF
--- a/packages/api/src/entity/newsletter_email.ts
+++ b/packages/api/src/entity/newsletter_email.ts
@@ -34,4 +34,13 @@ export class NewsletterEmail {
 
   @OneToMany(() => Subscription, (subscription) => subscription.newsletterEmail)
   subscriptions!: Subscription[]
+
+  @Column('text')
+  folder!: string
+
+  @Column('text')
+  name?: string | null
+
+  @Column('text')
+  description?: string | null
 }
--- a/packages/api/src/resolvers/function_resolvers.ts
+++ b/packages/api/src/resolvers/function_resolvers.ts
@@ -319,6 +319,19 @@ export const functionResolvers = {
       if (article.wordCount) return article.wordCount
       return article.content ? wordsCount(article.content) : undefined
     },
+    async labels(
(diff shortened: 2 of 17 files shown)


packages/api/src/entity/newsletter_email.ts
```diff
@@ -34,4 +34,13 @@ export class NewsletterEmail {
 
   @OneToMany(() => Subscription, (subscription) => subscription.newsletterEmail)
   subscriptions!: Subscription[]
+
+  @Column('text')
```
jacksonh, line 38: I'd maybe argue this could be null and we'd fallback in code, but probably either way is the same.

packages/api/src/resolvers/function_resolvers.ts
```diff
@@ -319,6 +319,19 @@ export const functionResolvers = {
       if (article.wordCount) return article.wordCount
       return article.content ? wordsCount(article.content) : undefined
     },
+    async labels(
```
jacksonh, line 322: How come this change is needed?

jacksonh (approved):

packages/api/src/entity/newsletter_email.ts
sywhb: i guess we need it because the existing newsletters have to be saved in `inbox` and newer ones in `following`

packages/api/src/resolvers/function_resolvers.ts
sywhb: hah, the labels we got from `inner join` do not have `source` in it which i think is something with typeorm 😭

PR #3261, opened 2023-12-18
feature/add state for non fetch content

sywhb:
- add content_not_fetched state for items saved without content fetched
- if fetch_content is false, set the state to content_not_fetched and use preview content
- run readability on the preview content


jacksonh:
Looks good to me, just re-running the tests.

sywhb:
> Looks good to me, just re-running the tests.

it's weird that the tests keeps failing

jacksonh (approved):

PR #3264, opened 2023-12-19
add endpoint to prune items from folder older than some days

sywhb:


packages/api/src/routers/svc/links.ts
```diff
@@ -52,5 +80,38 @@ export function linkServiceRouter() {
     }
   })
 
+  router.post('/prune', async (req, res) => {
+    logger.info('prune expired items in folder')
+
+    const { message: msgStr, expired } = readPushSubscription(req)
```
jacksonh, line 86: This one will validate the token, right?

sywhb: Yeah, validate the token from the pubsub payload

jacksonh (approved):

PR #3266, opened 2023-12-19
android: add swipe actions in library

remychantenay:
# Summary
Adding swipe actions (archive and delete) to Library items.

* `Left-to-right`: archive (or unarchive, depending on the state)
* `Right-to-left`: delete

Please note that there is no "undo" possibility with the current implementation. We can iterate later.

# Screen Record (Loom)
The screen record below shows 1) archiving two items, 2) unarchiving two items and 3) deleting two items. In that order.

https://www.loom.com/share/89f0efeea90345e19d34cddb504dc005?sid=3cc71a55-7283-413a-b58c-1e309ee9e983

### DIFF
--- a/android/Omnivore/app/src/main/java/app/omnivore/omnivore/ui/library/LibraryView.kt
+++ b/android/Omnivore/app/src/main/java/app/omnivore/omnivore/ui/library/LibraryView.kt
@@ -2,6 +2,10 @@ package app.omnivore.omnivore.ui.library
 
 import android.content.Intent
 import android.util.Log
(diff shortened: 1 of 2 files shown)


jacksonh (approved):

jacksonh:
I approved all these without merging in case you think they need to go in a specific order

remychantenay:
> I approved all these without merging in case you think they need to go in a specific order

Sounds good @jacksonh, thanks for reviewing 🙏 

The order doesn't really matter, also I can't merge due to Vercel's failed deployments

PR #3280, opened 2023-12-26
execute update query in batch to avoid slow bulk action query

sywhb:


jacksonh:
0_0 

Was going to ask about this as some of the clean up features will rely heavily on bulk.

sywhb:
@jacksonh i think this still might not be the best way to achieve incremental update but that's so far the best way i can do:

in the db transaction:
* Get the total count of rows to be updated
* Calculate the number of batches
* Open a cursor
* Loop through batches
* Fetch the next batch of IDs
* Perform incremental update for the current ID
* Close the cursor

let me know if you got some better idea or alternatives 👍

sywhb:
similar idea could be applied to prune user accounts -- incremental delete of library items and then delete user account

jacksonh (approved):

PR #3165, opened 2023-11-23
[Feature] Omnivore Discover: A way to discover new content. 

Podginator:
Just moving over the PR my fork. 

# omnivore-discover

## What is this?  

One of my biggest problems is actually discoverability of articles. I have my five sites, and my link aggregators like Reddit. This is a bubble, and I miss a lot this way.

So I wanted to see if I could create something that would enable discoverability from Omnivore.

I had a few goals when creating Omnivore Discover.

Example Screen

### Automatic Categorisation

A while ago I worked a proof of concept for automatically adding user tags to an article. I ultimately still need to work on that further, but the basics for it worked well.

I wanted to take the learnings from this and use it to add automatic categorisation of stories.

I created a few topics, and added some descriptions to them. I generate an Embedding from this using OpenAIs embedding. These can be seen below.

Topics

When ingesting articles (see Ingesting Articles) we use their title and small description to create an Embedding. We can then use Cosine Similarity to identify which category this story should be a part of.

This is of course not 100% accurate, but it does a good enough job at categorising articles.

#### Discord Integration.

I created Omnivore Discover, and added it to the Omnivore WebApp.

I wanted to also add some social features to this. We have a fantastic community within the Omnivore Discord. I have found a lot of interesting reads in the #recommendations channel.

I wanted to be able to take these recommendations, and expose them to the Omnivore Community.

We do this using a Discord Bot. In order to moderate these recommendations a moderator must add an emoji (🦥) to the story.

This then gets ingested in the same way as the other stories. Meaning that it is also categorised. It also gets added to the Community Picks tab.

Tomnivore Slack

community tab

#### Popularity

There is also a popularity feed. This provides a score based on recent saves, weighting more heavily for newer articles. This allows us to have a popular tab, which shows in order the most popular stories on Omnivore Right now according to the community

Popular Items

### Ingesting Articles

I ensured that articles could come from multiple locations. This is why I chose an RXJS Poller.

This project also started from the automatic labelling project. So that too was an important part of the decision to enable ingestion from multiple plages. Including a PubSub queue.

I wanted one of the main sources of the articles to be RSS Feeds.

I did this because I thought that some of this functionality might, in the future, be extendable to other RSS Feeds.

I have chose 3 article sources for now, Wired, ArsTechnica, and The Atlantic.

## Technologies

Below is a list of the technologies that were used to design this feature. This repository represents the RXJS side.

* RxJS 
* Typescript 
* Axios 
* PGVector
* Discord Bot
* PubSub

## Running

[...]

Issue #2996, opened 2023-10-23
Unable to make two highlights one after another


duxrat:
Demonstrated in gif, it's really important to me to be able to have the highlights separate to color/label/process them differently. Now I have to cut off the last word of a previous highlight to prevent them from automatically merging, overriding the first one's data.

![2023-10-23 at 14 15 03](https://github.com/omnivore-app/omnivore/assets/69042276/ffd09c5b-f1e3-46d7-a2fb-39279e0b7641)

I don't think there should be any circumstance under which two highlights merge automatically, even when they have the same color/tags. I make them separately so they can stay separate.

PR #3282, opened 2023-12-28
Add documentation for highlight and fix bug

aaachen:
This change fixes the merge highlight issue: https://github.com/omnivore-app/omnivore/issues/2996

## Problem

Consecutive highlights will automatically get merged into one. 

The root cause is
1. The highlight end offset is non-inclusive (highlight is only supported over text. The offset is counted by characters for all elements' text content)
1. The overlapping highlight calculation treats the end offset as inclusive (see `useSelection.ts`) 

## Changes

1. Corrected the logic
2. Added documentations
3. Correct a new visual bug resulting from changes of 1). See below section for description

### Visual Bug

Example sequence:

selection-edge-case
**selection cursor ends at empty area, selecting no text at the focus node** 

selection-edge-case-2
**create highlight (note: the word `share` is not selected)** 

selection-edge-case-3
**create another highlight, note the highlight overlap visual bug** 

selection-edge-case-4
**removing one of highlight sometimes result in the patching to go wrong and jumble the source text**#3165 

The fix for this visual bug is to clip the selection to the preceding/succeeding DOM element by updating the selection range.

## Testing

Manually tested

1. Forward select

selection-fix-edge-forward-select-1
selection-fix-edge-forward-select-2
selection-fix-edge-forward-select-3

2. Reverse select

selection-fix-edge-backward-select-1
selection-fix-edge-backward-select-2


jacksonh:
FYI, this one looks good to me. Just fighting a couple fires so I don't want to merge it until early next week.

jacksonh:
OK, going to get this one into our test environment finally. Sorry on the delay. Putting out some fires this week

PR #3294, opened 2024-01-01
Improve keyboard shortcuts modal 

cdev147:
Better Model so that code looks cleaner

### DIFF
--- a/packages/api/src/resolvers/function_resolvers.ts
+++ b/packages/api/src/resolvers/function_resolvers.ts
@@ -27,6 +27,7 @@ import { findLabelsByLibraryItemId } from '../services/labels'
 import { findRecommendationsByLibraryItemId } from '../services/recommendation'
 import { findUploadFileById } from '../services/upload_file'
 import {
+  highlightDataToHighlight,
   isBase64Image,
   recommandationDataToRecommendation,
   validatedDate,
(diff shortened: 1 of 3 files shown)


cdev147:
> Thanks, one issue is it looks like a different prettier config. If you'd like I can try running ours (the main thing i notice is the semicolons.

yes sure, also let me know if I can help for the same as I have just started open source and it would help me grow.

jacksonh:
> > Thanks, one issue is it looks like a different prettier config. If you'd like I can try running ours (the main thing i notice is the semicolons.
> 
> yes sure, also let me know if I can help for the same. @jacksonh should I remove the semi colons?

Have you used prettier before? If you are using Visual Studio code it should format it according to our coding standards for you. But sometimes it takes some manual setup.

jacksonh:
What i am going to do is merge this, then run prettier locally, then create a new PR.

However, I think this is mostly a cosmetic code change.

jacksonh:
Actually, I need to back this one out. It was hard to tell with the prettier changes but this isn't valid:

```
const initalKeyMapping = useMemo(
    () => useInitialKeyMapping(commands),
    [commands]
  )
```

you get this error:

```
64:11  Error: React Hook "useInitialKeyMapping" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function.  react-hooks/rules-of-hooks
```

Issue #3298, opened 2024-01-02
Update default highlights saved search


jacksonh:
For `Highlights` we want to also show items in the archive so it should be `in:all has:highlights mode:highlights`

tooniz:
Thanks, I also found this to be annoying. My archived highlights are no longer showing, no longer syncing with Obsidian etc.

Please either use the suggested filter or allow the user to save a custom one

jacksonh:
> Thanks, I also found this to be annoying. My archived highlights are no longer showing, no longer syncing with Obsidian etc.
> 
> Please either use the suggested filter or allow the user to save a custom one

You can use a custom query in the Obsidian plugin settings. In the interm you can also create a custom saved search.

jacksonh:
This one is done now.

PR #3303, opened 2024-01-03
Update the default Highlights query

jacksonh:
This query is more of what people want, all their created
highlights, not just ones in their current inbox. Since a normal
workflow would be: read + highlight, archive, review highlights.

### DIFF
--- a/packages/api/src/services/create_user.ts
+++ b/packages/api/src/services/create_user.ts
@@ -160,7 +160,7 @@ const createDefaultFiltersForUser =
         filter: 'in:inbox sort:read-desc is:reading',
       },
       { name: 'Non-Feed Items', filter: 'no:subscription' },
-      { name: 'Highlights', filter: 'has:highlights mode:highlights' },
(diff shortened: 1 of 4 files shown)


jacksonh:
This will be a pretty big update for the database so I'm not sure if we should run it outside the deploy

jacksonh:
https://github.com/omnivore-app/omnivore/issues/3298

jacksonh:
@sywhb we can discuss this in discord, I feel like psql should run it fine but maybe i should run some batches outside of the deploy

jacksonh:
There is a race condition here too. When we deploy the migration will be run, then any new users that are created before the the backend is deployed will have the old version. So we need to re-run the update.

sywhb:
> There is a race condition here too. When we deploy the migration will be run, then any new users that are created before the the backend is deployed will have the old version. So we need to re-run the update.

nice catch. probably we should move the default filters logic to the database like creating a function/trigger @jacksonh

jacksonh:
Yeah that would be good.

Issue #3059, opened 2023-11-01
Feature request: Subscribe to RSS feed via extension


lily-mara:
When the browser is navigated to an RSS feed page, and I click the Omnivore extension in the toolbar, the extension should notice that I'm on an RSS feed and not a website, and subscribe me to that feed. This would be simpler than the current process of copying the URL and pasting it in the settings screen.

PR #3305, opened 2024-01-03
add browser extension feature to subscribe to feed

aaachen:
https://github.com/omnivore-app/omnivore/issues/3059

Since XML pages are not html, the toolbox/notification cannot be shown (same issue with PDF even though in code it tries to display toolbox). UI to add notification can be iterated later, currently just logs in console.

### DIFF
--- a/pkg/extension/src/scripts/content/prepare-content.js
+++ b/pkg/extension/src/scripts/content/prepare-content.js
@@ -62,6 +62,28 @@
     })
   }
 
+  /**
+   * Determines whether a page is a feed
+   * Currently only feed pages of the XML family and selected feed format in it are recognized (i.e. RSS or Atom)
+   * 
+   * @returns true if it's feed page, false otherwise
+   */
+  function isFeedPage() {
+    const fileExtension = window.location.pathname.slice(-4).toLowerCase()
+    const hasXMLExtension = fileExtension === '.xml'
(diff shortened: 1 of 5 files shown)


pkg/extension/src/scripts/content/prepare-content.js
```diff
@@ -62,6 +62,28 @@
...
+  /**
+   * Determines whether a page is a feed
+   * Currently only feed pages of the XML family and selected feed format in it are recognized (i.e. RSS or Atom)
+   * 
+   * @returns true if it's feed page, false otherwise
+   */
+  function isFeedPage() {
+    const fileExtension = window.location.pathname.slice(-4).toLowerCase()
+    const hasXMLExtension = fileExtension === '.xml'
+    const xmlContentTypes = [
```
jacksonh, line 74: I think we have a few other types we handle here /cc @sywhb

sywhb: yeah, i think `.rss` and `.html` could belong to here too

jacksonh: ah the problem with HTML though is then it wont work for nomal pages, so maybe we have to just be conservative and handle XML here, even though a lot of sites return it as HTML.

PR #3307, opened 2024-01-03
Add german translation

LennardSchwarz:
# Description

This PR adds German translation for:
- Andoird localization strings
- iOS localization strings

# Questions for reviewers:
- [ ] Could somebody please compile and check if I missed any entries in the translation files?
- [ ] Are there any files I missed?
- [ ] Do you want me to translate the [AppStore localization](https://github.com/omnivore-app/docs.omnivore/tree/main/Documents-for-AppStore) as well? Is there something similar for Android?


jacksonh:
Now that a couple emergency releases to the clients are out I am going to get these translations merged and released.

jacksonh:
@LennardSchwarz yeah the appstore translation would be really handy too. I will investigate if the same is available on the Play store.

PR #3312, opened 2024-01-04
Update app/build.gradle to allow debug builds w/o a keystore

gannonlawlor:
# Context

When you first import and try to build the `android` project. You are met with the following error:

```
A problem occurred evaluating project ':app'.
omnivore\android\Omnivore\app\external\keystore.properties (The system cannot find the file specified)
```

This will only occur for new  (or potential new) contributors as they won't have a `keystore.properties` setup. Since Android studio will automatically sign debug builds with the [default keystore](https://developer.android.com/studio/publish/app-signing#debug-mode) . Requiring a custom one for debug seems like a extra unnecessary step to start building

# Fix

This change checks to verify that `external/keystore.properties` exists before loading it and checks to see if `demoKeyPassword` and `demoStorePassword` exist before creating a debug signing config.

### DIFF
--- a/android/Omnivore/app/build.gradle
+++ b/android/Omnivore/app/build.gradle
@@ -8,7 +8,9 @@ plugins {
 
 def keystorePropertiesFile = rootProject.file("app/external/keystore.properties");
 def keystoreProperties = new Properties()
-keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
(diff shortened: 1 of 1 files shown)


jacksonh:
awesome thanks!

jacksonh (approved):

PR #3316, opened 2024-01-04
improve rss feeder performance

sywhb:


packages/api/package.json
```diff
@@ -82,6 +82,7 @@
     "pg": "^8.3.3",
     "postgrator": "^4.2.0",
     "private-ip": "^2.3.3",
+    "redis": "^4.3.1",
```
jacksonh, line 85: :-)

packages/api/src/server.ts
```diff
@@ -157,6 +158,8 @@ const main = async (): Promise<void> => {
   // as healthy.
   await appDataSource.initialize()
 
+  await redisClient.connect()
```
jacksonh, line 161: Maybe to make self hosters live easier we can make this optional, like `if env.redis { await redisClient.connect() }`

sywhb: hah, good timing. i just added redis to the docker compose. will revert the change 👍

jacksonh: Yeah i mean ideally we can make redis optional for API, it doesn't look like that would be too hard since its mostly contained still

sywhb: yeah redis should be optional

packages/api/src/server.ts
```diff
@@ -157,6 +158,11 @@ const main = async (): Promise<void> => {
   // as healthy.
   await appDataSource.initialize()
 
+  // redis is optional
+  if (env.redis.url) {
```
jacksonh, line 162: ❤️

packages/api/src/utils/helpers.ts
```diff
@@ -407,3 +408,16 @@ export const isRelativeUrl = (url: string): boolean => {
...
 }
+
+export const setRecentlySavedItemInRedis = async (
+  userId: string,
+  url: string
+) => {
+  // save the url in redis for 8 hours so rss-feeder won't try to re-save it
+  const redisKey = `recent-saved-item:${userId}:${url}`
+  const ttlInSeconds = 60 * 60 * 8
+  return redisClient.set(redisKey, 1, {
```
jacksonh, line 419: one question: does this throw an exception? I think if this set fails (or if the get fails), we still want to save the item and not fail the job,.

packages/rss-handler/src/index.ts
```diff
@@ -53,6 +58,30 @@ export const isOldItem = (item: RssFeedItem, lastFetchedAt: number) => {
   )
 }
 
+const feedFetchFailedRedisKey = (feedUrl: string) =>
+  `feed-fetch-failure:${feedUrl}`
+
+const isFeedBlocked = async (feedUrl: string, redisClient: RedisClient) => {
```
jacksonh, line 64: awesome

jacksonh: we probably do need a way of logging when this happens. Maybe just a log line is good enough for now

jacksonh: Ideally would be a posthog event but we don't have that all setup yet

packages/rss-handler/src/index.ts
```diff
@@ -214,13 +243,34 @@ const sendUpdateSubscriptionMutation = async (
   }
 }
 
+const isItemRecentlySaved = async (
+  redisClient: RedisClient,
+  userId: string,
+  url: string
+) => {
+  const key = `recent-saved-item:${userId}:${url}`
+  const result = await redisClient.get(key)
```
jacksonh, line 252: Same thing about exception, if it fails to connect to redis i think we can just continue and assume the item isn't saved already

jacksonh (commented):
Just some comments about fallbacks when connecting to redis fails

packages/rss-handler/src/index.ts
sywhb: noted

packages/rss-handler/src/index.ts
sywhb: noted

packages/api/src/utils/helpers.ts
sywhb: yeah, this will throw an exception. let me change the part to catch the error and log

sywhb:
hey @jacksonh , i have updated the code and could you please review again?

jacksonh (approved):

PR #3319, opened 2024-01-05
[Android] Update LibraryView to use ModalBottomSheet

gannonlawlor:
# Context

When using the "BottomSheets" before on the LibraryView, there was a custom composable created to emmulate how a BottomSheet is supposed to look and behave. There were issues with dismissing the bottom sheet by swiping away, you couldn't tap away off the bottom sheet to dismiss, and hitting the back button would close the app and not dismiss the bottomsheet

# Fix

1. Have a BottomSheetState which controls which bottomsheet is shown. This will scale much better than having a flag for each bottomsheet on the LibraryView
2. Use that flag on LibraryView and show the proper Composeable in `ModalBottomSheet`
3. Once again, when building for debug. I ran into a issue with EventTracker where the analytics system was not initialized since I didn't fill in the secrets.xml with a valid api key. I have no plans to create a account with that analytics service. I figure y'all don't care about analytics for developers. So I just created a isDebug in eventtracker which won't initialize if is debug

`Sorry for the length of this PR`

### DIFF
--- a/android/Omnivore/app/src/main/java/app/omnivore/omnivore/EventTracker.kt
+++ b/android/Omnivore/app/src/main/java/app/omnivore/omnivore/EventTracker.kt
@@ -24,12 +24,15 @@ class EventTracker @Inject constructor(val app: Context) {
     PostHog.setSingletonInstance(posthog)
   }
 
-  fun registerUser(userID: String, intercomHash: String?) {
+  fun registerUser(userID: String, intercomHash: String?, isDebug: Boolean) {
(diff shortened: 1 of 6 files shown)


jacksonh:
Ah i see we have someone that understands jetpack way more than me here ❤️

jacksonh:
Hoping to do an Android release later today. Thanks!

gannonlawlor:
Release to store? Wouldn't do that on a Friday tbh

PR #3330, opened 2024-01-06
Uncomment theme colors and use them in bottomsheet

gannonlawlor:
Uncomment all the colors. Also use theme instead of hardcoding colors for the bottom sheet content

# Light Mode 

Screenshot_1704556745 

# Dark Mode

Screenshot_1704556720


jacksonh:
Nice, we have some new dark mode colours I'll try to add after pulling this in.

PR #3353, opened 2024-01-10
Create a debug package name so release and debug can be on same device

gannonlawlor:
# Context

This will allow developers to have both production and dev version on their phones at the same time. Also changed the name of the Dev version to Dev - Omnivore

If someone has the orginal icon files, I would prefer to change the debug icon to be a different color to distinguish it more. But  I couldn't find the main psd or svg file that the pngs are generated from


jacksonh:
@HW05 can you make a special version of the icon with a little `DEBUG` on the bottom?

HW05:
@jacksonh yeah sure!

jacksonh:
Hey @gannonlawlor does this work for the icon? 
app-icon

PR #3357, opened 2024-01-11
Some updates to eslint

jacksonh:


jacksonh:
put up another method of doing this

PR #3358, opened 2024-01-11
Fix issues with eslint from liqe library

jacksonh:


packages/liqe/package.json
```diff
@@ -28,8 +28,6 @@
     "benny": "^3.7.1",
     "coveralls": "^3.1.1",
     "del-cli": "^4.0.1",
-    "eslint": "^7.32.0",
-    "eslint-config-canonical": "^28.0.0",
```
sywhb, line 32: nice 👍

jacksonh:
I'll merge this one so we can see what happens

PR #3359, opened 2024-01-11
Use bullmq to process jobs

jacksonh:
There are a few pieces here that are important and a few that can be ignored.

The important changes:

- in `packages/api` I have created a new entry point `queue-processor` that is used when the backend is started in queue processor mode. In a normal setup we'd deploy a scaling group of backend instances in api mode, and a scaling group of instances in queue-processor mode.

- in `packages/api` I have modified the `rss/fetchAll` endpoint to simply enqueue a `refresh-all-feeds` job.

- in `packages/api` I added a job handler: `refresh-all-feeds`: this will perform the normal refresh query, and then for each subscription group enqueue a `refresh-feed` job. It explicitly sets the job id to a hash based on the feed URL and the list of subscribed users. This means if fetch all was called twice before the job was processed, it would not be added to the queue twice. NOTE: setting the job id does have some consequences. bullmq has a 'completed' state for jobs, normally it will keep completed jobs around for a period of time, or until the queue of compelted jobs is too long. Because we don't want these completed jobs to interfere, we have to mark these jobs to remove immediately after being completed. 

- I migrated code from `rss-handler` into `packages/api/src/jobs/rss` and created a new `refresh-feed` job. This is basically identical to the old handler and will still use cloud task to create the refresh jobs.

The other code can mostly be ignored, but i started to create an admin view for bullmq. We will need this to know whats going on with bull and to be able to purge queues.

## Issues

- One outstanding issue i have is the redis connection used by the rss handlers. Because `queue-processor` is now a long running process instead of just handling individual http requests, that redis connection needs to stay alive and reconnect many times. Right now I think the reconnection strategy will have it shut down eventually.

- I copied in the `task.ts` file from `rss-handler` we probably shouldn't use this and use some of our other service code.

The biggest issue i noticed in testing is that these redis connections can hang and redis will eventually be starved of connections. We need to ensure clean shutdown when the queue-processor jobs are shut down. We will probably need k8s to reap old instances of the queue processors periodically and restart them.


packages/api/src/jobqueue.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const rssRefreshFeedJobQueue = new Queue('rssRefreshFeed', {
+  connection,
+})
+
+export const addRefreshFeedJob = async (jobid: string, payload: any) => {
+  return rssRefreshFeedJobQueue.add('rssRefreshFeed', payload, {
+    jobId: jobid,
+    removeOnComplete: true,
+    removeOnFail: true,
```
sywhb, line 27: Nice

packages/api/src/utils/createTask.ts
```diff
@@ -649,6 +651,12 @@ export const enqueueRssFeedFetch = async (
     folders: subscriptionGroup.folders,
   }
 
+  let jobid = `refresh-feed_${stringToHash(
+    subscriptionGroup.url
+  )}_${stringToHash(JSON.stringify(subscriptionGroup.userIds.sort()))}`
```
sywhb, line 656: Do we need user id here after we already group the subscriptions by url?

jacksonh: @sywhb I think we still do, because when the feed is refreshed its added to the supplied list of users. So imagine we refreshed the feed by just URL, then another user added the feed, then we refreshed again. I think if the key was still "in flight" it would generate two entries, right?

sywhb: yeah, you are right

sywhb:
> The biggest issue i noticed in testing is that these redis connections can hang and redis will eventually be starved of connections. We need to ensure clean shutdown when the queue-processor jobs are shut down. We will probably need k8s to reap old instances of the queue processors periodically and restart them.

we call `quit` command when the http request is completed and it will tell redis server to close the connection with the client.
```
    } finally {
      await redisClient.quit()
      console.log('Redis client disconnected')
    }
```

so i think we can safely keep the connection with redis server from the queue-processor and quit the connection after gracefully shutting down the app as long as we keep the number of redis connections under a certain number, say 100 instances running concurrently

jacksonh:
> > The biggest issue i noticed in testing is that these redis connections can hang and redis will eventually be starved of connections. We need to ensure clean shutdown when the queue-processor jobs are shut down. We will probably need k8s to reap old instances of the queue processors periodically and restart them.
> 
> we call `quit` command in rss-handler when the http request is completed and it will tell redis server to close the connection with the client.
> 
> ```
>     } finally {
>       await redisClient.quit()
>       console.log('Redis client disconnected')
>     }
> ```
> 
> so i think we can safely keep the connection with redis server from the queue-processor and quit the connection after gracefully shutting down the app as long as we keep the number of redis connections under a certain number, say 100 instances running concurrently

oh thank you. I was looking for this part

packages/api/src/gcp-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+// `omnivore-project/secrets/my-secrets/latest`
+//
+export const loadEnvFromGCPSecrets = async (): Promise<
+  BackendEnv | undefined
+> => {
+  if (process.env.GCP_SECRETS_NAME && process.env.GCP_PROJECT_ID) {
+    const client = new SecretManagerServiceClient()
+    const [version] = await client.accessSecretVersion({
+      name: `projects/${process.env.GCP_PROJECT_ID}/secrets/${process.env.GCP_SECRETS_NAME}/versions/latest`,
+    })
```
sywhb, line 19: not sure if this could fail and how we handle it

packages/api/src/queue-processor.ts
```diff
@@ -0,0 +1,126 @@
...
+  const gracefulShutdown = async (signal: string) => {
+    console.log(`[queue-processor]: Received ${signal}, closing server...`)
+    server.close(async () => {
+      console.log(
+        '[queue-processor]: Server closed. shuting down and exiting process...'
+      )
+      await worker.close()
+      await redisDataSource.shutdown()
+      process.exit(0)
+    })
```
sywhb, line 116: will this be called twice when server is closed?

pkg/bull-queue-admin/index.js
```diff
@@ -0,0 +1,114 @@
...
+        tls: {
+          ca: secrets.REDIS_CERT,
+          rejectUnauthorized: false,
+        },
+      }
+    }
+    return {}
+  }
+
+  const connection = new Redis(secrets.REDIS_URL, redisOptions(secrets))
```
sywhb, line 59: REDIS_URL could be null

sywhb (approved):
LGTM and just some small comments

packages/api/src/gcp-utils.ts
jacksonh: Yeah just removed all this, I handle this in k8s now

packages/api/src/util.ts
```diff
@@ -205,11 +177,17 @@ const envParser =
...
+interface Dict<T> {
+  [key: string]: T | undefined
+}
+
+export function getEnv(from: Dict<string>): BackendEnv {
   // Dotenv parses env file merging into proces.env which is then read into custom struct here.
   dotenv.config()
 
-  const parse = envParser(process.env)
+  process.env
```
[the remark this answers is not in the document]
jacksonh: 😬

packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -0,0 +1,88 @@
...
+  payload: any
+): Promise<Job | undefined> => {
+  const queue = createBackendQueue()
+  if (!queue) {
+    return undefined
+  }
+  return queue.add('refresh-feed', payload, {
+    jobId: jobid,
+    removeOnComplete: true,
+    removeOnFail: true,
```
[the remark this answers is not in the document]
jacksonh: Yeah my understand is ioredis handles the retry on error here, but if not, we should probably throw an exception so the whole task is retried.

PR #3364, opened 2024-01-12
fix: change user email address after user deleted account

sywhb:


jacksonh (commented):
One question...what happens if they delete twice?

sywhb:
> One question...what happens if they delete twice?

Both accounts would be marked as deleted in db

jacksonh (approved):

jacksonh:
> > One question...what happens if they delete twice?
> 
> Both accounts would be marked as deleted in db

but would there be a conflict if they had the same email?

jacksonh (approved):

sywhb:
> > > One question...what happens if they delete twice?
> > 
> > 
> > Both accounts would be marked as deleted in db
> 
> but would there be a conflict if they had the same email?

the `user_id` of each account would be different

Issue #3377, opened 2024-01-16
Unable to edit labels & Delete -> Recreate entires


0xdeee:
[Screen Recording 16-01-2024 at 12.56.webm](https://github.com/omnivore-app/omnivore/assets/21086243/eef3c490-4141-4b94-97eb-046ca57794e3)

### Issue 1: Unable to edit labels

For some reason, in a specific entry the labels had duplicated so I tried to deduplicate it and met with failure... 

Some Info: I may have renamed the labels before or after assigning them to this entry, but other entires seem fine AFAIC.

#### Request

```
curl 'https://api-prod.omnivore.app/api/graphql' \
  -H 'authority: api-prod.omnivore.app' \
  -H 'accept: */*' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'content-type: application/json' \
  -H 'cookie: XXXX
  -H 'dnt: 1' \
  -H 'origin: https://omnivore.app' \
  -H 'referer: https://omnivore.app/' \
  -H 'sec-ch-ua: "Not_A Brand";v="8", "Chromium";v="120"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: same-site' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36' \
[...]

Issue #3219, opened 2023-12-08
When saving a page with highlights/notes/labels make sure to merge


jacksonh:
If we are saving a page that has a duplicate we need to merge the labels. For example if a user has two RSS feeds that both have the same link, the first RSS feed could add the item, the user could label the item, and then the second feed could add the same item and the labels would be overwritten.

PR #3380, opened 2024-01-16
fix/recreate item

sywhb:
- delete all the labels and highlights attached to the item when item was deleted
- merge labels and highlights when saving a duplicate item

### DIFF
--- a/packages/api/src/resolvers/article/index.ts
+++ b/packages/api/src/resolvers/article/index.ts
@@ -80,6 +80,7 @@ import {
   findLibraryItemByUrl,
   findLibraryItemsByPrefix,
   searchLibraryItems,
+  softDeleteLibraryItem,
   sortParamsToSort,
   updateLibraryItem,
   updateLibraryItemReadingProgress,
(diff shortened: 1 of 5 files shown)


sywhb:
fix #3377 & #3219

jacksonh:
I think this might not be the correct logic, because it does mean if someone undeletes they lose those labels and highlights, right?

sywhb:
> I think this might not be the correct logic, because it does mean if someone undeletes they lose those labels and highlights, right? 

Yeah 😂 you are absolutely right

sywhb:
hey @jacksonh , i have made a change to the code in this branch and it will now only delete labels and highlights attached to the deleted item during recreation.

for the duplicate labels, sorry that i made a big mistake. the unique constraint i created in `entity_labels` table does not work at all because the constraint is set on three columns (`label_id`, `library_item_id`, `highlight_id`) where `library_item_id` or `highlight_id` can be `null` which i thought postgres would treat `null` value as distinct but it doesn't prior to postgres 15.

this could be fixed by replacing the existing unique constraint with two partial unique index where either `library_item_id` is null or `highlight_id` is null but the challenge is to remove all the existing duplicates in the db and we probably have to run a big migration script which might affect the current usage.

jacksonh (approved):

PR #3385, opened 2024-01-17
Helm Chart for self-hosting

se-jaeger:
Hi,

this PR makes some minor changes and adds a basic structure and documentation to show how to self-host omnivore on an existing Kubernetes Cluster.

Have this running since some days/weeks without any issues.


jacksonh:
Nice! We are starting to deploy more of our infra on k8s too, so hopefully we can use some of this for our deployments.

PR #3393, opened 2024-01-19
Fix refresh feed job priority

jacksonh:
- Improve RSS priorities for refreshing feeds
- Add some more debug output

### DIFF
--- a/packages/api/src/jobs/rss/refreshAllFeeds.ts
+++ b/packages/api/src/jobs/rss/refreshAllFeeds.ts
@@ -4,6 +4,7 @@ import { QUEUE_NAME } from '../../queue-processor'
 import { redisDataSource } from '../../redis_data_source'
 import { RssSubscriptionGroup } from '../../utils/createTask'
 import { stringToHash } from '../../utils/helpers'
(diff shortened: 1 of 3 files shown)


sywhb (approved):
Nice

packages/api/src/queue-processor.ts
```diff
@@ -61,11 +61,17 @@ const main = async () => {
     throw '[queue-processor] error redis is not initialized'
   }
 
+  const queue = new Queue(QUEUE_NAME)
```
sywhb, line 64: 👍

packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -52,6 +53,12 @@ const updateSubscriptionGroup = async (group: RssSubscriptionGroup) => {
     console.error('no userlist for feed group', group)
     return
   }
+
+  try {
+    feedURL = validateUrl(feedURL).toString()
+  } catch (err) {
+    console.log('not refreshing invalid feed url: ', { feedURL })
+  }
```
sywhb, line 61: thanks!

PR #3398, opened 2024-01-20
More logging for refreshing feeds

jacksonh:
- Some debug
- Remove non-useful logging line
- Add a refresh context to make debugging rss jobs easier
- Add a refresh context to make debugging refresh runs easier


packages/api/src/jobs/save_page.ts
```diff
@@ -353,11 +357,18 @@ export const savePageJob = async (data: Data, attemptsMade: number) => {
 
   try {
     const url = encodeURI(data.url)
+    console.log(`savePageJob: ${userId} ${url}`)
 
     // get the fetch result from cache
     const { title, content, contentType, readabilityResult } =
       await getCachedFetchResult(url)
 
+    if (!title || !content || !contentType || !readabilityResult) {
```
sywhb, line 366: content and readability can be null if content type is pdf

sywhb (changes requested):

packages/api/src/jobs/save_page.ts
```diff
@@ -382,30 +337,45 @@ export const savePageJob = async (data: Data, attemptsMade: number) => {
...
-      publishedAt,
-      source,
-      folder,
-    })
-    if (!apiResponse) {
-      throw new Error('error while saving page')
+    if (!content) {
+      throw new Error(
+        'Invalid SavePage job, fetch result missing required data'
+      )
```
sywhb, line 343: probably have to remove this check

packages/content-fetch/src/request_handler.ts
```diff
@@ -138,7 +138,7 @@ export const contentFetchRequestHandler: RequestHandler = async (req, res) => {
     console.log('cacheFetchResult result', cacheResult)
 
     const jobs = await queueSavePageJob(savePageJobs)
-    console.log('save-page jobs queued', jobs.length)
+    console.log('save-page jobs queued', { jobs })
```
sywhb, line 141: not sure if we need to log everything here if we want to reduce the log size

sywhb (approved):
LGTM 🚢

packages/api/src/jobs/save_page.ts
jacksonh: I think we need it as we can't pass `originalContent` if its null.

packages/content-fetch/src/request_handler.ts
jacksonh: Yeah I'll remove this now

PR #3399, opened 2024-01-20
Run SwiftFormat manually using plugin system

satindar:
This updates where SwiftFormat is run from. Instead of using the BuildTools command line package this pulls in the SwiftFormat package using the projects main Package.swift file. Doing this gives the option to run SwiftFormat manually by right clicking on the Omnivore workspace in the file viewer. 

This means it will be run manually now instead of using a build phase and running on every build. Maybe that's preferred. Or not! 

SwiftLint also has a plugin that can be triggered from a Build Phase but I was running into package resolution issues when trying to add it to the main package. Something conflicts with the swift markdown package. If we can get that working then we only need BuildTools for GraphQL generation and most users won't have to bother with BuildTools (there was a user a few months back running into issues with BuildTools because of their macOS version). 

Screenshot 2024-01-20 at 8 20 58 AM

### DIFF
--- a/apple/BuildTools/Package.swift
+++ b/apple/BuildTools/Package.swift
@@ -5,7 +5,6 @@ let package = Package(
   name: "BuildTools",
   platforms: [.macOS(.v10_11)],
   dependencies: [
-    .package(url: "https://github.com/nicklockwood/SwiftFormat", from: "0.48.11"),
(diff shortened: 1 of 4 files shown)


satindar:
Didn't commit the changes that SwiftFormat would make. Mostly removing `self` calls where they aren't needed.

PR #3402, opened 2024-01-22
fix: ignore error when saving existing library item in db

sywhb:
- fix: do not throw error when creating duplicate item in db
- fix tests

### DIFF
--- a/packages/api/src/repository/library_item.ts
+++ b/packages/api/src/repository/library_item.ts
@@ -1,9 +1,26 @@
+import { DeepPartial } from 'typeorm'
 import { appDataSource } from '../data_source'
 import { LibraryItem } from '../entity/library_item'
+import { wordsCount } from '../utils/helpers'
 
 export const libraryItemRepository = appDataSource
   .getRepository(LibraryItem)
   .extend({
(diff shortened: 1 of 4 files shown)


packages/api/src/repository/library_item.ts
```diff
@@ -1,9 +1,26 @@
...
+    save(libraryItem: DeepPartial<LibraryItem>) {
+      return this.createQueryBuilder()
+        .insert()
+        .values({
+          ...libraryItem,
+          wordCount:
+            libraryItem.wordCount ??
+            wordsCount(libraryItem.readableContent || ''),
+        })
+        .orIgnore() // ignore if the item already exists
```
jacksonh, line 18: If there was just a random failure, like a connection timeout would this be ignored?

sywhb: no, only unique constraint error will be ignored. i know this `orIgnore()` is misleading

sywhb: here the `orIgnore` is actually `on conflict do nothing`

jacksonh (approved):

PR #3410, opened 2024-01-22
improve contrast on active table row background color

jeffgreco:
A friend is unable to see any change when a table row is set to active due to the very low degree of color contrast shift. This subs in Omnivore yellow color as is used elsewhere in the app for similar purposes.

Open to other solutions but this seemed like a quick one :)


jacksonh:
Hey, any chance you have a screenshot?

jacksonh:
Going to merge this one then overwrite with hex colour and a dark mode change.

PR #3414, opened 2024-01-23
fix: finalUrl defaults to the url of the page saved

sywhb:


packages/puppeteer-parse/src/index.ts
```diff
@@ -201,13 +200,11 @@ export const fetchContent = async (
         page = result.page
       }
       if (result && result.finalUrl) {
-        finalUrl = result.finalUrl
+        url = result.finalUrl
```
jacksonh, line 203: So final URL only gets updated if fetch succeeds, right? One thing I noticed is the cache uses finalurl also, do you think that needs to be updated?

jacksonh (approved):

PR #3416, opened 2024-01-23
Dont pass library item id to backend when updating highlights

jacksonh:
The native (iOS and Android) code needs this param but we
cant send it to the web.


sywhb (approved):
LGTM

PR #3419, opened 2024-01-23
Discover

jacksonh:


jacksonh:
Going to attempt to pull in some of our tsconfig changes to reduce this one a bit before merging.

Podginator (commented):
These filenames will need to changed, as they've already been applied in main.

jacksonh:
@Podginator question on the inter-service communication here. Does API ever call the discover service? I'm going to try and set these up in k8s

PR #3420, opened 2024-01-23
fix: catch rejected promise if it is in a try-catch block

sywhb:


sywhb:
fix this sentry report caused by `unhandled rejected promise`
https://omnivore.sentry.io/issues/4907060105/?project=5466498&referrer=slack

PR #3431, opened 2024-01-24
fix/subscription

sywhb:
- add REFRESH_ERROR to the status of subscriptions and add a new column refreshed_at
- add refreshedAt to the gql schema
- update refreshed_at when feed is processed
- mark subscription as error if item was failed
- validate feed item url and convert relative url to absolute url
- get active user subscriptions only

### DIFF
--- a/packages/api/src/entity/subscription.ts
+++ b/packages/api/src/entity/subscription.ts
@@ -86,6 +86,9 @@ export class Subscription {
   @Column('timestamp', { nullable: true })
   refreshedAt?: Date | null
 
+  @Column('timestamp', { nullable: true })
+  failedAt?: Date | null
+
   @Column('boolean')
   isPrivate?: boolean | null
 
--- a/packages/api/src/entity/subscription.ts
+++ b/packages/api/src/entity/subscription.ts
@@ -16,7 +16,6 @@ export enum SubscriptionStatus {
   Active = 'ACTIVE',
   Deleted = 'DELETED',
   Unsubscribed = 'UNSUBSCRIBED',
-  RefreshError = 'REFRESH_ERROR',
 }
 
 export enum SubscriptionType {
--- a/packages/api/src/entity/subscription.ts
+++ b/packages/api/src/entity/subscription.ts
@@ -7,11 +7,22 @@ import {
   PrimaryGeneratedColumn,
   UpdateDateColumn,
 } from 'typeorm'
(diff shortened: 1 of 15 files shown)


packages/api/src/entity/subscription.ts
```diff
@@ -7,12 +7,23 @@ import {
...
 import { NewsletterEmail } from './newsletter_email'
 import { User } from './user'
 
 export const DEFAULT_SUBSCRIPTION_FOLDER = 'following'
 
+export enum SubscriptionStatus {
+  Active = 'ACTIVE',
+  Deleted = 'DELETED',
+  Unsubscribed = 'UNSUBSCRIBED',
+  RefreshError = 'REFRESH_ERROR',
```
jacksonh, line 19: oh shoot, i have a feeling this could cause issues in other places where we filter based on status. We might need to do ACTIVE || REFRESH_ERROR. I wonder if it should be a separate property.

Actually, maybe it should be a timestamp? refresh_error_at and then we know when it went into error state.

packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -21,28 +20,33 @@ export const refreshAllFeeds = async (db: DataSource): Promise<boolean> => {
...
+        ARRAY_AGG(s.id) AS "subscriptionIds",
+        ARRAY_AGG(s.user_id) AS "userIds",
+        ARRAY_AGG(s.most_recent_item_date) AS "mostRecentItemDates",
+        ARRAY_AGG(coalesce(s.scheduled_at, NOW())) AS "scheduledDates",
+        ARRAY_AGG(s.last_fetched_checksum) AS checksums,
+        ARRAY_AGG(s.fetch_content) AS "fetchContents",
+        ARRAY_AGG(coalesce(s.folder, $3)) AS folders
+      FROM
+        omnivore.subscriptions s
+      INNER JOIN
```
jacksonh, line 34: nice

packages/api/src/entity/subscription.ts
sywhb: good catch! so we filter subscriptions based on status = 'active' in some clients? then a separate column makes sense.

yeah, i agree with `refresh_error_at` and we could do a `error_count` as well

jacksonh (approved):

Issue #3285, opened 2023-12-28
Code block not formatted correctly


sywhb:
https://mp.weixin.qq.com/s/mJlfLdtscRLGB6D5V9xgMw

sywhb:
> @sywhb do we have an ETA for this case? thx

hey, i am pretty occupied this week. will relook at it next week only

a19284:
@sywhb  Kindly let me know if you are available to work on this case, thx

sywhb:
i'm working on this :)

sywhb:
this is fixed now. @a19284 @thiswillbeyourgithub could you please verify it on your end?

thiswillbeyourgithub:
Waaaa it works :D 
![image](https://github.com/omnivore-app/omnivore/assets/26625900/8ebf4528-ed17-4c48-a731-f6f051d001b1)


On web as well as on android. Many thanks @sywhb !!

a19284:
Thanks for your strong support! @sywhb 
- Checked the [link](https://mp.weixin.qq.com/s/mJlfLdtscRLGB6D5V9xgMw) , most code sessions work except one, I attached snapshot below for your reference
- btw, is there any chance to format the code session with different background/font or even line numbers, to improve the readability? apprecicate 

 
![issues](https://github.com/omnivore-app/omnivore/assets/5847069/42d3a821-c435-4b68-a091-a5941d679b53)

PR #3435, opened 2024-01-24
fix/readability

sywhb:
- run readability in save-page instead of puppeteer
- throw error if save-page job failed


sywhb:
fix #3285

jacksonh (approved):

PR #3442, opened 2024-01-26
move trigger rule into a job

sywhb:


jacksonh (approved):
Beautiful. One thing I want to verify: changing the entityCreated stuff wont interfere with webhooks at all, right?

sywhb:
> Beautiful. One thing I want to verify: changing the entityCreated stuff wont interfere with webhooks at all, right?

no because we still publish the pubsub event while enqueuing a new job to trigger rules. i will disable the rule pubsub subscription in demo so no duplicate rules are triggered.

next we can move webhook into a job too

PR #3454, opened 2024-01-29
reduce log size

sywhb:
- remove redundant logs
- reduce the size of the log entry by truncating any string values to 500 characters

### DIFF
--- a/packages/api/src/utils/logger.ts
+++ b/packages/api/src/utils/logger.ts
@@ -34,11 +34,9 @@ export class CustomTypeOrmLogger
   }
 
   logQuery(query: string, parameters?: any[], queryRunner?: QueryRunner) {
-    this.logger.info(
-      `query: ${query} -- PARAMETERS: ${super.stringifyParams(
-        parameters || []
-      )}`
-    )
+    this.logger.info(query, {
+      parameters,
+    })
   }
 
   log(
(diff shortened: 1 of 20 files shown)


packages/readabilityjs/Readability.js
```diff
@@ -3082,7 +3082,6 @@ Readability.prototype = {
     this._removeScripts(this._doc);
 
     this._prepDocument();
-    console.log(this._doc.body.innerHTML);
```
sywhb, line 3085: @jacksonh i removed this is the line which caused we log full html content

packages/api/src/utils/logger.ts
```diff
@@ -34,11 +34,9 @@ export class CustomTypeOrmLogger
   }
 
   logQuery(query: string, parameters?: any[], queryRunner?: QueryRunner) {
-    this.logger.info(
-      `query: ${query} -- PARAMETERS: ${super.stringifyParams(
-        parameters || []
-      )}`
-    )
```
sywhb, line 41: @jacksonh i changed this line to make sql parameters part of the log entry object so we can reduce the size of parameters accordingly

packages/api/src/utils/logger.ts
```diff
@@ -133,10 +131,8 @@ const truncateObjectDeep = (object: any, length: number): any => {
 
 class GcpLoggingTransport extends LoggingWinston {
   log(info: any, callback: (err: Error | null, apiResponse?: any) => void) {
-    const sizeInfo = jsonStringify(info).length
-    if (sizeInfo > MAX_LOG_SIZE) {
-      info = truncateObjectDeep(info, 500) as never // the max length for string values is 500
-    }
```
sywhb, line 139: @jacksonh this is the change where we reduce the size of logs by truncating any string value to 500 char; e.g. `content` in gql request and sql parameters

jacksonh (approved):

PR #3460, opened 2024-01-30
Write behind cache for reading progress

jacksonh:


packages/api/src/datasources/reading_progress_data_source.ts
```diff
@@ -0,0 +1,84 @@
...
+      updatedAt: new Date(),
+    }
+
+    this.cacheItems[cacheKey] = cacheItem
+    if (await redisDataSource.redisClient?.hmset(cacheKey, cacheItem)) {
+      console.log('cached reading progress')
+    } else {
+      console.log('failed to cache reading progress')
+    }
+  }
```
sywhb, line 56: @jacksonh i'm not sure if concurrency here will be a problem.

this could be rare but if two updates of reading progress happen at the same time and probably the old value could be used.

jacksonh: yeah i agree, the whole thing feels a little brittle.

packages/api/src/queue-processor.ts
```diff
@@ -132,6 +160,21 @@ const main = async () => {
     }
   )
 
+  const queue = await getBackendQueue()
+  if (queue) {
+    await queue.add(
```
sywhb, line 165: i think we might want to use `jobId` here to deduplicate jobs

sywhb: if we have ten instances, will it create ten jobs?

sywhb:
i think the consistency problem might exist. when there is another reading progress api request coming in while we are updating the progress in db.

we read the value from the list before another value is added to the list then the latest value will be lost when we delete the key.

packages/api/src/queue-processor.ts
jacksonh: I _think_ bullmq handles this for us. It generates a unique jobid for each run using the job name. I think beca

sywhb (approved):

jacksonh:
rebased everything

PR #3466, opened 2024-01-30
create a job for db trigger

sywhb:


jacksonh (approved):

packages/api/src/queue-processor.ts
```diff
@@ -34,6 +47,43 @@ export const getBackendQueue = async (): Promise<Queue | undefined> => {
   return backendQueue
 }
 
+export const createWorker = (connection: ConnectionOptions) =>
```
jacksonh, line 50: 🚀

packages/api/src/services/labels.ts
```diff
@@ -218,12 +233,32 @@ export const deleteLabels = async (
...
+
+  if (!deleteResult.affected) {
+    return false
+  }
+
+  const data = libraryItemIds.map((libraryItemId) => ({
+    libraryItemId,
+    userId,
+  }))
+  await bulkEnqueueUpdateLabels(data)
```
jacksonh, line 251: nice way of doing it

jacksonh (commented):
So does this mean highlights and labels are not cached on the library item anymore?

sywhb:
> So does this mean highlights and labels are not cached on the library item anymore?

no, they will still be cached if i understand the question correctly

PR #3468, opened 2024-01-31
Prevent add label if press enter in composition session

kawamataryo:
Thank you for this fantastic app!

As a user from Japan, I came across an issue where a label was erroneously being created upon pressing the 'Enter' key to confirm Japanese Kanji conversion during the label addition process. I have resolved this problem.

ref: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing

CleanShot 2024-01-31 at 13 43 43

### DIFF
--- a/pkg/extension/src/scripts/content/toast.js
+++ b/pkg/extension/src/scripts/content/toast.js
@@ -575,7 +575,7 @@
       }
       case 'enter': {
         if (event.target.id == 'omnivore-edit-label-input') {
-          if (event.target.value) {
+          if (event.target.value && !event.isComposing) {
             const labelList = event.target.form.querySelector('#label-list')
             addLabel(labelList, event.target, event.target.value)
           }


jacksonh:
wow this is really interesting, I did not know about this

kawamataryo:
@jacksonh 
Thank you for your comment! Given that it is a function specific to Japanese character input, it's not very common. 

Below are the changes before and after:

### before

https://github.com/omnivore-app/omnivore/assets/11070996/824f55ab-fb8c-4e9b-a778-7bb95cac702d

### After

https://github.com/omnivore-app/omnivore/assets/11070996/c9608b8b-6076-4b8e-ac37-0ae09e010d52

The `event.isComposing` is always `false` on English or other language inputs, so it shouldn't have any adverse effects.

jacksonh:
@kawamataryo do you mind signing the CLA thing?

kawamataryo:
@jacksonh 
Sorry, I had missed it. I just now signed.

PR #3476, opened 2024-02-01
fix: update-label job uses too much cpu

sywhb:
- remove the inner join with highlights table in the sql of update-label job
- change the update-db job priority to 5 and use exponential backoff strategy for retrying


sywhb:
next we need to figure out a plan to merge highlight labels into the label_names column in library_item table too

jacksonh (approved):

PR #3482, opened 2024-02-01
fix: dedupe update-db job by using library_item_id as job_id

sywhb:


jacksonh:
We need to test it out a bit, i had some trouble initially using this trick, i think maybe because the items stayed in the completed state, but the docs say that should work.

jacksonh (approved):

sywhb:
i will rebase it and test

PR #3484, opened 2024-02-01
fix: replace webhook endpoint with a bullmq job

sywhb:


packages/api/src/utils/createTask.ts
```diff
@@ -669,6 +670,20 @@ export const enqueueTriggerRuleJob = async (data: TriggerRuleJobData) => {
...
 }
 
+export const enqueueWebhookJob = async (data: CallWebhookJobData) => {
+  const queue = await getBackendQueue()
+  if (!queue) {
+    return undefined
+  }
+
+  return queue.add(CALL_WEBHOOK_JOB_NAME, data, {
+    priority: 1,
```
jacksonh, line 680: I think priority should be lower for this one. Probably 5 or 10

jacksonh (changes requested):
Just need to look at the priority. Probably we should make a dashboard or tool to list all of them in one place

packages/api/src/utils/createTask.ts
sywhb: yeah it makes sense.

i think we can set priority = 5 for this and trigger-rule jobs

sywhb:
> Just need to look at the priority. Probably we should make a dashboard or tool to list all of them in one place

we could probably show these in the admin portal

jacksonh (approved):

PR #3488, opened 2024-02-02
fix: drop db triggers which updates labels and highlights in library_item table for locking the table

sywhb:


jacksonh:
This is already applied, right?

sywhb:
> This is already applied, right?

no, not yet

jacksonh (approved):

PR #3492, opened 2024-02-03
Add an index to the subscriptions name column

jacksonh:


packages/db/migrations/0161.do.add_subscriptions_name_index.sql
```diff
@@ -0,0 +1,7 @@
+-- Type: DO
+-- Name: add_subscriptions_name_index
+-- Description: Add an index to the subscriptions name column
+
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS subscriptions_name_index ON omnivore.subscriptions (name);
```
sywhb, line 6: i think the index should be on `user_id` and `name`

sywhb (changes requested):

jacksonh:
Thanks agreed on adding user id. Here are the two where statements pganalyze found:

`WHERE ( ("Subscription"."name" = $1 AND "Subscription"."type" = $2 AND "Subscription__Subscription_user"."id" = $3) ) AND ( "Subscription"."id" IN ($0) )`

and

`WHERE ("Subscription"."name" = $1 AND "Subscription"."type" = $2 AND "Subscription__Subscription_user"."id" = $3)`

sywhb (approved):

PR #3495, opened 2024-02-03
feat/save page retries

jacksonh:
- Backoff for save page
- Add retries for pdf handler job

### DIFF
--- a/packages/content-fetch/src/job.ts
+++ b/packages/content-fetch/src/job.ts
@@ -50,7 +50,7 @@ const getOpts = (job: savePageJob): BulkJobOptions => {
     priority: getPriority(job),
     backoff: {
       type: 'exponential',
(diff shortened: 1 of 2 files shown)


packages/content-fetch/src/job.ts
```diff
@@ -48,6 +48,10 @@ const getOpts = (job: savePageJob): BulkJobOptions => {
     // removeOnFail: true,
     attempts: getAttempts(job),
     priority: getPriority(job),
+    backoff: {
+      type: 'exponential',
+      delay: 1000,
```
sywhb, line 53: should we also change `delay` to `2000`?

jacksonh: agreed!

sywhb (approved):

PR #3502, opened 2024-02-05
increase the seed valud of backoff delay to 2 seconds

sywhb:


packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -137,8 +137,6 @@ export const queueRSSRefreshFeedJob = async (
   }
   return queue.add('refresh-feed', payload, {
     jobId: jobid,
-    removeOnComplete: true,
```
jacksonh, line 140: I think we might need this for this job because of the jobid hack

packages/api/src/queue-processor.ts
```diff
@@ -49,6 +49,14 @@ export const getBackendQueue = async (): Promise<Queue | undefined> => {
   }
   backendQueue = new Queue(QUEUE_NAME, {
     connection: redisDataSource.workerRedisClient,
+    defaultJobOptions: {
+      backoff: {
+        type: 'exponential',
+        delay: 2000, // 2 seconds
+      },
+      removeOnComplete: true,
```
jacksonh, line 57: Have to test but this could mess up the bullmq UI

packages/api/src/jobs/rss/refreshAllFeeds.ts
sywhb: yeah, i moved this config into queue configuration

packages/api/src/queue-processor.ts
sywhb: yeah, how about we keep jobs based on their age?

for example, we could only keep failed jobs to 24 hours so we can avoid out of memory error and save costs

jacksonh: Yeah agreed. I think it does do that now, but I'm not exactly sure what the default parameters are.

jacksonh:
I think this is good, I'd like to mess around with it a bit on demo to make sure the removeOnComplete stuff doesn't interfere with the jobid stuff  / dashboard stuff too much.

jacksonh (approved):

packages/api/src/queue-processor.ts
sywhb: let me set the parameters explicitly and test

sywhb: @jacksonh i think by default bullmq keeps all the completed and failed jobs so here i change it to only keep completed jobs for 24 hours and failed job for 7 days

i think the bullmq ui will be messed up once this is live

PR #3503, opened 2024-02-05
do not update db immediately for force update of reading progress

sywhb:


jacksonh:
Ah the failures are probably because some of the tests rely on the DB being updated :-/

sywhb:
> Ah the failures are probably because some of the tests rely on the DB being updated :-/

i think it's the returned value not updated. just fixed that

packages/api/src/resolvers/article/index.ts
```diff
@@ -619,13 +619,43 @@ export const saveArticleReadingProgressResolver = authorized<
...
+    // because the function resolver will handle that for us when
+    // it resolves the properties of the Article object
+    let updatedItem = await findLibraryItemById(id, uid)
+    if (!updatedItem) {
+      return {
+        errorCodes: [SaveArticleReadingProgressErrorCode.Unauthorized],
+      }
+    }
+
+    if (env.redis.cache && env.redis.mq) {
```
jacksonh, line 633: We don't have to right now, but we can probably remove all the `if (redis)` type checks in backend now and make it mandatory, since we are simplifying so much of self hosting with it.

jacksonh (approved):

PR #3523, opened 2024-02-09
Android - Update label chips to M3

stefanosansone:
This PR resolves a crash in the previous build caused by a chips library incompatibility with the new Compose version. It includes the following:

- Updates to the chips library and Compose for compatibility
- Migration of the chips component to Material 3 (M3)
- General code cleanup for improved readability


jacksonh:
0_0

PR #3545, opened 2024-02-20
fix/pg pool connection timeout

sywhb:
- **get postgres connection pool size from env**
- **connection idle timeout in 10 seconds**

### DIFF
--- a/packages/api/src/queue-processor.ts
+++ b/packages/api/src/queue-processor.ts
@@ -153,8 +153,12 @@ const main = async () => {
     subscribers: [__dirname + '/events/**/*{.js,.ts}'],
     namingStrategy: new SnakeNamingStrategy(),
     logger: new CustomTypeOrmLogger(['query', 'info']),
-    connectTimeoutMS: 40000, // 40 seconds
+    connectTimeoutMS: 10000, // 10 seconds
     maxQueryExecutionTime: 10000, // 10 seconds
+    extra: {
+      max: env.pg.pool.max,
+      idleTimeoutMillis: 10000, // 10 seconds
+    },
   })
 
   // respond healthy to auto-scaler.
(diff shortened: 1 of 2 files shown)


packages/api/src/queue-processor.ts
```diff
@@ -153,8 +153,12 @@ const main = async () => {
     subscribers: [__dirname + '/events/**/*{.js,.ts}'],
     namingStrategy: new SnakeNamingStrategy(),
     logger: new CustomTypeOrmLogger(['query', 'info']),
-    connectTimeoutMS: 40000, // 40 seconds
+    connectTimeoutMS: 10000, // 10 seconds
     maxQueryExecutionTime: 10000, // 10 seconds
+    extra: {
```
jacksonh, line 158: are we losing this `options: process.env.PG_EXTRA_OPTIONS,` ? Right now we use that to set the lock timeout.  its set like this in k8s `PG_EXTRA_OPTIONS: '-c lock_timeout=500ms'`

sywhb: > are we losing this `options: process.env.PG_EXTRA_OPTIONS,` ? Right now we use that to set the lock timeout. its set like this in k8s `PG_EXTRA_OPTIONS: '-c lock_timeout=500ms'`

no, we won't but i noticed we didn't have it for queue-processor in the first place. did you do it on purpose?

sywhb: the option was set in api-backend only

jacksonh: Oh, yeah I think we will want it everywhere

jacksonh: Currently i think the AppEngine queue-processor is using it

sywhb: but it is not in the code

sywhb: ohh, you set it on cloud sql

sywhb: @jacksonh i have removed the `setOptions` in queue-processor

jacksonh (approved):

PR #3553, opened 2024-02-21
fix: save url operation is delayed

sywhb:
normal save url operation could be delayed by jobs from lower priority queue so we deprioritize these jobs by setting priority to 10

### DIFF
--- a/packages/api/src/jobs/rss/refreshAllFeeds.ts
+++ b/packages/api/src/jobs/rss/refreshAllFeeds.ts
@@ -3,9 +3,12 @@ import { DataSource } from 'typeorm'
 import { v4 as uuid } from 'uuid'
 import { getBackendQueue, JOB_VERSION } from '../../queue-processor'
 import { validateUrl } from '../../services/create_page_save_request'
-import { RssSubscriptionGroup } from '../../utils/createTask'
+import { getJobPriority, RssSubscriptionGroup } from '../../utils/createTask'
 import { stringToHash } from '../../utils/helpers'
 import { logger } from '../../utils/logger'
+
+export const REFRESH_ALL_FEEDS_JOB_NAME = 'refresh-all-feeds'
+export const REFRESH_FEED_JOB_NAME = 'refresh-feed'
 
 export type RSSRefreshContext = {
   type: 'all' | 'user-added'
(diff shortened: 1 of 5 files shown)


packages/api/src/utils/createTask.ts
```diff
@@ -37,6 +42,37 @@ import View = google.cloud.tasks.v2.Task.View
...
+/**
+ * we want to prioritized jobs by the expected time to complete
+ * lower number means higher priority
+ * priority 1: jobs that are expected to run immediately
+ * priority 5: jobs that are expected to run in less than 10 seconds
+ * priority 10: jobs that are expected to run in less than 1 minute
+ * priority 50: jobs that are expected to run in less than 30 minutes
+ * priority 100: jobs that are expected to run in less than 1 hour
+ **/
+export const getJobPriority = (jobName: string): number => {
```
jacksonh, line 54: Ah this is nice. One thing I'd say is we should probably log when we hit default so we handle jobs not having priority set properly.

sywhb:
> One question on this that I couldn't tell: currently we have two ways a feed can be refreshed: either from the cron job or by a user adding it. When the user adds it we want to give a high priority refresh. Do we lose that with this change?

i fixed the bug in the previous code. the user-added rss feed was used to be given a lower priority(higher number)

packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -129,15 +132,15 @@ type QueuePriority = 'low' | 'high'
 export const queueRSSRefreshFeedJob = async (
   jobid: string,
   payload: any,
-  options = { priority: 'high' as QueuePriority }
```
sywhb, line 132: @jacksonh the default priority previously was `high` which was assigned to user added rss feed refresh job

packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -129,15 +132,15 @@ type QueuePriority = 'low' | 'high'
...
+  options = { priority: 'low' as QueuePriority }
 ): Promise<Job | undefined> => {
   const queue = await getBackendQueue()
   if (!queue) {
     return undefined
   }
-  return queue.add('refresh-feed', payload, {
+  return queue.add(REFRESH_FEED_JOB_NAME, payload, {
     jobId: `${jobid}_${JOB_VERSION}`,
-    priority: options.priority == 'low' ? 10 : 50,
```
sywhb, line 140: @jacksonh the priority for `high` is larger than `low` and this means user added rss feeds were processed after cronjob triggered jobs

PR #3554, opened 2024-02-21
Android - Fix current item edit

stefanosansone:
Fix current item edit in library by replacing livedata with the ui state flow


jacksonh:
Oh man, so simple. I couldn't figure out how to do that.

PR #3566, opened 2024-02-23
perf/optimize slow query

sywhb:
- **create a multi-column index on user_id and state columns**
- **use low priority queue if user saved 5 items in the last minute. previously it was 5 items created**
- **add an index for api key**
- **create an index for upload_file_id in library_item table**


packages/api/src/services/create_page_save_request.ts
```diff
@@ -38,12 +38,12 @@ const SAVING_CONTENT = 'Your link is being saved...'
...
 const isPrivateIP = privateIpLib.default
 
-// 5 articles added in the last minute: use low queue
+// 5 items saved in the last minute: use low queue
 // default: use normal queue
 const getPriorityByRateLimit = async (
   userId: string
 ): Promise<'low' | 'high'> => {
-  const count = await countByCreatedAt(userId, new Date(Date.now() - 60 * 1000))
+  const count = await countBySavedAt(userId, new Date(Date.now() - 60 * 1000))
```
jacksonh, line 46: nice catch

packages/api/src/services/update_pdf_content.ts
```diff
@@ -40,8 +40,7 @@ export const updateContentForFileItem = async (msg: UpdateContentMessage) => {
         .withRepository(libraryItemRepository)
         .createQueryBuilder('item')
         .innerJoinAndSelect('item.uploadFile', 'file')
-        .where('item.user = :userId', { userId: uploadFile.user.id })
-        .andWhere('file.id = :fileId', { fileId })
+        .where('file.id = :fileId', { fileId })
```
jacksonh, line 43: the user id check isn't needed because of RLS?

jacksonh (approved):

packages/api/src/services/update_pdf_content.ts
sywhb: > the user id check isn't needed because of RLS?

yes.

also, it could help postgres to use the `upload_file_id` index when issuing the query

PR #3567, opened 2024-02-23
feature/feed summary

sywhb:
- **rename preview_content to feed_content in library_item table and gql schema**
- **replace fetch_content with fetch_content_type as a enum: ALWAYS**
- **add representation column to the highlight table and gql schema which is either CONTENT or FEED_CONTENT**

### DIFF
--- a/packages/api/src/generated/graphql.ts
+++ b/packages/api/src/generated/graphql.ts
@@ -368,6 +368,7 @@ export type CreateHighlightInput = {
   patch?: InputMaybe<Scalars['String']>;
   prefix?: InputMaybe<Scalars['String']>;
   quote?: InputMaybe<Scalars['String']>;
+  representation?: InputMaybe<RepresentationType>;
   sharedAt?: InputMaybe<Scalars['Date']>;
   shortId: Scalars['String'];
   suffix?: InputMaybe<Scalars['String']>;
@@ -1024,6 +1025,7 @@ export type Highlight = {
   quote?: Maybe<Scalars['String']>;
   reactions: Array<Reaction>;
   replies: Array<HighlightReply>;
+  representation: RepresentationType;
   sharedAt?: Maybe<Scalars['Date']>;
   shortId: Scalars['String'];
   suffix?: Maybe<Scalars['String']>;
@@ -1281,6 +1283,7 @@ export type MergeHighlightInput = {
   patch: Scalars['String'];
   prefix?: InputMaybe<Scalars['String']>;
   quote: Scalars['String'];
+  representation?: InputMaybe<RepresentationType>;
   shortId: Scalars['ID'];
   suffix?: InputMaybe<Scalars['String']>;
 };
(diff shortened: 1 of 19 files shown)


packages/api/src/generated/graphql.ts
```diff
@@ -1017,6 +1025,7 @@ export type Highlight = {
   quote?: Maybe<Scalars['String']>;
   reactions: Array<Reaction>;
   replies: Array<HighlightReply>;
+  representation: RepresentationType;
```
jacksonh, line 1028: I'm starting to wonder if representation is the best word now. It makes sense in the context of "available representations" but do you think there is a better term?

packages/api/src/generated/schema.graphql
```diff
@@ -905,6 +913,7 @@ type Highlight {
   quote: String
   reactions: [Reaction!]!
   replies: [HighlightReply!]!
+  representation: RepresentationType!
```
jacksonh, line 916: Can we make it so it defaults to content and isn't required? I think that would make it easier for all the clients

packages/api/src/jobs/rss/refreshFeed.ts
```diff
@@ -272,8 +278,12 @@ const createTask = async (
     return true
   }
 
-  if (folder === 'following' && !fetchContent) {
-    return createItemWithPreviewContent(userId, feedUrl, item)
+  const feedContent = item.content || item.contentSnippet || item.summary
```
jacksonh, line 281: nice.

packages/api/src/resolvers/highlight/index.ts
```diff
@@ -51,6 +52,7 @@ export const createHighlightResolver = authorized<
         highlightType: input.type || HighlightType.Highlight,
         highlightPositionAnchorIndex: input.highlightPositionAnchorIndex || 0,
         highlightPositionPercent: input.highlightPositionPercent || 0,
+        representation: input.representation || RepresentationType.Content,
```
jacksonh, line 55: Yeah i think we will have to make this nullable and just default to the reader content representation otherwise all clients will need to be updated before we deploy this.

packages/api/src/generated/graphql.ts
sywhb: could we use `source`?

packages/api/src/generated/schema.graphql
sywhb: Correct me if I'm wrong but I think clients will only fail if is is a required argument/parameter in the graphql request This is a required field in the response which I think should be fine

packages/api/src/resolvers/highlight/index.ts
sywhb: The `input.representation` is nullable and defaults to `Content`

packages/api/src/generated/schema.graphql
sywhb: You can test it locally

jacksonh: Oh got it, thank you!

packages/api/src/resolvers/highlight/index.ts
jacksonh: thanks!

jacksonh (approved):

packages/api/src/generated/graphql.ts
sywhb: @jacksonh

jacksonh (approved):

PR #3571, opened 2024-02-24
Unify wording of removing articles

henlium:
Most of the places use "remove", but "delete" is used in this dropdown menu, which seems inconsistent.


jacksonh:
hah thanks, I just added this in another place last week and had trouble remembering which is used.

PR #3580, opened 2024-02-26
replace segment with posthog

sywhb:


packages/api/src/resolvers/article/index.ts
```diff
@@ -491,7 +491,7 @@ export const getArticleResolver = authorized<
 //         source: 'resolver',
 //         resolver: 'setShareArticleResolver',
 //         articleId: article.id,
-//         userId: uid,
+//         distinctId: uid,
```
jacksonh, line 494: 😬

jacksonh (approved):
Thanks, was there any initial analytics init code that needs to be changed, like the old identify call? I think thats all done when user requests labels now though so probably not an issue.

packages/api/src/resolvers/article/index.ts
sywhb: 😆

sywhb:
> Thanks, was there any initial analytics init code that needs to be changed, like the old identify call? I think thats all done when user requests labels now though so probably not an issue.

Yeah, it was set in the labels resolver but I think we could change it to `set_once` which will once set the user properties are not there.

```
      analytics.capture({
        distinctId: uid,
        event: 'labels',
        properties: {
          env: env.server.apiEnv,
          $set: {
            email: user.email,
            username: user.profile.username,
          },
        },
      })
```

Since Email and Username should not be changed, so I think they are safe to "set once" only.

PR #3593, opened 2024-02-28
Fix bulk action on large library causing server to crash

sywhb:
- **searchLibraryItem**: returns only count if size is 0 because limit 0 is not allowed in typeorm


jacksonh (approved):
LGTM. I guess because they have their own take/skip stuff they don't allow it

PR #3598, opened 2024-02-29
fix search results for archived items which state is not archived

sywhb:


jacksonh (approved):
Assumming tests pass lets try this out in demo

PR #3618, opened 2024-03-04
Allow web extension to get the content text direction and save

sywhb:


jacksonh (approved):
Want to test this one a bit to see if `dir` can be null/undefined.

PR #3620, opened 2024-03-04
Add authorization with notion

sywhb:


packages/api/src/util.ts
```diff
@@ -165,6 +170,9 @@ const nullableEnvVars = [
   'MQ_REDIS_CERT',
   'IMPORTER_METRICS_COLLECTOR_URL',
   'INTERNAL_API_URL',
+  'NOTION_CLIENT_ID',
```
jacksonh, line 173: Do we need to do something to sign up for official credentials?

jacksonh (approved):
wow nice work

PR #3637, opened 2024-03-06
Refer to "Omnivore for iOS" as "Omnivore for iOS and macOS"

jeffbowen:
Since the Omnivore iOS app seems to also be installable on macOS, it should be referred to as such.

Note: It's also explicitly referred to as "Omnivore for iOS" in its App Store description.

### DIFF
--- a/packages/web/components/elements/MobileInstallHelp.tsx
+++ b/packages/web/components/elements/MobileInstallHelp.tsx
@@ -144,7 +144,7 @@ export default function MobileInstallHelp({
         }}
       >
         With the Omnivore for iOS and macOS app installed you can save any link
-        using our share extension.
+        using our Share extension.
         <br />
         {!onboarding && (
           <StyledAnchor
--- a/packages/web/components/elements/MobileInstallHelp.tsx
+++ b/packages/web/components/elements/MobileInstallHelp.tsx
@@ -118,7 +118,7 @@ export default function MobileInstallHelp({
             },
           }}
         >
-          Install Omnivore for iOS
+          Install Omnivore for iOS and macOS
         </StyledText>
       </Box>
       <StyledText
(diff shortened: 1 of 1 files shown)


jacksonh:
Can you show how you installed the iOS app on Macos?

jeffbowen:
Just click the App Store button on the [Install page](https://omnivore.app/settings/installation) to go to the [App Store listing](https://apps.apple.com/us/app/omnivore-read-it-later/id1564031042). The app appears to be compatible with iOS, iPadOS, and visionOS. Installed on macOS for me just fine.

If the visionOS compatibility is intentional, perhaps we need to think of another way to structure this section? Maybe instead of referring to the OS, refer to the device? This seems to be how Cultured Code [does it](https://culturedcode.com/things/) for Things (first example I could think of) — Mac, iPhone, iPad, Vision Pro. WhatApps [refers to](https://www.whatsapp.com/download) — Mac, iOS, Android. Could explore more examples.

jacksonh:
I think the changes still make sense though, its the same appstore link so doesn't really matter to the user. I was wondering if you somehow used the Catalyst stuff.

jeffbowen:
image

jacksonh:
> image

yeah thats the mac app. Uses some shared code but is a different app

jeffbowen:
Gotcha

jacksonh:
Thanks for the contribution!

jeffbowen:
Of course! Just found Omnivore today after being frustrated that highlighting more than three things in an article in Matter required a paid subscription. Felt like there should be an open-source alternative and was pleasantly surprised to discover Omnivore. Look forward to getting to know it better.

jacksonh:
> Of course! Just found Omnivore today after being frustrated that highlighting more than three things in an article in Matter required a paid subscription. Felt like there should be an open-source alternative and was pleasantly surprised to discover Omnivore. Look forward to getting to know it better.

nice. I do recommend the web app over the mac. There's a lot more there.

PR #3649, opened 2024-03-07
Discover clean

Podginator:


packages/pdf-handler/tsconfig.json
```diff
@@ -1,9 +1,14 @@
 {
-  "extends": "@tsconfig/node14/tsconfig.json",
+  "extends": "../../tsconfig.json",
```
jacksonh, line 2: I think we need to remove all these, as the cloud run handlers can't access things outside their directory.

Podginator: Reverted all the ones that were @tsconfig on Main so that they don't access things outside their dir

PR #3660, opened 2024-03-12
do not get content from db when exporting item to integrations

sywhb:


packages/api/src/services/library_item.ts
```diff
@@ -700,10 +700,16 @@ export const findRecentLibraryItems = async (
 }
 
 export const findLibraryItemsByIds = async (ids: string[], userId: string) => {
+  const selectColumns = getColumns(libraryItemRepository)
```
jacksonh, line 703: Just wondering, would this get called by search or get article APIs?

sywhb: No, it only gets called here

jacksonh (approved):

PR #3661, opened 2024-03-12
Pull metadata from YouTube for library items

jacksonh:
- YouTube job handler
- Pull duration and description from YouTube metadata


packages/api/src/pubsub.ts
```diff
@@ -89,7 +103,17 @@ export const createPubSubClient = (): PubsubClient => {
...
 
       if (await findFeatureByName(FeatureName.AISummaries, userId)) {
-        await enqueueAISummarizeJob({
+        // await enqueueAISummarizeJob({
+        //   userId,
+        //   libraryItemId,
+        // })
+      }
+
+      if (
```
jacksonh, line 112: @sywhb this is the other issue we kind of have with better job handling. Might need a new way of adding jobs based on library item.

sywhb (approved):
👍 👍 👍

PR #3653, opened 2024-03-08
iOS read position and view hierarchy fixes

jacksonh:
- Properly observe model items so library position can be updated
- Dont show footer during loading
- Fix issues with search view hierarchy

Issue #3648, opened 2024-03-07
After unsubscribing from a subscription on web, can't unscribe from more until page refresh


WillTwait:
repro: 
- unsub from a subscription in `/settings/subscriptions`
- can no longer click `...` button to unsub from more
- refresh page, and buttons are now clickable

jacksonh:
I suspect this is related to a bug we've seen in our modals library, thanks for reporting.

jacksonh:
Deploying a fix for this one now.

PR #3716, opened 2024-03-23
fix: page becomes unresponsive after closing modal

egrajeda:
Came across this issue while migrating to Omnivore, but I saw it also reported in issue #3648. There was a first attempt to fix in #3653, but I believe it's an issue in the library itself, which is not there anymore in 2.0.6.

It also fixes the "too much recursion" error that shows in the console when the modal is open, both in the Feeds page and in the Inbox (e.g. when clicking on Edit Info).

### DIFF
--- a/packages/web/components/patterns/ConfirmationModal.tsx
+++ b/packages/web/components/patterns/ConfirmationModal.tsx
@@ -6,7 +6,6 @@ import {
 import { VStack, HStack } from '../elements/LayoutPrimitives'
 import { Button } from '../elements/Button'
 import { StyledText } from '../elements/StyledText'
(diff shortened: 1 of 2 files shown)


jacksonh:
Ah awesome. I have had a few issues when upgrading before (menus stopped working altogether), but hopefully this one works. Going to do some local testing.

egrajeda:
Great, I did some manual tests in the inbox and some settings page and it seemed to work. If you find a problem elsewhere do let me know and I can take a look.

jacksonh:
Working pretty well for me locally. Going to merge and do some more testing in dev.

PR #3741, opened 2024-03-27
Vertically align the flair icon with the text

egrajeda:
It's a real minor thing but it's an easy fix. I tried to explain the misalignment and the alignment in this picture:

image

The fix itself is short, but I did a small refactoring to not have to repeat the CSS in different places.


jacksonh (approved):

jacksonh:
Thank you!

PR #3746, opened 2024-03-28
cache me api response in redis

sywhb:


jacksonh:
I think we also might need to invalidate the cache when the user features change

sywhb:
> I think we also might need to invalidate the cache when the user features change

Yeah that is a good one

sywhb:
> I think we also might need to invalidate the cache when the user features change

Actually features are not cached because they are not queried from the function_resolver

packages/api/src/resolvers/function_resolvers.ts
```diff
@@ -377,7 +381,14 @@ export const functionResolvers = {
         return undefined
       }
 
-      return findUserFeatures(ctx.claims.uid)
+      let features = await getCachedFeatures(ctx.claims.uid)
+      if (!features) {
+        features = await findUserFeatures(ctx.claims.uid)
+
+        await cacheFeatures(ctx.claims.uid, features)
```
pradeepmurugesan, line 388: Why do we need to `await` for the `caching` to be completed? Can't this happen in the asynchronously ?

jacksonh: generally we await on all our backend functions, one reason is it can better handle errors.

jacksonh:
I will close this one as I think it didn't quite handle some of the edge cases we came up with.

PR #3748, opened 2024-03-29
depreciate features fields in get user api in web

sywhb:


packages/web/pages/settings/account.tsx
```diff
@@ -441,19 +442,23 @@ export default function Account(): JSX.Element {
             <StyledLabel>Beta features</StyledLabel>
             {!isValidating && (
               <>
-                {viewerData?.me?.features.map((feature) => {
+                {viewerData?.me?.featureList.map((feature) => {
```
jacksonh, line 445: Does featurelist only include a feature if its granted?

sywhb: No, it includes all the features

jacksonh:
Ah sorry didn't realize `userHasFeature` was already updated, now this all makes sense!

jacksonh (approved):

PR #3759, opened 2024-03-31
store the rate limit data in redis if redis is connected

sywhb:


jacksonh:
This is great, there is also `hourlyLimiter` in `auth_router` but you might want to add that one in a separate PR just to ensure it doesn't conflict with the recaptcha PR.

jacksonh (approved):

PR #3763, opened 2024-04-01
use redis for hourly rate limiter

sywhb:


jacksonh:
I think we might need to set a prefix for the redis stores for each of these limiters or they could conflict. I haven't found exact docs yet but i see this:

```
prefix
The text to prepend to the key in Redict/Redis.

Defaults to rl:.
```

jacksonh (approved):
Ah very clean way of doing this

PR #3773, opened 2024-04-02
Start to add explain feature on iOS

jacksonh:
- Add explain backend components
- Clean up imports
- Use the theme colour for the lightbox background
- Add migrations for ai tasks
- Create a new ai task that is run as a job and its status can be polled for
- Update remove ai-summary and replace with more generic ai task


packages/api/src/entity/ai_tasks.ts
```diff
@@ -0,0 +1,79 @@
...
+  name!: string
+
+  @Column({ type: 'text', name: 'display_text', nullable: false })
+  displayText!: string
+
+  @Column({ type: 'text', name: 'template', nullable: false })
+  template!: string
+
+  @Column('text', { array: true, nullable: true })
+  variables!: string
```
sywhb, line 28: Should this be an `string[]`?

packages/db/migrations/0170.do.create_ai_tasks_table.sql
```diff
@@ -0,0 +1,74 @@
...
+  FOR INSERT TO omnivore_user
+  WITH CHECK (true);
+
+CREATE POLICY update_ai_task_results on omnivore.ai_task_results
+  FOR UPDATE TO omnivore_user
+  WITH CHECK (true);
+
+GRANT SELECT, INSERT, UPDATE ON omnivore.ai_task_results TO omnivore_user;
+
+GRANT SELECT ON omnivore.ai_prompts TO omnivore_user;
```
sywhb, line 64: Why we do not create RLS for `ai_prompts`?

packages/api/src/jobs/ai-summarize.ts
```diff
@@ -67,21 +80,18 @@ export const aiSummarize = async (jobData: AISummarizeJobData) => {
...
       async (t) => {
-        return t.getRepository(AISummary).save({
+        return t.getRepository(AITaskResult).save({
+          request: aiTaskRequest,
           user: { id: jobData.userId },
-          libraryItem: { id: jobData.libraryItemId },
-          title: libraryItem.title,
-          slug: libraryItem.slug,
-          summary: summary,
+          libraryItem: aiTaskRequest?.libraryItem,
```
sywhb, line 93: I guess there is no need to put `?` here

packages/api/src/routers/ai_summary_router.ts
```diff
@@ -6,50 +6,139 @@ import express from 'express'
...
+    const extraText = ''
+
+    const aiTaskResult = await authTrx(
+      async (t) => {
+        // Check to see if there is already a result
+        return await t.getRepository(AITaskResult).findOne({
+          where: {
+            user: { id: uid },
+            libraryItem: { id: libraryItemId },
+            request: { promptName, extraText },
```
sywhb, line 90: Do we need `extraText` here since `promptName` is unique

packages/api/src/routers/ai_summary_router.ts
```diff
@@ -6,50 +6,139 @@ import express from 'express'
...
-        summary: result?.summary,
+    // TODO: do a proper find or create, just easier to use typeorm for now
+    const aiTaskRequest = await authTrx(
+      async (t) => {
+        const current = await t.getRepository(AITaskRequest).findOne({
+          where: {
+            user: { id: uid },
+            libraryItem: { id: libraryItemId },
+            prompt: { name: promptName },
+            extraText,
```
sywhb, line 113: Same here

packages/api/src/routers/ai_task_router.ts
```diff
@@ -0,0 +1,184 @@
...
+    // -- return the request task ID for any of the above
+    // -- client can then query for the result based on the task ID
+    const existingTaskRequest = await authTrx(
+      async (t) => {
+        return await t.getRepository(AITaskRequest).findOne({
+          where: {
+            user: { id: uid },
+            libraryItem: { id: libraryItemId },
+            prompt: { name: promptName },
+            extraText,
```
sywhb, line 50: Same here

sywhb (approved):
LGTM.

Just some notes:
- I know this is a simple version but we probably should think about race conditions in the future. 
- Probably we can wrap the Database queries in try-catch block in the routers so it won't throw error to the user
- I wonder if we could have a status column in the `ai_result` table and instead of deleting it so to avoid multiple deletion of the same result

packages/api/src/jobs/ai-summarize.ts
jacksonh: A library item could be optional, because we might have some tasks that don't require them, maybe something along the lines of "define this word"

packages/db/migrations/0170.do.create_ai_tasks_table.sql
jacksonh: They are just a global table that all users can use.

jacksonh:
Closing this one.

PR #3779, opened 2024-04-03
Also use mailjet for reset password emails

jacksonh:


packages/api/src/services/send_emails.ts
```diff
@@ -104,6 +104,10 @@ export const sendPasswordResetEmail = async (user: {
     link,
   }
 
+  if (process.env.USE_MAILJET) {
+    return sendWithMailJet(user.email, link)
```
sywhb, line 108: Should we change the subject and title accordingly?

PR #3791, opened 2024-04-05
fix/labels

sywhb:
- **do not create label if label name case-insensitively matches existing label name**
- **allow delete internal labels**
- **deduplicate labels when there are case-insensitive identical names**
- **show label_attached instead of label_created in rules**

### DIFF
--- a/packages/api/src/services/labels.ts
+++ b/packages/api/src/services/labels.ts
@@ -148,6 +148,7 @@ export const saveLabelsInLibraryItem = async (
       {
         id: libraryItemId,
         labels: labels.map((l) => deepDelete(l, columnsToDelete)),
+        labelNames: labels.map((l) => l.name),
       },
       userId
     )
--- a/packages/api/src/services/labels.ts
+++ b/packages/api/src/services/labels.ts
@@ -41,21 +41,18 @@ export const findOrCreateLabels = async (
   labels: CreateLabelInput[],
   userId: string
 ): Promise<Label[]> => {
-  // create labels if not exist
-  await authTrx(
-    async (tx) =>
-      tx.withRepository(labelRepository).createLabels(labels, userId),
-    undefined,
-    userId
-  )
-
-  // find labels
   return authTrx(
-    async (tx) =>
-      tx.withRepository(labelRepository).findByNames(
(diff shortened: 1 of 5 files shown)


packages/api/src/services/labels.ts
```diff
@@ -41,21 +41,18 @@ export const findOrCreateLabels = async (
...
-        labels.map((l) => l.name),
-        userId
-      ),
+    async (tx) => {
+      const repo = tx.withRepository(labelRepository)
+      // create labels if not exist
+      await repo.createLabels(labels, userId)
+
+      // find labels by names
+      return repo.findBy({
```
jacksonh, line 51: is this part case insensitive?

packages/api/test/resolvers/labels.test.ts
```diff
@@ -160,17 +150,30 @@ describe('Labels API', () => {
 
       before(async () => {
         existingLabel = await createLabel('label3', '#ffffff', user.id)
```
jacksonh, line 152: Maybe we should create this one as `LABEL3` to test the case sensitivity

packages/api/src/services/labels.ts
sywhb: No, this part is case sensitive so we ensure only one RSS label would be matched if there is another `rss` label created by users

packages/api/test/resolvers/labels.test.ts
```diff
@@ -160,17 +150,30 @@ describe('Labels API', () => {
...
+        }).expect(200)
+
+        expect(res.body.data.createLabel.errorCodes).to.eql([
+          'LABEL_ALREADY_EXISTS',
+        ])
+      })
 
-        expect(res.body.data.createLabel.errorCodes).to.eql(['BAD_REQUEST'])
+      it('returns error code LABEL_ALREADY_EXISTS if case-insensitive label name exists', async () => {
+        const name = existingLabel.name.toUpperCase()
```
sywhb, line 170: > Maybe we should create this one as `LABEL3` to test the case sensitivity

It's here 😆

sywhb (commented):
Comments

jacksonh (approved):

PR #3792, opened 2024-04-05
show failed_at timestamp if a feed failed to fresh

sywhb:


packages/web/pages/settings/feeds/index.tsx
```diff
@@ -229,10 +229,17 @@ export default function Rss(): JSX.Element {
                   }}
                 >
                   <SpanBox>{`URL: ${subscription.url}`}</SpanBox>
-                  <SpanBox>{`Last refreshed: ${
-                    subscription.lastFetchedAt
-                      ? formattedDateTime(subscription.lastFetchedAt)
-                      : 'Never'
+                  {/* show failed timestamp instead of last refreshed timestamp if the feed failed to refresh */}
+                  <SpanBox>{`${
```
jacksonh, line 233: one small tweak I'd make here is we could set the color to read if failedAt is displayed. Probably something like this:

```
<SpanBox css={{ color: failedAt ? 'red': undefined }}>
```

sywhb: Screenshot 2024-04-08 at 4 32 49 PM

Looks good to me

sywhb: Comparison:

Screenshot 2024-04-08 at 4 33 42 PM

PR #3796, opened 2024-04-08
feature: add captcha to email-login and reset-password pages

sywhb:
- **move Recaptcha to element**
- **add recaptcha to the email-login page**
- **add recaptcha to the reset-password page**
- **verify captcha in email-login and forgot-password api**

### DIFF
--- a/packages/api/src/routers/auth/auth_router.ts
+++ b/packages/api/src/routers/auth/auth_router.ts
@@ -416,6 +416,7 @@ export function authRouter() {
       interface LoginRequest {
         email: string
         password: string
+        recaptchaToken?: string
       }
       function isValidLoginRequest(obj: any): obj is LoginRequest {
         return (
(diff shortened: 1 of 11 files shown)


jacksonh:
On the login page the recaptcha component gets reloaded during typing which causes it to flash. I think this is because the form and the recaptcha are in the same parent element. In EmailSignUp.tsx you can see how I created `<SignUpForm />` to move most of the form elements into a separate component.

sywhb:
@jacksonh I have made the change to use a higher level component for the `LoginForm` and `ForgotPasswordForm` so the states are not shared by the `Recaptcha` component.

Could you please let me know if it is okay?

jacksonh:
I suspect there is an issue with how we are including the recaptcha library. Sometimes when i click on the Forgot Password buytton i get this error

Screenshot 2024-04-13 at 09 45 28

jacksonh:
I played around with it and it seems like if we don't have the provider in `_app` (i even tried it in a shared layout for all the auth pages), it seems to throw this exception. 

I'm wondering if we can use one of the other methods of including it. Seems like in a perfect world it wouldn't be on all pages.


```
    <GoogleReCaptchaProvider
      type="v2-checkbox"
      isEnterprise={true}
      siteKey={process.env.NEXT_PUBLIC_RECAPTCHA_CHALLENGE_SITE_KEY ?? ''}
    >
      <KBarProvider actions={generateActions(router)}>
        <KBarPortal>
          <KBarPositioner style={{ zIndex: 100 }}>
            <KBarAnimator style={animatorStyle}>
              <KBarSearch style={searchStyle} />
              <KBarResultsComponents />
            </KBarAnimator>
          </KBarPositioner>
        </KBarPortal>
        <IdProvider>
          <Component {...pageProps} />
        </IdProvider>
      </KBarProvider>
    </GoogleReCaptchaProvider>
```

jacksonh:
Does seem that based on docs it has to be that way though:

```
GoogleReCaptchaProvider's responsibility is to load the necessary reCaptcha script and provide access to reCaptcha to the rest of your application.

Usually, your application only needs one provider. You should place it as high as possible in your React tree. It's to make sure you only have one instance of Google Recaptcha per page, and it doesn't reload unnecessarily when your components re-rendered.

Same thing applied when you use this library with framework such as Next.js or React Router and only want to include the script on a single page. Try to make sure you only have one instance of the provider on a React tree and to place it as high (on the tree) as possible.
```

jacksonh:
Still needs more testing.

PR #3812, opened 2024-04-12
feature/digest api

sywhb:
- **stub create digest job**
- **create digest job api**
- **get digest api**
- **send feedback api**
- **use another queue for digest job**


jacksonh (approved):

donaldguy:
CleanShot 2024-04-20 at 14 14 38@2x

It seems like some files have gone missing? (unless they are supposed to autogen somehow in which case config is missing?)

At least:
- `DigestViewModel`
- `DigestView`

(and when I backed up before 47150c29913ecece27a755509815eecc8642bd17, `ExplainView` but it seems like that got factored out?

Built okay at e30fa13a2e5ef4581827e0bb0623f75d1c8b81a0 )

PR #3815, opened 2024-04-14
feat: bulk mark as read action

kimlimjustin:
Added mark all as read action button, as opposed to the current ‘select articles to mark as read’ button.


jacksonh:
Thanks. Can i ask why you prefer to mark as read vs archive?

kimlimjustin:
> Thanks. Can i ask why you prefer to mark as read vs archive?

So that it still appears in the inbox tab, but not in the continue reading; sometimes there are some important articles that I want to see again in the future, but definitely not in the continue reading tab

jacksonh:
> > Thanks. Can i ask why you prefer to mark as read vs archive?
> 
> So that it still appears in the inbox tab, but not in the continue reading; sometimes there are some important articles that I want to see again in the future, but definitely not in the continue reading tab

Got it, thank you for explaining!

Issue #3820, opened 2024-04-15
LABEL_ALREADY_EXISTS issue


clementbrizard:
Hi,

First, thank you for your job :)

I am often unable to create a label. When I check the API calls, I see the following answer:

```json
{
  "data": {
    "createLabel": {
      "errorCodes": [
        "LABEL_ALREADY_EXISTS"
      ]
    }
  }
}
```

Whereas the label does not exist, at least in my labels.

Thank you in advance :)

jacksonh:
Hi @clementbrizard can i ask when this occurs for you? You are using curl or you are doing it in the UI and it occurs?

sywhb:
Hey @clementbrizard , this should be fixed now :)

Sorry for the inconvenience caused

sywhb:
Nice. I will close this issue as completed ;)

PR #3822, opened 2024-04-16
fix: LABEL_ALREADY_EXIST bug occured during label creation

sywhb:
* comparing both user_id and label name when looking for an existing matching label


sywhb:
Fix #3820

jacksonh (approved):

jacksonh (approved):

PR #3823, opened 2024-04-16
add features back to the schema to be compatible with browser cache

sywhb:


jacksonh:
I guess we can close this one. Feel free to re-open if I was wrong.

PR #3826, opened 2024-04-17
feat: increase daily upload limit

kimlimjustin:
I think the current static daily upload limit of 1 is kinda low, which can be increased dynamically according to the env variable value.


jacksonh:
hah, we did put this in when the uploader could seriously effect performance of the system. Its a bit better now, but probably what we should do is combine this with either a fix to not have to fetch all content for imports, or to track and limit the daily number of library items imported

jacksonh:
I dropped the default to `5`. Really what we need to do is enforce the total daily items.

Issue #2870, opened 2023-10-06
OPML import/export for RSS feeds.


Southpaw1496:
[OPML](https://en.wikipedia.org/wiki/OPML) is an XML format that is commonly used by RSS readers to exchange lists of feeds. If Omnivore had support for OPML importing, it would allow users to migrate an existing list of RSS/Atom feeds into Omnivore without having to do them all manually. An export feature is less critical, but it would give users peace of mind that they can retrieve their feeds and migrate to another application at any time, making them more likely to try Omnivore over their existing solution.

cocoonkid:
sorry, bumping this :-) I know that's not helpful but imho Omnivore without this basic feature is neither.

Benbinbin:
In my opinion, it would be great if Omnivore could support both importing and exporting OPML files at the same time. This way, Omnivore would be seen as more welcoming and accommodating, without any intention of luring users onto the platform. We understand how important it is for users to have control over their data, so we ensure that they can easily back up their OPML files whenever they need to.

Jolg42:
I really like Omnivore but I'm not ready to switch without this feature honestly.

After a quick look, I found the following
- the current add feed page: https://github.com/omnivore-app/omnivore/blob/main/packages/web/pages/settings/feeds/add.tsx uses `subscribeMutation()`
- `subscribeMutation()` https://github.com/omnivore-app/omnivore/blob/main/packages/web/lib/networking/mutations/subscribeMutation.ts#L30
- its resolver is here https://github.com/omnivore-app/omnivore/blob/main/packages/api/src/resolvers/subscriptions/index.ts#L174

Along the way I found this important detail, the maximum number of feeds is hard-coded, I can understand why it exists though it would be better to have this configurable for self-hosting I think.
https://github.com/omnivore-app/omnivore/blob/b464734cc9c3bdff4f605f6674566001bb3ef7ba/packages/api/src/resolvers/subscriptions/index.ts#L223-L224C1

Jolg42:
Additionally
- Here is the spec for OPML: http://opml.org/spec2.opml
- This looks like an interesting implementation of an import & export https://github.com/chhoumann/PodNotes/blob/83f3dd96e0f7b071baf2e7f9796f6ea52c4650e9/src/opml.ts#L63

cori:
[...]

PR #3836, opened 2024-04-18
feat: import opml handler

kimlimjustin:
Created OPML import handler, as discussed on #2870, by integrating https://github.com/danmactough/node-opmlparser solution into the `packages/import-handler` directory.

Yet, please note that due to insufficient up-to-date contributing docs and guidelines, this PR is made by literally reading the whole codebase manually, without actually running the application locally; this PR is only testified by running the mocha testing under the `import-handler` package. 

That being said, please try this PR and please let me know if any errors occurred or if any changes should be made further to complete this pull request,

Thank you and hope that the OPML handler can be part of omnivore soon shortly!

### DIFF
--- a/android/Omnivore/app/src/main/graphql/schema.graphqls
+++ b/android/Omnivore/app/src/main/graphql/schema.graphqls
@@ -2702,6 +2702,7 @@ enum UploadImportFileType {
   MATTER
   POCKET
   URL_LIST
+  OPML
 }
 
 type User {
--- a/apple/OmnivoreKit/Sources/Services/DataService/GQLSchema.swift
+++ b/apple/OmnivoreKit/Sources/Services/DataService/GQLSchema.swift
@@ -38617,6 +38617,8 @@ extension Enums {
         case pocket = "POCKET"
 
         case urlList = "URL_LIST"
+
+        case opml = "OPML"
     }
 }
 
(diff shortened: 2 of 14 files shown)


kallekn:
This is a deal breaker. Omnivore looks nice, but I need to be able to import my feeds in order to use it.

natarslan:
Following this! Would be great to bulk import feeds (opml).

cori:
> I don't think our handling of lots of RSS feeds is great yet

What's the internal perception of "lots" here? My OPML file has ~5 dozen feeds. Is that a lot in this context?

axgdcode:
> Hey this is an interesting way of doing it. We actually do have an OPML API also but we haven't exposed it on the web, just because I don't think our handling of lots of RSS feeds is great yet.
In anticipation of the self-hosted mode, I believe this feature should be enabled exclusively for that mode.

jacksonh:
> > Hey this is an interesting way of doing it. We actually do have an OPML API also but we haven't exposed it on the web, just because I don't think our handling of lots of RSS feeds is great yet.
> 
> In anticipation of the self-hosted mode, I believe this feature should be enabled exclusively for that mode. @jacksonh

HI, yes you could call this API on your self hosted version now. Just use the `scanFeeds` API.

PR #3838, opened 2024-04-19
do not upload/download audio file in gcs

sywhb:


sywhb:
> hey Honbgo, do you think this will reduce our cache hits?

Yeah it does not work well and I will close it for now

PR #3839, opened 2024-04-19
feature/upload original content

sywhb:
- **fix failed to get thumbnail from rss feed**
- **reduce get text nodes timeout to 60 seconds to avoid search api timeout**
- **upload and download original content from GCS**


packages/api/src/jobs/process-youtube-video.ts
```diff
@@ -382,7 +378,7 @@ export const processYouTubeVideo = async (
         // enqueue a job to process the full transcript
         const updatedContent = await addTranscriptPlaceholdReadableContent(
           libraryItem.originalUrl,
-          libraryItem.originalContent
+          libraryItem.readableContent
```
jacksonh, line 381: hmm, surprised this worked before

jacksonh (approved):

PR #3846, opened 2024-04-22
fix: export all items job stuck

sywhb:


jacksonh:
Seems like tests hanging with this one.

sywhb:
> Seems like tests hanging with this one.

I removed the `setTimeout` thing as it is not working anyways

PR #3863, opened 2024-04-25
increase rss parser timeout to 20 seconds

sywhb:


jacksonh:
A little worried about this one, but lets see.

PR #3881, opened 2024-05-01
feature/onboarding

sywhb:


jacksonh (approved):

packages/api/src/entity/library_item.ts
```diff
@@ -138,9 +138,9 @@ export class LibraryItem {
   @Column('text')
   readableContent!: string
 
-  // NOT SUPPORTED IN TYPEORM
-  // @Column('vector', { nullable: true })
-  // embedding?: number[]
+  // typeorm does not support vector type, so we store it as a string
+  @Column('text')
+  embedding?: string
```
jacksonh, line 143: Do you think we should use a separate table for these? I know we talked about it with some other attributes.

packages/api/src/entity/library_item.ts
```diff
@@ -204,4 +204,7 @@ export class LibraryItem {
 
   @Column('text')
   highlightAnnotations?: string[]
+
+  @Column('timestamptz')
+  sharedAt?: Date
```
jacksonh, line 209: Thinking out loud, but also wonder if this could be `is_discover_item` not as flexible for future use, but maybe more explicit.

packages/api/src/resolvers/article/index.ts
```diff
@@ -399,13 +399,12 @@ export const getArticleResolver = authorized<
...
           'recommendations_recommender'
         )
-        .where('libraryItem.user_id = :uid', { uid })
 
       // We allow the backend to use the ID instead of a slug to fetch the article
       // query against id if slug is a uuid
       slug.match(/^[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}$/i)
-        ? qb.andWhere('libraryItem.id = :id', { id: slug })
-        : qb.andWhere('libraryItem.slug = :slug', { slug })
+        ? qb.where('libraryItem.id = :id', { id: slug })
```
jacksonh, line 406: What does this change do?

jacksonh: > What does this change do?

oh i see, we are removing the uid check above. Its a little scary to fully rely on RLS for that.

packages/api/src/entity/library_item.ts
sywhb: I think this and above are similar question: _Should we have a join table for things like `discover_item` and `item_embedding`_.

There are several considerations:
1. search performance for inner joins:
Probably not so bad for one-to-one relations like `discover_item` and `item_embedding`. I think we can benchmark the query.

2. frequently update column: I agree with your thought and it has been a big concern for us. Worth to move these columns to a separate table.

3. optimization:
Instead of running cosine similarity query on the fly, we could pre-generate the topics or some other features and store them in another table.

4. naming: 
I think if we want to be more explicit, we should better go with a separate table `discover_item` which links to the `libraryItem` table for normalization and reducing the size of the table.

Having considered all these concerns, I think we should
1. Test and benchmark the search query against join tables (probably only one-to-one relations) in cloned DB
2. If (1) result is not bad, we could create two additional separate tables: `discover_item` and `item_embedding`
3. Create another job to generate topic when the `embedding` is updated and stored in the table

packages/api/src/resolvers/article/index.ts
sywhb: The motive is to use the `slug` index to improve query performance. I think RLS should be quite reliable but alternatively we could update the index to use both `user_id` and `slug` columns too

packages/api/src/entity/library_item.ts
Podginator: Hi @sywhb I also mentioned this on Slack, but a lot of what you've talked about in here is already done as part of my Discover update. 

The optimisation part is already done as part of the import in discover, see: 
https://github.com/omnivore-app/omnivore/blob/main/packages/db/migrations/0168.do.add_discover_feeds.sql

The discover feeds contain already the topics and embeddings. There's also a separate table for discover items, which have their embeddings stored, whether a user has saved a story - which is used to calculate the popularity score in discover too. 

The calculations of the embeddings is done via https://github.com/omnivore-app/omnivore/blob/main/packages/discover/src/index.ts - This deliberately slows down the embedding process so that you never exceeds the RPS that the API supports.

sywhb: Thank you @Podginator ! Taking a look now

PR #3891, opened 2024-05-02
select model from prompt and defaults to openai

sywhb:
- **select model from prompt and defaults to openai**
- **allow random model**
- **upload summaries to GCS**


jacksonh:
love it

jacksonh (approved):

PR #3894, opened 2024-05-03
feat/Android: Add plus and minus buttons to sliders on reader preference view

jivesh:
Adds + and - buttons on the sliders for font size, margin, and line spacing. Most times when I want to change these reader preferences, my changes are more likely to be smaller changes. But, smaller changes are challenging to do precisely with a slider (rip my fat fingers). 

Screenshot_20240504_014142

I defaulted to adding new Composable UI components for the plus button, minus button, and for the slider integrated with the plus and minus buttons. Let me know if that's not the preferred way to add UI components, or if they should be named more specific to the Reader Preferences context! My suspicion is that the AddButton and SubtractButton probably are not useful as general components

### DIFF
--- a/android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/components/AddButton.kt
+++ b/android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/components/AddButton.kt
@@ -1,0 +1,45 @@
+package app.omnivore.omnivore.feature.components
+
+import androidx.compose.foundation.interaction.MutableInteractionSource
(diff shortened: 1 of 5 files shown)


jivesh:
Screenshot 2024-05-04 at 1 48 48 AM

The web version of this looks a LOT nicer wow, I do not know or trust myself to recreate these nice icons but could the plus and minus be a stopgap 😄

stefanosansone (approved):

jacksonh (approved):

PR #3909, opened 2024-05-07
Save Android offline content to files instead of in DB

jacksonh:
This works around max size constraints in the Room DB, we
will need to add something to clean up old files though.


jacksonh:
A couple open questions on this one:

1. Whats best way to clean up old files here. I suppose we can check for old files, then check if they are in the local database and clear them.
2. Where are these files on the filesystem? Can users access them and are we cluttering a shared directory?

PR #3916, opened 2024-05-08
feat: library as a channel and move digest to library api

sywhb:


packages/api/src/jobs/ai/create_digest.ts
```diff
@@ -608,10 +609,57 @@ const sendEmail = async (
   })
 }
 
-const sendNotifications = async (
+const findThumbnail = async (chapters: Chapter[]) => {
+  const images = await Promise.all(
+    chapters
+      .filter((chapter) => chapter.thumbnail)
+      .map((chapter) => getImageSize(chapter.thumbnail as string))
```
jacksonh, line 616: Does this have to make an HTTP request to fetch the image?

jacksonh (approved):

packages/api/src/jobs/ai/create_digest.ts
sywhb: Yeah 😞

jacksonh: I guess two suggestions: 1. lets make sure it can't fail, and 2. We could randomize the list and then just pick the first match that meets our size criteria

PR #3923, opened 2024-05-09
feat: update buttons in digest email

sywhb:
- **add "Update digest preference" button to the digest email**
- **only include the "Read in Omnivore" button if the digest is saved in library**

### DIFF
--- a/packages/api/src/jobs/ai/create_digest.ts
+++ b/packages/api/src/jobs/ai/create_digest.ts
@@ -602,13 +602,26 @@ const sendEmail = async (user: User, digest: Digest) => {
 
       <button style="background-color: #FFEAA0;
                     border: 0px solid transparent;
-                    color: rgb(42, 42, 42);
                     padding:15px 32px;
                     font-size: 14px;
                     margin: 20px 0;
                     font-family: Inter, sans-serif;
                     border-radius: 5px;">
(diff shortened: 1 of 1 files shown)


packages/api/src/jobs/ai/create_digest.ts
```diff
@@ -600,15 +601,33 @@ const sendEmail = async (user: User, digest: Digest) => {
...
+        isInLibrary
+          ? `<button style="background-color: #FFEAA0;
                     border: 0px solid transparent;
-                    color: rgb(42, 42, 42);
                     padding:15px 32px;
                     font-size: 14px;
                     margin: 20px 0;
                     font-family: Inter, sans-serif;
                     border-radius: 5px;">
-        <a href="${env.client.url}/digest/${digest.id}">Read in Omnivore</a>
```
jacksonh, line 611: In theory we could link directly to the library item here, like `omnivore.app/<username>/{id}` but actually i think this will just work as `digest` is ignored and the username if authenticated is gotten from the token, and then `id` matches the library item id, right?

jacksonh (approved):

packages/api/src/jobs/ai/create_digest.ts
sywhb: Yeah, that's right. The `username` or `digest` is ignored currently

PR #3934, opened 2024-05-10
Only allow eligible users to enable digest

jacksonh:


packages/api/src/services/features.ts
```diff
@@ -157,3 +172,13 @@ export const createFeature = async (feature: DeepPartial<Feature>) => {
 export const createFeatures = async (features: DeepPartial<Feature>[]) => {
   return getRepository(Feature).save(features)
 }
+
+export const userDigestEligible = async (uid: string): Promise<boolean> => {
+  const subscriptionsCount = await getRepository(Subscription).count({
+    where: { user: { id: uid }, status: SubscriptionStatus.Active },
+  })
+  const libraryItemsCount = await libraryItemRepository.count({
```
sywhb, line 180: I think we need `authTrx` to get over RLS

sywhb (approved):
I guess there could be an RLS issue and other than that is good

PR #3942, opened 2024-05-13
fix: puppeteer protocol timeout

sywhb:
- **print browser log**
- **add more logs to debug browser context**
- **do not pass browser instance to content-handler**
- **upgrade puppeteer-core to prevent ProtocolTimeout and adding more debug logs**
- **catch puppeteer page error**
- **capture content-fetch success and error events**

### DIFF
--- a/packages/content-fetch/src/request_handler.ts
+++ b/packages/content-fetch/src/request_handler.ts
@@ -1,5 +1,6 @@
 import { fetchContent } from '@omnivore/puppeteer-parse'
 import { RequestHandler } from 'express'
+import { analytics } from './analytics'
 import { queueSavePageJob } from './job'
 import { redisDataSource } from './redis_data_source'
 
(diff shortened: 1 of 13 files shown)


packages/content-fetch/src/request_handler.ts
```diff
@@ -147,11 +148,29 @@ export const contentFetchRequestHandler: RequestHandler = async (req, res) => {
       logRecord.error = 'unknown error'
     }
 
+    // capture error event
+    users.forEach((user) => {
+      analytics.capture({
+        distinctId: user.id,
+        event: 'content-fetch-failure',
```
jacksonh, line 155: Not sure why we picked it but we seem to use underscores so this would be `content_fetch_failure`

packages/content-fetch/src/request_handler.ts
```diff
@@ -147,11 +148,29 @@ export const contentFetchRequestHandler: RequestHandler = async (req, res) => {
...
+        properties: logRecord,
```
jacksonh, line 156: I think `logRecord` might have too much stuff in it. Maybe just send the URL?

packages/puppeteer-parse/src/index.ts
```diff
@@ -308,119 +316,122 @@ async function retrievePage(
...
-  // set locale for the page
-  if (locale) {
-    await page.setExtraHTTPHeaders({ 'Accept-Language': locale })
-  }
+    if (!enableJavascriptForUrl(url)) {
+      await page.setJavaScriptEnabled(false)
+    }
+    await page.setUserAgent(userAgentForUrl(url))
 
-  // set timezone for the page
```
jacksonh, line 324: Hard to tell from the PR, but did we lose the timezone setting in this patch?

packages/puppeteer-parse/src/index.ts
```diff
@@ -308,119 +316,122 @@ async function retrievePage(
...
-  await client.send('Network.setRequestInterception', {
-    patterns: [
-      {
-        urlPattern: '*',
-        resourceType: 'Document',
-        interceptionStage: 'HeadersReceived',
-      },
-    ],
-  })
+    // set timezone for the page
```
jacksonh, line 339: nevermind my above timezone comment

jacksonh (approved):

packages/content-fetch/src/request_handler.ts
sywhb: Yeah, good catch.

PR #3956, opened 2024-05-15
Upgrade to gpt-4o

jacksonh:


sywhb (approved):
I wonder if reading the model from the gist file might be better

jacksonh:
Yeah I was thinking about that although technically that model isn't used just by the digest config.

PR #3957, opened 2024-05-15
Cast digestConfig to any to serialize properly

jacksonh:


packages/api/src/resolvers/user_personalization/index.ts
```diff
@@ -17,19 +17,12 @@ export const setUserPersonalizationResolver = authorized<
...
-        digestConfig: () => {
-          return JSON.stringify(input.digestConfig)
-        },
-      }
-    : {}
   const result = await authTrx(async (t) => {
     return t.getRepository(UserPersonalization).upsert(
       {
         user: { id: uid },
+        digestConfig: input.digestConfig as any,
```
sywhb, line 24: I wonder if changing the entity class attribute type could be easier

sywhb: Something like `digestConfig: { channels: string[] }`

jacksonh: Yeah I have tried but fighting a bit with all the Maybe type stuff.

sywhb (approved):

PR #3958, opened 2024-05-15
fix: add author to the chapter in digest

sywhb:


jacksonh (approved):

packages/api/src/jobs/ai/create_digest.ts
```diff
@@ -832,6 +837,7 @@ export const createDigest = async (jobData: CreateDigestData) => {
         thumbnail: item.libraryItem.thumbnail ?? undefined,
         wordCount: speechFiles[index].wordCount,
         html: summariesInHtml[index],
+        author: item.libraryItem.author ?? '',
```
jacksonh, line 840: I think its better to leave null/undefined here if its not set. That way i wont do something like `by ${chapter.author}` on an empty item

packages/api/src/services/digest.ts
```diff
@@ -10,6 +10,7 @@ export interface Chapter {
   wordCount: number
   thumbnail?: string
   html: string
+  author: string
```
jacksonh, line 13: I think can be optional author?:

packages/api/src/jobs/ai/create_digest.ts
sywhb: Got it. I have thought double checking with you

PR #3960, opened 2024-05-15
allow downloading/uploading readable content

sywhb:


packages/api/src/jobs/process-youtube-video.ts
```diff
@@ -199,90 +187,21 @@ export const addTranscriptPlaceholdReadableContent = async (
...
-    readableStream.push(content)
-    readableStream.push(null) // Signal the end of the stream
-
-    // Pipe the readable stream to the writable stream to upload the file content
-    await new Promise((resolve, reject) => {
-      readableStream
-        .pipe(writableStream)
-        .on('finish', resolve)
-        .on('error', reject)
-    })
```
sywhb, line 263: I think `streaming` is not needed here because we still load the whole content in the memory and push it to the `readStream` at once so it should be the same effect as uploading the whole content at once

packages/api/src/jobs/process-youtube-video.ts
```diff
@@ -426,79 +326,63 @@ export interface ProcessYouTubeTranscriptJobData {
 export const processYouTubeTranscript = async (
   jobData: ProcessYouTubeTranscriptJobData
 ) => {
-  try {
```
sywhb, line 429: Removed the `try catch` block otherwise the failed jobs would never be retried

sywhb:
I simplified the code a bit, hopefully 😄

jacksonh (approved):

PR #3962, opened 2024-05-17
Link to chromestore is for old version of the store

Tsadoq:
I understand you might want to update the original link and not hardcode the link to the store.


jacksonh:
Thanks, yeah i will update the redirect

PR #3966, opened 2024-05-18
feature/add index for archived at and folder

sywhb:
- **use IN query for state**
- **add index for archived_at column**
- **fix archived query**
- **add index for folder column**
- **drop multiple-column index for user_id and state columns in library_item table since we create a new index for count covers them**


jacksonh (commented):
Probably just have to be careful about the deploy. Likely will timeout, right?

jacksonh (approved):

sywhb:
> Probably just have to be careful about the deploy. Likely will timeout, right?

Yeah, I plan to run the migration and deploy script separately and the migration script has a long timeout

Issue #3973, opened 2024-05-20
No content showing on E-Reader (android)


jason-mehmel:
I'm using a Boox e-reader android device, and today found that none of the articles I'd saved were visible when I loaded the app. 

I also noticed the logo changed, there must have been a recent update! I was using the app the day before with no problems.

It's working correctly on a Firefox browser and a Pixel 7, so this may be just for devices with e-ink displays?

jacksonh:
Hey, thanks we've had a few people report this. I suspect something to do with file permissions. Boox is a great clue, thank you.

jason-mehmel:
Ah, good to know! The fact that I use it on multiple platforms helped me notice the discrepancy.

jacksonh:
@jason-mehmel there is a new version `0.211.0` that i think should resolve it. I wasn't able to replicate on my Boox after updating.

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #3974, opened 2024-05-20
Use internal storage for content files

jacksonh:


jacksonh:
https://github.com/omnivore-app/omnivore/issues/3973 hopefully resolves this. On my Boox I've had trouble replicating consistently, but was able to in-consistently get the same failure.

PR #3982, opened 2024-05-22
feat: just read api

sywhb:


jacksonh:
LGTM

jacksonh:
nice work

packages/api/src/apollo.ts
```diff
@@ -32,6 +32,7 @@ import { ClaimsToSet, RequestContext, ResolverContext } from './resolvers/types'
 import ScalarResolvers from './scalars'
 import typeDefs from './schema'
 import { batchGetHighlightsFromLibraryItemIds } from './services/highlights'
+import { batchGetJustReadFeedItems } from './services/just_read_feed'
```
jacksonh, line 35: I think `home_feed` will make the most sense for this. If we want we can even just call it `Home` and `Home` has `HomeItem`s

packages/api/src/entity/library_item.ts
```diff
@@ -204,4 +204,16 @@ export class LibraryItem {
 
   @Column('text')
   highlightAnnotations?: string[]
+
+  @Column('timestamptz')
+  seenAt?: Date
+
+  @Column('text')
+  topic?: string
```
jacksonh, line 212: Should we do this as a tree? Probably fine either way for now.

packages/api/src/entity/public_item_interaction.ts
```diff
@@ -0,0 +1,47 @@
...
+  seenAt!: Date
+
+  @Column('timestamptz')
+  savedAt?: Date
+
+  @Column('timestamptz')
+  likedAt?: Date
+
+  @Column('timestamptz')
+  broadcastedAt?: Date
```
jacksonh, line 37: nice

packages/api/src/entity/public_item.ts
```diff
@@ -0,0 +1,73 @@
...
+  type!: string
+
+  @Column('text')
+  title!: string
+
+  @Column('text')
+  url!: string
+
+  @Column('boolean')
+  approved!: boolean
```
jacksonh, line 39: Thank you.

packages/api/src/jobs/update_just_read_feed.ts
```diff
@@ -0,0 +1,397 @@
...
+    .filter((item) => !!item.subscription)
+    .map((item) => item.subscription as string)
+
+  const subscriptions = await findSubscriptionsByNames(
+    userId,
+    subscriptionNames
+  )
+
+  // map library items to candidates and limit to 70
+  const privateCandidates: Array<Candidate> = libraryItems
```
jacksonh, line 128: I suspect we can tweak these counts a lot.

packages/api/src/jobs/update_just_read_feed.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  // map public items to candidates and limit to the remaining vacancies
+  const publicCandidates: Array<Candidate> = publicItems
+    .map(publicItemToCandidate)
+    .slice(0, 100 - privateCandidatesSize)
+
+  const publicCandidatesSize = publicCandidates.length
+  logger.info(`Found ${publicCandidatesSize} public candidates`)
+
+  // return 100 candidates and 70 from private and 30 from public
```
jacksonh, line 150: Small typo in this i think

jacksonh (changes requested):
I think the main change is renaming to `Home` that will be consistent with our UI

packages/api/src/entity/library_item.ts
sywhb: Yep, I forget to change this 😢

packages/api/src/jobs/update_just_read_feed.ts
sywhb: Yeah

packages/api/src/jobs/update_just_read_feed.ts
```diff
@@ -0,0 +1,397 @@
...
+        subscription_type: item.subscription?.type,
+        author: item.author,
+        word_count: item.wordCount,
+        published_at: item.publishedAt,
+      } as Feature
+      return acc
+    }, {} as Record<string, Feature>),
+  }
+
+  const newScores = await getScores(data)
```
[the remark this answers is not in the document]
sywhb: Currently I will update the score for a private library item once it is saved so if this score is available, we will use it instead of calling the score API again

jacksonh (approved):

PR #3998, opened 2024-05-28
fix/home API

sywhb:
- **rename just-read-feed to home in redis**
- **fix removing new items from redis**
- **add start and end cursor to the pageinfo**

### DIFF
--- a/packages/api/src/jobs/update_home.ts
+++ b/packages/api/src/jobs/update_home.ts
@@ -209,7 +209,7 @@ const rankCandidates = async (
   return candidates
 }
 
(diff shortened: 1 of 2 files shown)


packages/api/src/jobs/update_home.ts
```diff
@@ -209,7 +209,7 @@ const rankCandidates = async (
   return candidates
 }
 
-const redisKey = (userId: string) => `just-read-feed:${userId}`
```
jacksonh, line 212: 👍

jacksonh (approved):

PR #4000, opened 2024-05-28
fix: home items order is not accurate

sywhb:


sywhb:
@jacksonh Does `public_item` also need a slug?

jacksonh (approved):

sywhb:
> @jacksonh Does `public_item` also need a slug?

@jacksonh Probably you missed out the question here

sywhb:
> > @jacksonh Does `public_item` also need a slug?
> 
> @jacksonh Probably you missed out the question here

@jacksonh I will merge this first so we can better test

PR #4001, opened 2024-05-29
feat: create an api to get subscription by id

sywhb:


packages/api/src/scalars.ts
```diff
@@ -14,7 +14,7 @@ export class SanitizedString extends GraphQLScalarType {
   ) {
     super({
       // Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ as per graphql-js
-      name: `SanitizedString_${allowedTags}_${maxLength}_${pattern}`.replace(
+      name: `SanitizedString_${allowedTags}_${maxLength}_${minLength}_${pattern}`.replace(
```
jacksonh, line 17: Was this one required? I'm a bit worried this will change some API signatures.

sywhb: Yeah it is required otherwise `type`s with different `minLength` directive would fail.

Why will this change API signatures?

jacksonh: I think it creates a new type. I'll pull this and run the iOS binding to verify its not an issue

sywhb: Thanks!

jacksonh: ok yeah seems to have no issues

jacksonh (approved):

PR #4011, opened 2024-05-30
Add thumbsup/down feedback buttons for the home feed

jacksonh:


packages/web/pages/justread/index.tsx
```diff
@@ -145,9 +154,9 @@ const HomeItemView = (props: HomeItemViewProps): JSX.Element => {
       }}
       onClick={(event) => {
         if (event.metaKey || event.ctrlKey) {
-          window.open(props.homeItem.url, '_blank')
+          window.open(`/me/${props.homeItem.slug}`)
```
sywhb, line 157: Is there any reason we use `window.open` here?

packages/web/pages/justread/index.tsx
```diff
@@ -329,6 +369,30 @@ const SubscriptionSourceHoverContent = (
...
+            event.preventDefault()
+            event.stopPropagation()
+          }}
+        >
+          <ThumbsUp weight="fill" />
+        </Button>
+        <Button
+          style="plainIcon"
+          onClick={(event) => {
+            sendHomeFeedback('MORE')
```
sywhb, line 387: This should be `LESS`

packages/web/pages/justread/index.tsx
```diff
@@ -281,6 +290,37 @@ const SubscriptionSourceHoverContent = (
     return undefined
   }, [subscriptions])
 
+  const sendHomeFeedback = useCallback(
+    (feedbackType: SendHomeFeedbackType) => {
+      ;(async () => {
```
sywhb, line 295: Why do we need IIFE here?

packages/api/src/entity/home_feedback.ts
```diff
@@ -0,0 +1,45 @@
...
+export class HomeFeedback {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string
+
+  @ManyToOne(() => User)
+  @JoinColumn({ name: 'user_id' })
+  user!: User
+
+  @Column('text', { nullable: true })
+  subscription!: string | null
```
sywhb, line 30: This should be `optional`

packages/api/src/resolvers/home_feedback/index.ts
```diff
@@ -0,0 +1,100 @@
...
+      feedback: input.feedbackType,
+    },
+  })
+
+  if (!author && !site && !subscription) {
+    log.error('author, site, or subscription is required')
+    return {
+      errorCodes: [ErrorCode.BadRequest],
+    }
+  }
```
sywhb, line 44: Probably better added the `sanitised` directive to the field

packages/db/migrations/0178.do.create_home_feedback_table.sql
```diff
@@ -0,0 +1,29 @@
...
+-- Name: create_home_feedback
+-- Description: Create a table for user feedback of the Home feed
+
+BEGIN;
+
+CREATE TYPE omnivore.feedback_type_enum AS ENUM ('MORE', 'LESS');
+
+CREATE TABLE IF NOT EXISTS omnivore.home_feedback (
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v1mc(),
+    user_id uuid NOT NULL REFERENCES omnivore.user(id) ON DELETE CASCADE,
```
sywhb, line 11: Probably should create an index for `user_id`

packages/api/src/resolvers/home_feedback/index.ts
```diff
@@ -0,0 +1,100 @@
...
+})
+
+export const homeFeedbackResolver = authorized<
+  HomeFeedbackSuccess,
+  HomeFeedbackError
+>(async (_parent, _args, { claims: { uid }, log }) => {
```
sywhb, line 71: Seems the `limit` and `offset` are not used?

packages/api/src/services/home_feedback.ts
```diff
@@ -0,0 +1,103 @@
...
+      )
+      if (result) {
+        return result
+      }
+    }
+    throw e
+  }
+}
+
+export const deleteHomeFeedback = async (
```
sywhb, line 82: The `DELETE` permission is not granted

sywhb (commented):
Thanks for the PR. I just added a few comments

packages/db/migrations/0178.do.create_home_feedback_table.sql
```diff
@@ -0,0 +1,29 @@
...
+    site TEXT,
+    author TEXT,
+    subscription TEXT,
+    feedback_type omnivore.feedback_type_enum,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    UNIQUE (user_id, subscription, feedback_type),
```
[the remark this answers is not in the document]
jacksonh: Yeah you can have MORE and LESS, which would just get canceled out when scoring

packages/web/pages/justread/index.tsx
jacksonh: Yeah this is if you are holding down the CTRL key it will open in a new tab.

packages/db/migrations/0178.do.create_home_feedback_table.sql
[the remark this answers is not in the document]
sywhb: What if I do:

More -> Less -> More again?

packages/web/pages/justread/index.tsx
jacksonh: Ah i guess this used to be needed, not so much now, you can make useCallback async

packages/api/src/resolvers/home_feedback/index.ts
jacksonh: @sywhb so I need to check if at least one is set, but others can be null. Can I do that with `sanitised`?

sywhb (approved):
LGTM

PR #4013, opened 2024-05-30
Android app improvements

mhss1:
- Added background sync for library when adding item from share menu
(Also, Replaced using the ViewModel in SaveSheetActivity with SaveURLWorker to ensure that the url will be saved. When using the ViewModel and user exists quickly, the url doesn't get saved because the viewmodel gets cleared before saving)
- Changed typeahead search to get data from offline db instead of network
- Improved the handling of Jetpack Compose side effects in some screens
- Removed some unnecessary code
- Added animations for saved items list when an item is added/removed


jacksonh:
Thank you!

Issue #3970, opened 2024-05-19
RSS feed Article thumbnails are not displaying for many Websites.


Nexus-Gits:
Fix failed to get thumbnail from rss feed. Rss does not show any thumbail to this rss feed:

1. https://beebom.com/feed/
2. https://techcrunch.com/feed/

Also add the support for avatars in account.  @sywhb @jacksonh @satindar

sywhb:
Subscribed to them for testing

sywhb:
Avatars are a nice thing to add 😄

sywhb:
The fix is in demo environment and the thumbnails are shown correctly

PR #4019, opened 2024-06-01
fix: fail to display the thumbnail of some webpages

sywhb:
* The current regex we use will match `og:image:type` for thumbnail url
* Fix that by only matching `og:image`, `og:image:url` or `og:image:secure_url`

References: https://ogp.me/


sywhb:
Fix #3970

jacksonh (approved):

PR #4040, opened 2024-06-10
Update layout navigation with left menu

jacksonh:
- Setup new left navigation system
- Start adding allotment cleaning up navigation
- Work on new left menu w/shortcuts
- Update SWR usage, move highlights temporarily
- Add migrations


packages/api/src/routers/shortcuts_router.ts
```diff
@@ -0,0 +1,107 @@
...
+        logger.info('failed to authorize')
+        return res.status(401).send('UNAUTHORIZED')
+      }
+    } catch (e) {
+      logger.info('failed to authorize', e)
+      return res.status(401).send('UNAUTHORIZED')
+    }
+
+    try {
+      const shortcuts = await setShortcuts(claims.uid, req.body.shortcuts)
```
sywhb, line 64: We should have some data validation here

sywhb: It would be the best to have some integration tests

packages/db/migrations/0180.do.add_shortcuts_to_user_personalization.sql
```diff
@@ -0,0 +1,9 @@
+-- Type: DO
+-- Name: add_shortcuts_to_user_personalization
+-- Description: Add a new shortcuts column to the user personalization table
+
+BEGIN;
+
+ALTER TABLE omnivore.user_personalization ADD COLUMN shortcuts JSONB NULL;
```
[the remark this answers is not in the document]
jacksonh: Yeah i think it is the same but made it more consistent

PR #4041, opened 2024-06-10
feat: folder expiration policy api

sywhb:
# Folder expiration policy API

There is a folder expiration policy and it defines how `long` will the items in `folder` be `done`

Existing users will be asked to opt in and new users will be opted in by default

- inbox: archived after 60 days
- following: deleted after 7 days

## Backend implementation:
We have a cronjob runs every day at the off-peak time
- triggers a job to "expire items in folders"
	- runs a query to get all folder policy
	- for each folder policy
		- `update library_item set state = 'DELETED', deleted_at = now() where user_id='1' and folder = 'following' and saved_at < '2021';`
- a folder_policy table
	- id
	- user_id
	- folder
	- action
	- after_days
	- Unique(user_id, folder)
- API: graphql
  - **add list folder policy api**
  - **add create folder policy api**
  - **add update folder policy api**
  - **add delete folder policy api**
  - **add folder expiration jobs**

## challenge
update query is slow and locking rows
Commit each batch in the transaction


jacksonh:
Seems good to me, I realize now that its "opt in" so OK to merge without worrying too much about the performance issues.

PR #4046, opened 2024-06-11
open youtube transcript to public

sywhb:


jacksonh (approved):

jacksonh:
Once this is live we can update web.

PR #4052, opened 2024-06-12
feature/prune trash items

sywhb:
- **create a function to batch delete items in trash**
- **update job name**
- **catch sync reading progress query error**
- **worker can concurrently process 10 jobs**
- **commit the delete query for each user in the loop to reduce lock time**
- **remove timeout in highlight generator**
- **fix typo in sql**

### DIFF
--- a/packages/api/src/jobs/upload_content.ts
+++ b/packages/api/src/jobs/upload_content.ts
@@ -4,7 +4,7 @@ import { logger } from '../utils/logger'
 import { htmlToHighlightedMarkdown, htmlToMarkdown } from '../utils/parser'
 import { isFileExists, uploadToBucket } from '../utils/uploads'
 
-export const UPLOAD_CONTENT_JOB = 'UPLOAD_CONTENT_JOB'
+export const UPLOAD_CONTENT_JOB = 'upload-content'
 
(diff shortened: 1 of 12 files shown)


packages/api/src/jobs/upload_content.ts
```diff
@@ -4,7 +4,7 @@ import { logger } from '../utils/logger'
 import { htmlToHighlightedMarkdown, htmlToMarkdown } from '../utils/parser'
 import { isFileExists, uploadToBucket } from '../utils/uploads'
 
-export const UPLOAD_CONTENT_JOB = 'UPLOAD_CONTENT_JOB'
+export const UPLOAD_CONTENT_JOB = 'upload-content'
```
jacksonh, line 7: with this name change, we might have a small issue with items in the queue during deploy.

jacksonh (approved):
Looks good to me. Maybe we can test on the replica?

packages/api/src/jobs/upload_content.ts
sywhb: I wonder if I can pause the queue a little bit when all the `upload-content` jobs are done

PR #4075, opened 2024-06-18
feat: posts api

sywhb:
# Post
a public link omnivore.app/list/userId/posts/id which omnivore user shares through social apps
- a list of library items
- selected highlights (optional)
- thoughts (optional)
- thumbnail picture (optional)

## List
a link omnivore.app/list/userId/posts contains a list of "Post" shared by the user which user can share through social apps.

It can be accessed if user make the profile public

### Changes
- **create post table and entity**
- **add list posts api and tests**
- **tidy**
- **add get post graphql api and tests**
- **add create post graphql api and tests**
- **add update post graphql api and tests**
- **add delete post graphql api and tests**


sywhb:
There are lots of "tidy" changes in this PR which you can ignore

packages/api/test/resolvers/post.test.ts
```diff
@@ -0,0 +1,492 @@
...
+      await deleteLibraryItemById(libraryItem1.id, loginUser.id)
+    })
+
+    it('should return an error if the args are invalid', async () => {
+      const response = await graphqlRequest(query, '', {
+        first: 100,
+        userId: loginUser.id,
+      })
+
+      expect(response.body.data.posts.errorCodes).to.eql(['BAD_REQUEST'])
```
pradeepmurugesan, line 99: newbie question. What is the invalid arg here ??  I see a similar request below, testing for an empty array..  Is `100` the invalid arg here ?

```
 it('should return empty array', async () => {
          const response = await graphqlRequest(query, '', {
            first: 10,
            userId: loginUser.id,
          })

          expect(response.body.data.posts.edges).to.be.empty
        })
```

jacksonh (approved):

jacksonh (approved):

packages/api/test/resolvers/post.test.ts
sywhb: Hey @pradeepmurugesan , sorry for the late reply.

I should be more clear here but yes 100 is invalid because the maximum number of posts allowed to requests a time is 50.

PR #4078, opened 2024-06-18
create index concurrently on topic column in library_item table

sywhb:


.github/workflows/run-tests.yaml
```diff
@@ -96,21 +96,21 @@ jobs:
           PG_LOGGER: debug
           REDIS_URL: redis://localhost:${{ job.services.redis.ports[6379] }}
           MQ_REDIS_URL: redis://localhost:${{ job.services.redis.ports[6379] }}
-  build-docker-images:
```
jacksonh, line 99: This one isnt needed anymore?

jacksonh (approved):

.github/workflows/run-tests.yaml
sywhb: I was testing it just now and reverted the changes

PR #4082, opened 2024-06-19
Add user field to the ArticleSavingRequest resolver

jacksonh:


sywhb (approved):

packages/api/src/resolvers/function_resolvers.ts
```diff
@@ -785,6 +785,9 @@ export const functionResolvers = {
   ArticleSavingRequest: {
     status: (item: LibraryItem) => item.state,
     url: (item: LibraryItem) => item.originalUrl,
+    async user(item: LibraryItem, __: unknown, ctx: WithDataSourcesContext) {
+      return ctx.dataLoaders.users.load(item.userId)
```
sywhb, line 789: Probably not necessary but better to use `ctx.uid` as the key

jacksonh: Ah thanks, much better

PR #4122, opened 2024-06-30
Migrate deprecated Google SignIn + disable android backup

mhss1:
- Migrated the deprecated Google SignIn library and used the new [Credential Manager](https://developer.android.com/identity/sign-in/credential-manager-siwg)
- Disabled android automatic backup as it uploads the api token in plain text and the cached sqlite data as well which might cause problems with data already on server

### DIFF
--- a/android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/onboarding/auth/AuthProviderScreen.kt
+++ b/android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/onboarding/auth/AuthProviderScreen.kt
@@ -17,7 +17,6 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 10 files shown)


jacksonh:
Sorry for the delay on this one. This stuff is a pain to test because of the signing.

android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/onboarding/auth/AuthProviderScreen.kt
```diff
@@ -27,28 +26,25 @@ import app.omnivore.omnivore.feature.onboarding.OnboardingViewModel
...
 import app.omnivore.omnivore.feature.onboarding.auth.provider.GoogleAuthButton
 import app.omnivore.omnivore.navigation.Routes
-import com.google.android.gms.common.GoogleApiAvailability
 
 @Composable
 fun AuthProviderScreen(
     welcomeNavController: NavHostController,
     viewModel: OnboardingViewModel
 ) {
-    val isGoogleAuthAvailable: Boolean =
```
jacksonh, line 37: Do we have to remove this? We have some cases where the user installs not through Play.

android/Omnivore/app/src/main/res/xml/data_extraction_rules.xml
```diff
@@ -5,15 +5,17 @@
 -->
 <data-extraction-rules>
     <cloud-backup>
-        <!-- Use <include> and <exclude> to control what is backed up.
-        <include .../>
-        <exclude .../>
-        -->
+        <exclude domain="root" />
```
jacksonh, line 8: Sorry what is this stuff for?

android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/onboarding/auth/AuthProviderScreen.kt
mhss1: It was checking if google play services is available or not. It wasn't checking if the app is installed from Google Play.
Someone can still install it from somewhere else and still has Google play services.

android/Omnivore/app/src/main/res/xml/data_extraction_rules.xml
mhss1: It disables [Android Auto Backup](https://developer.android.com/identity/data/autobackup) which uploads the app data like SQLite and datastore data which includes the auth token. Uploading SQLite data isn't necessary and might even cause problems because the data might have been updated after the backup. And uploading the auth token is obviously bad.

jacksonh: oh awesome, thanks!

jacksonh (approved):

PR #4136, opened 2024-07-02
fix: pdf handler sends update-content-job to the wrong redis

sywhb:
* create a shared utils package for creating redis connection


packages/pdf-handler/src/index.ts
```diff
@@ -130,6 +144,7 @@ export const pdfHandler = Sentry.GCPFunction.wrapHttpFunction(
       } finally {
         // Always update the state, even if we fail to parse
```
jacksonh, line 145: ❤️

jacksonh (approved):

Issue #3789, opened 2024-04-05
Highlight order in the notebook


sywhb:
【圆圈高亮标注内容】 应该要在 【方框高亮标注内容】 之后在符合行文逻辑

![image](https://github.com/omnivore-app/omnivore/assets/12963965/ec10e42f-63ac-475e-8fa1-6835bc1a8481)

kuosandys:
I'd like to help solve this issue - could I be assigned to this if no one is currently working on it?

jacksonh:
@kuosandys thanks!

aassoiants:
I still find this to be an issue

PR #4141, opened 2024-07-03
Fix highlights order in notebooks

kuosandys:
Fixes an issue with highlights display order in notebooks, as described in https://github.com/omnivore-app/omnivore/issues/3789:
- highlights with the same highlight position (i.e. in same line) are now sorted by their location within the line
- the same highlight sorting was performed in a few places - the sorting function been refactored to one common location

I also see that there's a new Highlights view in the works - I didn't want to increase the scope of this bug fix PR to include feature work, but let me know if any help is needed there!


jacksonh (approved):

jacksonh:
Beautiful, thanks

PR #4142, opened 2024-07-04
increase score api timeout to 20 seconds

sywhb:


jacksonh:
A bit worried about this one!

jacksonh (approved):

PR #4152, opened 2024-07-05
fix: use software graphic rendering instead of gpu and reduce browser launch timeout to 10 seconds

sywhb:


jacksonh (approved):

packages/puppeteer-parse/src/browser.ts
```diff
@@ -56,7 +57,7 @@ export const getBrowser = async (): Promise<Browser> => {
     },
     executablePath: process.env.CHROMIUM_PATH,
     headless: !!process.env.LAUNCH_HEADLESS,
-    timeout: 30_000, // 30 seconds
+    timeout: 10_000, // 10 seconds
```
jacksonh, line 60: ❤️

PR #4153, opened 2024-07-05
have separate intercom secret in env var and return the hash based on the client

sywhb:


packages/api/src/util.ts
```diff
@@ -268,6 +274,9 @@ export function getEnv(): BackendEnv {
   const intercom = {
     token: parse('INTERCOM_TOKEN'),
     secretKey: parse('INTERCOM_SECRET_KEY'),
+    webSecret: parse('INTERCOM_WEB_SECRET'),
```
jacksonh, line 277: Are these ones mandatory by default?

sywhb: No

sywhb: Should we make them mandatory by default? @jacksonh

jacksonh: @sywhb no, i think they should be optional so it doesn't break for self hosting. I just wanted to confirm.

jacksonh (approved):

PR #4162, opened 2024-07-06
Handle foreground/background sync on older + newer Android versions

jacksonh:


android/Omnivore/app/src/main/java/app/omnivore/omnivore/feature/save/SaveSheetActivity.kt
```diff
@@ -133,28 +135,25 @@ class SaveSheetActivity : AppCompatActivity() {
...
+        val syncWork = OneTimeWorkRequestBuilder<LibrarySyncWorker>()
+            .setConstraints(Constraints.Builder()
+                .setRequiredNetworkType(NetworkType.CONNECTED)
+                .build())
+            .setInitialDelay(5, TimeUnit.SECONDS)
             .build()
 
-        beginWith(saveURLWorkRequest)
-            .then(syncWorkerRequest)
+        WorkManager.getInstance(context)
```
mhss1, line 154: This is not needed btw, this is an Extension funtion that uses the WorkManager instance already created in onCreate.
So it can be like this without it:
```
beginUniqueWork("saveAndSync", ExistingWorkPolicy.REPLACE, saveWork)
            .then(syncWork)
            .enqueue()
```
also the context parameter can be removed

PR #4171, opened 2024-07-09
Fix string interp on update label mutation

jacksonh:
We should always use input instead of strings like this.


noghartt (approved):
Really cool, thanks at all!

jacksonh:
> Really cool, thanks at all!

Will be a bit before its deployed as we have a bunch of web stuff in the works.

PR #4185, opened 2024-07-10
fix/home

sywhb:
- **fix: include just added items in top picks**
- **move items which have less than 250 words to quick links section**


jacksonh:
Thank you!

PR #4188, opened 2024-07-10
Add Amplenote origins to cors config

debanjandhar12:
Hello,

[Amplenote](https://amplenote.com/) is a note-taking app similar to Obsidian and Logseq. I'm currently working on developing a [plugin](https://github.com/debanjandhar12/my-amplenote-plugins-v2/tree/master/src-omnivore) for Amplenote that syncs with Omnivore.  However, I am facing issues due to CORS.

This change will fix that issue. Would it be possible to merge this PR?

### DIFF
--- a/packages/api/src/utils/corsConfig.ts
+++ b/packages/api/src/utils/corsConfig.ts
@@ -14,6 +14,8 @@ export const corsConfig = {
     env.client.url,
     'lsp://logseq.io',
     'app://obsidian.md',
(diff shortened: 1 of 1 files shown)


jacksonh:
Thanks @debanjandhar12, so i guess this is a chromium app? Just wondering are you testing on both desktop and mobile? I've noticed with some of the others it gives different origins based on the platform.

jacksonh (approved):

PR #4196, opened 2024-07-11
create a GraphQL API to create youtube transcript job on demand

sywhb:


packages/api/src/server.ts
```diff
@@ -63,6 +65,7 @@ export const createApp = (): Express => {
   app.use(cookieParser())
   app.use(json({ limit: '100mb' }))
   app.use(urlencoded({ limit: '100mb', extended: true }))
+  app.use(cors(corsConfig))
```
jacksonh, line 68: Is this just needed for all requests?

sywhb: Good catch~

sywhb:
> Looks good to me, one thought: we have the `Task` API in GQL, not sure if you considered that?

I looked at the GQL schema and didn't find it. Could you show me the API?

sywhb:
@jacksonh Do you mean Subscription in GQL?

jacksonh:
> @jacksonh Do you mean Subscription API in GQL?

Oh i meant this:

```
  type Task {
    id: ID!
    name: String!
    state: TaskState!
    createdAt: Date!
    runningTime: Int # in milliseconds
    cancellable: Boolean
    progress: Float
    failedReason: String
  }
  ```

jacksonh:
Going to try and do the client side pieces on top of this one.

PR #4201, opened 2024-07-15
Added url matching option to library

AndreasThinks:
I've tried to add the option to match by URL.  Still testing this locally, but though I'd open PR in the interim for feedback.


jacksonh:
One thought is maybe it could be `startsWith` instead of `===`

jacksonh:
@sywhb can you look at this one?

sywhb (approved):

sywhb:
This filter would only be applied to rules and we have to update another method if we want to apply it to the search function

PR #4203, opened 2024-07-15
feat: task based content-fetch

sywhb:
- **create a worker to process content-fetch job**
- **enqueue content-fetch task to the queue**


jacksonh:
I guess we should try this one in demo?

sywhb:
> I guess we should try this one in demo?

Yeah

sywhb:
@jacksonh This is deployed in demo and doing some testing

jacksonh (approved):

PR #4207, opened 2024-07-18
fix/published date change in wechat

sywhb:
- **fix: wechat official account articles published date not captured because its format is changed**
- **remove metadata and cover image from content**


sywhb:
fix #4198

jacksonh (approved):

PR #4215, opened 2024-07-19
create a cloud run service for the exporter

sywhb:


jacksonh:
Looks great. I think before release we should also probably add highlights and page content, but this is a really good start.

sywhb:
> Looks great. I think before release we should also probably add highlights and page content, but this is a really good start.

Yeah, totally. I wonder what is the best way to export highlights and page content.

We have the content files ready in GCS which we could create multiple compressed files?

Regarding highlights, I'm not sure what to do

jacksonh (approved):

PR #4221, opened 2024-07-24
improve content-fetch

sywhb:
Some improvement on content-fetch:
1. remove scrapingbee
2. fix the sql to count saved items in database within a minute using `created_at` instead of `saved_at` as `saved_at` can be manually input
3. add an hour rate limiter to the API router which only allows 600 authenticated requests and 150 non-authenticated requests per hour on top of the minute rate limiter
4. hardcode some domains to block while fetching content and blocks a domain for a day if it was failed to fetch for 10 times in a day


packages/api/src/server.ts
```diff
@@ -68,7 +68,7 @@ export const createApp = (): Express => {
   app.set('trust proxy', env.server.trustProxy)
 
   // Apply the rate limiting middleware to API calls only
-  app.use('/api/', apiLimiter)
+  app.use('/api/', apiLimiter, apiHourLimiter)
```
jacksonh, line 71: oh wow, didn't know we could do this 👍

jacksonh (approved):

packages/api/src/server.ts
sywhb: Yeah, it will use the stricter limiter

sywhb:
> # Squawk Report
> ### **✅ 0** violations across **1** file(s)
> ### `packages/db/migrations/0184.do.add_user_id_created_at_index_to_library_item.sql`
> ```sql
> -- Type: DO
> -- Name: add_user_id_created_at_index_to_library_item
> -- Description: Add an index of columns user_id and created_at to the library_item table for counting query
> 
> CREATE INDEX CONCURRENTLY IF NOT EXISTS library_item_user_id_created_at_idx ON omnivore.library_item (user_id, created_at);
> ```
> 
> #### ✅ Rule Violations (0)
> No violations found.
> 
> [📚 More info on rules](https://github.com/sbdchd/squawk#rules)
> 
> ⚡️ Powered by [`Squawk`](https://github.com/sbdchd/squawk) (1.1.2), a linter for PostgreSQL, focused on migrations

@jacksonh I created an index on columns `user_id` and `created_at` on table `library_item` for the counting query

packages/api/src/jobs/bulk_action.ts
```diff
@@ -27,6 +27,7 @@ export const bulkAction = async (data: BulkActionData) => {
   for (let offset = 0; offset < count; offset += batchSize) {
     const searchArgs = {
       size: batchSize,
+      includePending: true,
```
jacksonh, line 30: oh thanks, i was going to ask for this

jacksonh (approved):
Actually a pretty big performance improvement too i think as the COUNT calls were adding up.

packages/api/src/jobs/bulk_action.ts
sywhb: Yeah, I found this bug while testing the content-fetch service

packages/api/src/repository/library_item.ts
```diff
@@ -25,7 +25,7 @@ export const libraryItemRepository = appDataSource
         .andWhere('md5(original_url) = md5(:url)', { url })
 
       if (forUpdate) {
-        qb.setLock('pessimistic_write')
+        qb.setLock('pessimistic_read')
```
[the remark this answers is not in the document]
sywhb: I replaced `pessimistic_write` with `pessimistic_read` lock strategy so it will prevent other transactions from modifying the data while it is being read but allows other transactions to read the data concurrently.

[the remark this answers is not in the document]
sywhb: @jacksonh

jacksonh:
LGTM

PR #4244, opened 2024-08-06
drop original_content column from library_item table

sywhb:


jacksonh:
Is it possible to do `IF EXISTS` with `DROP`?

sywhb:
> Is it possible to do `IF EXISTS` with `DROP`?

I don't think it is supported

sywhb:
`@jacksonh I guess I was wrong: `DROP [ COLUMN ] [ IF EXISTS ] column_name [ RESTRICT | CASCADE ]`
https://www.postgresql.org/docs/12/sql-altertable.html

sywhb:
@jacksonh Added `IF EXISTS` to the migration sql

jacksonh (approved):

PR #4245, opened 2024-08-06
stop adding original content from popular read

sywhb:


sywhb:
@jacksonh I guess we have to merge and deploy this branch before dropping the column as the we are using the `original_content` from `popular_read`

jacksonh (approved):

PR #4251, opened 2024-08-13
cache total count of library items in 60 seconds

sywhb:


jacksonh:
I think we could cache even more aggressively (longer timeline or more data, but this is a great start).

jacksonh (approved):

PR #4270, opened 2024-08-15
perf/api

sywhb:
- **perf: compress response body**
- **perf: stop serving content in search and update since api**
- **perf: reduce number of column returned by search query**
- **perf: add index on labels columns to library_item table**
- **Revert "perf: add index on labels columns to library_item table"**
- **perf: remove additional order by clause to use the correct order index**


packages/api/src/entity/library_item.ts
```diff
@@ -58,7 +58,7 @@ export class LibraryItem {
   @Column('text')
   originalUrl!: string
 
-  @Column('text', { nullable: true })
+  @Column('text', { nullable: true, select: false })
```
jacksonh, line 61: Can you tell me what this does? It looks like iOS expects this, but I think we handle it in the resolver now

sywhb: I don't think we use it at all. We planned to use it for downloading

PR #4310, opened 2024-08-22
fix/verification token

sywhb:
- **store verification token in redis with exp and destroy after use**
- **cache api key and invalidate it once api key is revoked**

### DIFF
--- a/packages/api/src/resolvers/api_key/index.ts
+++ b/packages/api/src/resolvers/api_key/index.ts
@@ -16,7 +16,11 @@ import {
 import { getRepository } from '../../repository'
 import { findApiKeys } from '../../services/api_key'
 import { analytics } from '../../utils/analytics'
-import { generateApiKey, hashApiKey } from '../../utils/auth'
+import {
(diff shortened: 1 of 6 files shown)


packages/api/src/resolvers/api_key/index.ts
```diff
@@ -92,6 +96,8 @@ export const revokeApiKeyResolver = authorized<
 
     const deletedApiKey = await apiRepo.remove(apiKey)
 
+    await deleteCachedClaims(deletedApiKey.key)
```
jacksonh, line 99: 👍

packages/api/src/utils/auth.ts
```diff
@@ -23,17 +23,20 @@ export const comparePassword = async (password: string, hash: string) => {
 }
 
 export const generateApiKey = (): string => {
-  // TODO: generate random string key
```
jacksonh, line 26: 😆

jacksonh (approved):

PR #4312, opened 2024-08-23
fix: failed to search items by rss feed url which has query parameters

sywhb:
* encode uri component of the rss feed url
* use descriminated unions to differetiate rss and newsletter subscription

### DIFF
--- a/packages/api/src/services/library_item.ts
+++ b/packages/api/src/services/library_item.ts
@@ -324,17 +324,16 @@ export const buildQueryString = (
               // return only deleted pages within 14 days
               return "(library_item.state = 'DELETED' AND library_item.deleted_at >= NOW() - INTERVAL '14 days')"
             default: {
-              let sql = 'library_item.archived_at IS NULL'
               if (useFolders) {
                 const param = `folder_${parameters.length}`
                 const folderSql = escapeQueryWithParameters(
                   `library_item.folder = :${param}`,
                   { [param]: value }
                 )
-                sql = `(${sql} AND ${folderSql})`
+                return `(library_item.archived_at IS NULL AND ${folderSql})`
               }
 
-              return sql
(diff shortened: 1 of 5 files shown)


packages/api/src/services/library_item.ts
```diff
@@ -321,17 +321,16 @@ export const buildQueryString = (
...
             default: {
-              let sql = 'library_item.archived_at IS NULL'
               if (useFolders) {
                 const param = `folder_${parameters.length}`
                 const folderSql = escapeQueryWithParameters(
                   `library_item.folder = :${param}`,
                   { [param]: value }
                 )
-                sql = `(${sql} AND ${folderSql})`
+                return `(library_item.archived_at IS NULL AND ${folderSql})`
```
jacksonh, line 330: Did you want these changes in this PR too?

sywhb: Thanks! I will move these backend changes to a separate PR

packages/web/pages/settings/feeds/index.tsx
```diff
@@ -266,7 +271,11 @@ export default function Rss(): JSX.Element {
                 </VStack>
               }
               onClick={() => {
-                router.push(`/home?q=in:inbox rss:"${subscription.url}"`)
+                router.push(
+                  `/home?q=in:inbox rss:"${encodeURIComponent(
```
jacksonh, line 275: Actually this should be `search?q=`

jacksonh (approved):

jacksonh:
Just one small change as we don't use home for queries anymore.

packages/web/pages/settings/feeds/index.tsx
sywhb: Thanks!

Issue #4321, opened 2024-08-25
CSV import errors on input aren't shown if the user has no existing links in library


tameTNT:
I tried to upload a malformed CSV file (timestamps not formatted correctly to millisecond precision) to my account which had no existing links (I'd cleared/archived the default pre-added ones). I uploaded the file and it just said "Error Uploading" in the upload box with no popups or further error details in the console or on the webpage.

After I added a link to another site manually, I tried importing the file again the same way and this time a popup did appear in the lower-right (in addition to the "Error Uploading" status) specifying a malformed timestamp on line 2, column 4 which I was then able to resolve and successfully import the links.

jacksonh:
Ahh thank you. I see how this can happen, the error provider is a component of the library but in empty state there is no library. Thank you for providing the extra details.

Podginator:
Closing due to age, and the self-hosted nature of Omnivore. Done as a batch action. If this issue still exists, please reopen it.

PR #4324, opened 2024-08-26
Move text utils into single place

jacksonh:


jacksonh:
Should help with this https://github.com/omnivore-app/omnivore/issues/4321#issuecomment-2309211175

PR #4327, opened 2024-08-26
feat: update exporter

sywhb:
- **exports/{userId}/{date}/{uuid}.zip   - metadata_{start_page}_to_{end_page}.json   - /content     - {slug}.html   - /highlights     - {slug}.md**
- **add export job and get api**

### DIFF
--- a/packages/api/src/jobs/export.ts
+++ b/packages/api/src/jobs/export.ts
@@ -1,0 +1,40 @@
+import axios from 'axios'
+import jwt from 'jsonwebtoken'
+import { env } from '../env'
+import { findActiveUser } from '../services/user'
+import { logger } from '../utils/logger'
+
+export interface ExportJobData {
+  userId: string
+}
+
+export const EXPORT_JOB_NAME = 'export'
+
+export const exportJob = async (jobData: ExportJobData) => {
+  const { userId } = jobData
+  const user = await findActiveUser(userId)
+  if (!user) {
+    logger.error('user not found', {
+      userId,
+    })
+    return
+  }
+
+  logger.info('exporting all items...', {
+    userId,
+  })
+
+  const token = jwt.sign(
+    {
+      uid: userId,
+    },
+    env.server.jwtSecret,
+    { expiresIn: '1d' }
+  )
+
+  await axios.post(env.queue.exportTaskHandlerUrl, undefined, {
+    headers: {
+      OmnivoreAuthorizationHeader: token,
+    },
+  })
+}
--- a/packages/api/src/routers/export_router.ts
+++ b/packages/api/src/routers/export_router.ts
@@ -1,0 +1,58 @@
+import cors from 'cors'
+import express, { Router } from 'express'
+import { getClaimsByToken, getTokenByRequest } from '../utils/auth'
+import { corsConfig } from '../utils/corsConfig'
+import { queueExportJob } from '../utils/createTask'
+import { logger } from '../utils/logger'
+
+export function exportRouter() {
+  const router = Router()
+
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  router.get('/', cors<express.Request>(corsConfig), async (req, res) => {
+    const token = getTokenByRequest(req)
+    // get claims from token
+    const claims = await getClaimsByToken(token)
+    if (!claims) {
+      logger.error('Token not found')
+      return res.status(401).send({
+        error: 'UNAUTHORIZED',
+      })
+    }
+
+    // get user by uid from claims
+    const userId = claims.uid
+
+    try {
+      const job = await queueExportJob(userId)
+
+      if (!job) {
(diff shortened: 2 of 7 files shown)


packages/export-handler/src/index.ts
```diff
@@ -81,103 +100,132 @@ export const exporter = Sentry.GCPFunction.wrapHttpFunction(
...
+          items.forEach((item) => {
+            const slug = item.slug
+            const content = item.content
+            const highlights = item.highlights
+            if (content) {
+              // Add content files to /content
+              archive.append(content, {
+                name: `content/${slug}.html`,
+              })
+            }
```
sywhb, line 202: @jacksonh I wonder if I should call the `/content` API instead to get the readable content so we can avoid long queries in db

packages/api/src/jobs/export.ts
```diff
@@ -0,0 +1,40 @@
...
+import { findActiveUser } from '../services/user'
+import { logger } from '../utils/logger'
+
+export interface ExportJobData {
+  userId: string
+}
+
+export const EXPORT_JOB_NAME = 'export'
+
+export const exportJob = async (jobData: ExportJobData) => {
```
sywhb, line 13: @jacksonh another question is if we should just use async jobs to export the data instead of calling the cloud run service

packages/export-handler/src/index.ts
```diff
@@ -81,103 +100,132 @@ export const exporter = Sentry.GCPFunction.wrapHttpFunction(
...
+            readingProgress: item.readingProgressPercent,
+            thumbnail: item.image,
+            labels: item.labels?.map((label) => label.name),
+            savedAt: item.savedAt,
+            updatedAt: item.updatedAt,
+            publishedAt: item.publishedAt,
+          }))
+
+          archive.append(JSON.stringify(metadata, null, 2), {
+            name: `metadata_${cursor}_to_${cursor + size}.json`,
```
sywhb, line 189: @jacksonh Here I actually export paginated `metadata` into multiple JSON files to avoid having a very large JSON file

packages/api/src/routers/export_router.ts
```diff
@@ -0,0 +1,58 @@
...
+import { getClaimsByToken, getTokenByRequest } from '../utils/auth'
+import { corsConfig } from '../utils/corsConfig'
+import { queueExportJob } from '../utils/createTask'
+import { logger } from '../utils/logger'
+
+export function exportRouter() {
+  const router = Router()
+
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
+  router.get('/', cors<express.Request>(corsConfig), async (req, res) => {
```
sywhb, line 12: This GET API is to enqueue the export job which calls the exporter endpoint

packages/api/src/utils/createTask.ts
```diff
@@ -1073,4 +1073,23 @@ export const enqueueExpireFoldersJob = async () => {
...
+    return undefined
+  }
+
+  return queue.add(
+    EXPORT_JOB_NAME,
+    { userId },
+    {
+      jobId: `${EXPORT_JOB_NAME}_${userId}_${JOB_VERSION}`,
+      removeOnComplete: true,
+      removeOnFail: true,
```
sywhb, line 1088: The export jobs are deduplicated but we probably also want to limit number of exports per user per day by checking the number of zip files in cloud storage

sywhb (commented):
@jacksonh Could you help me to clarify some questions in the PR? Thank you!

packages/api/src/jobs/export.ts
jacksonh: @sywhb maybe we should try to start with an async job and get a baseline of performance in demo. Both our accounts are quite large there so it would give an idea.

packages/api/src/utils/createTask.ts
jacksonh: Yeah agreed, or even create an entry in postgres, as annoying as that can be

packages/export-handler/src/index.ts
sywhb: Yeah, it would

packages/api/src/utils/createTask.ts
sywhb: Yeah, makes sense to me

sywhb:
Could you take a look at this when you have time? 
No rush 🙏 Thank you @jacksonh

jacksonh (approved):
I think this looks good, we definitely need to test with some larger libraries in demo. I'm a little worried it could time out and then just get stuck in a loop restarting from the beginning.

sywhb:
> I think this looks good, we definitely need to test with some larger libraries in demo. I'm a little worried it could time out and then just get stuck in a loop restarting from the beginning.

Yeah, I agree. Probably should create a batch of sub tasks so could better recover and rate limit the number of export tasks

PR #4337, opened 2024-08-28
remove tts cache

sywhb:
This is required to be deployed before we shutdown the tts cache server and we also have to update the `REDIS_URL` and `REDIS_CERT` env var in tts cloud run


jacksonh:
With this does GCS still get checked?

sywhb:
> With this does GCS still get checked?

Yes, we will still upload audio data & speech marks to GCS and download them when requested

jacksonh (approved):

PR #4350, opened 2024-08-29
feature/archive account

sywhb:
- **allow omnivore_admin to delete filters**
- **add archived status to user account**
- **redirect archived user to /export page when login**


packages/api/src/jobs/rss/refreshAllFeeds.ts
```diff
@@ -40,12 +40,11 @@ export const refreshAllFeeds = async (db: DataSource): Promise<boolean> => {
       FROM
         omnivore.subscriptions s
       INNER JOIN
-        omnivore.user u ON u.id = s.user_id
+        omnivore.user u ON u.id = s.user_id AND u.status = $4
```
sywhb, line 43: Optimized the query a little bit here

packages/api/src/services/library_item.ts
```diff
@@ -1039,7 +1039,10 @@ export const updateLibraryItemReadingProgress = async (
   }
 
   const updatedItem = result[0][0]
-  await pubsub.entityUpdated<ItemEvent>(EntityType.ITEM, updatedItem, userId)
+  if (updatedItem.readingProgressBottomPercent === 100) {
```
sywhb, line 1042: Here I try to reduce the number of PAGE_UPDATE event being sent so we only send the event when the item is fully read or reset to unread

packages/api/src/routers/auth/google_auth.ts
```diff
@@ -158,15 +157,16 @@ export async function handleGoogleWebAuth(
       }
     }
 
+    let redirectURL = `${baseURL()}${
+      user.status === StatusType.Archived ? '/export' : DEFAULT_HOME_PATH
```
jacksonh, line 161: Can we change this to `/account-archived`.

jacksonh:
Just one small change on the path

jacksonh (approved):

PR #4355, opened 2024-08-30
create a account-archived page on web for archived users to export data

sywhb:


packages/web/components/templates/ExportDataForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export function ExportDataButton(): JSX.Element {
+  const [disabledState, setDisabledState] = useState<boolean>(false)
+
+  const handleExportClick = async () => {
+    setDisabledState(true)
+
+    try {
+      // Make a GET request to the /export endpoint
+      const response = await fetch('/api/export', {
```
jacksonh, line 13: What happens if they click this multiple times?

sywhb: There are two cases:
1. If the export task is running, clicking this multiple times won't create multiple tasks and just return `200`. Probably we should return `202` instead
2. If the export tasks is failed/completed, clicking this multiple times will firstly check if the user has exceed the export quota == 3
2a. if not exceed, create a new export task
2b. if exceed, failed and return `400`

jacksonh: great! 

Is there any way in API to get the status of the export?

sywhb: Yeah, we can reuse the same export API by providing different status code like `200` or `204`. Alternatively there is a `/task/{{id}}` API which could be used to query the status of a task

sywhb: I think probably I will just reuse the `/export` API and use the status code and state in the response

PR #4378, opened 2024-09-16
fix: Quoted redis url in compose causing error

emilknievel:
When following the readme to run the app through docker compose the following error gets spammed in the logs:

```
omnivore-api            | [ioredis] Unhandled error event: Error: connect ENOENT %27redis://redis:6379%27
omnivore-api            |     at PipeConnectWrap.afterConnect [as oncomplete] (node:net:1494:16)
omnivore-api            |     at PipeConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```

This is most likely caused by the quotes being interpreted.


Podginator:
Fixed as part of the Self-hosting PR that will be merged soon.

PR #4494, opened 2024-11-08
Self hosting browser extension updates

Dadams2:
Overrides the request URL for users who choose to set their own API URL, otherwise leaves it as is. 

Changing the `content-security-policy` isn't ideal but I don't think we have another option. 

Having a look at the rest of the extension it might be possible to get it all working with manifest v3. But this should work now.

I'm not sure how much sanitation of user inputted urls should be done.

### DIFF
--- a/pkg/extension/src/scripts/api.js
+++ b/pkg/extension/src/scripts/api.js
@@ -1,8 +1,9 @@
 function gqlRequest(apiUrl, query) {
-  return getStorageItem('apiKey')
-    .then((apiKey) => {
+  return Promise.all([getStorageItem('apiKey'), getStorageItem('apiUrl')])
+    .then(([apiKey, savedUrl]) => {
       const auth = apiKey ? { Authorization: apiKey } : {}
-      return fetch(apiUrl, {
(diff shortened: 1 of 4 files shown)


oswinma (approved):

pkg/extension/src/scripts/api.js
```diff
@@ -1,8 +1,9 @@
 function gqlRequest(apiUrl, query) {
-  return getStorageItem('apiKey')
-    .then((apiKey) => {
+  return Promise.all([getStorageItem('apiKey'), getStorageItem('apiUrl')])
+    .then(([apiKey, savedUrl]) => {
       const auth = apiKey ? { Authorization: apiKey } : {}
-      return fetch(apiUrl, {
```
submatrix, line 5: I'm not sure it's the best idea to build the url this deep in the call stack. Minimally, I'm seeing that this is causing error messages generated in `background.js` to print out the wrong url since `omnivoreGraphqlURL` is hardcoded:

`const omnivoreGraphqlURL = "https://api-prod.omnivore.app/api/"`

which leads to misleading error messages when debugging the extension.

PR #4541, opened 2025-02-15
Fix: incorrect group name when reading custom environment

pgrosslicht:
This fixes the crash experienced after setting a custom environment.

This regression seems to have been introduced 8 months ago already in https://github.com/omnivore-app/omnivore/pull/4110.


Podginator:
I’ll merge this in - I’m hoping we can get Jackson to submit this version to the App Store. @jacksonh

PR #4576, opened 2025-05-03
Fixing issue with mail content from SNS that is base64 encoded

travisemslander:
I was not seeing newsletters and other test emails come through successfully to omnivore. After a bunch of debug logging calls I finally figured out that I was getting base64'ed content and it failed to be parsed by the simpleParser. By detecting when the content is base64 based on a property in the SNS notification and then base64 decoding I was able to get these messages to be correctly processed and then consumed into my omnivore library.

I figured others may appreciate this fix as well.

### DIFF
--- a/packages/local-mail-watcher/src/index.ts
+++ b/packages/local-mail-watcher/src/index.ts
@@ -7,6 +7,7 @@ import { SnsMessage } from './types/SNS'
 import { simpleParser } from 'mailparser'
 import axios from 'axios'
 import { convertToMailObject } from './lib/emailApi'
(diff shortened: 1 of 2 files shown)


Podginator:
Thanks for this! Appreciated indeed 👍🏻

lukebtaylor:
Hello! First time contributing here, so please let me know if I need to do something different. I saw the self-build was failing due to a PR related to this issue. I just submitted a fresh PR as a fix. A review is greatly appreciated. Thank you!
- https://github.com/omnivore-app/omnivore/pull/4581

PR #4621, opened 2025-10-17
feat: Add Podman Integration 

Poudlardo:
As an alternative for docker, this PR integrates a podman-compose.yml (slightly different from docker one, in the network settings and images links) and its own .env.example file.


Podginator:
Looks good to me. Would be nice to include some docs, but for now this looks fine. Feel free to open a new pr with docs, otherwise I will take care of it 👍🏻

PR #4642, opened 2026-03-24
Remove port exposing for Redis service

koluch:
There is no need to expose redis to public network, this could cause security issues


Podginator:
Thanks for this. Agreed. Originally that was open for me to debug an issue, but that is now no longer needed

PR #4644, opened 2026-04-11
adding android compatibility to manifest for firefox extension

tlonic:
The current Firefox extension does not show up in the extensions for Android, despite the fact that it does work (tested on Nightly). This is due to the extension not being specified as compatible.[1](https://discourse.mozilla.org/t/our-android-compatible-extension-does-not-show-up-in-search-on-ff-for-android-why/126069/3)[2](https://extensionworkshop.com/documentation/publish/version-compatibility/#browser-specific-settings)


Podginator:
Merged with new PR

PR #4678, opened 2026-08-28
fix: replace permanent auth token in SSO redirect URL with one-time exchange

### LINKED ISSUE #4676: Non-expiring auth credential exposed in SSO redirect URL (?tok=) enables account takeover
## Summary

In the default (`prod` appEnv / Vercel) Google sign-in flow, the API puts the **permanent** web auth JWT (`authToken`, the `auth` cookie value) inside an SSO JWT that is carried in the redirect URL as `/api/client/auth?tok=<ssoToken>`. JWT payloads are **signed, not encrypted**, so anyone who observes that URL (access logs, CDN/proxy/Vercel logs, browser history, HAR export, shared link) can decode the `tok` parameter and recover `authToken` — a credential that **never expires** and is accepted by the API with no session binding or revocation. This enables persistent account takeover.

## Impact and preconditions

- **Impact**: full, **permanent** account takeover (read articles/notes, manage subscriptions, run authenticated GraphQL operations). The credential cannot be invalidated by password change or logout — only by rotating `JWT_SECRET`.
- **Preconditions**:
  1. Deployment uses the default web app config (`NEXT_PUBLIC_APP_ENV` unset or `prod`) so Google sign-in posts to `/api/auth/vercel/gauth-redirect` (`isVercel=true`). This is the official deployment.
  2. `SSO_JWT_SECRET` is configured on API and web (standard key in `.env.template` / self-hosting `.env.example`).
  3. The victim logs in with Google.
  4. The attacker obtains the `tok` URL or a log/history record containing it.
- **Severity**: Medium (conditional on URL observation; impact is permanent full takeover).

## Root cause

`packages/api/src/routers/auth/jwt_helpers.ts` — `createWebAuthToken` signs `{ uid }` with `JWT_SECRET` and **no `expiresIn`** (no `exp` claim):

```ts
export async function createWebAuthToken(userId: string): Promise<string | undefined> {
  const authToken = await signToken({ uid: userId }, env.server.jwtSecret) // no exp -> never expires
  return authToken as string
}
```

`packages/api/src/utils/sso.ts` — that permanent `authToken` is embedded in an SSO JWT and placed in the URL query string:

```ts
export const createSsoToken = (authToken: string, redirectTo: string): string => {
  const ssoToken = jwt.sign({ authToken, redirectTo }, env.server.ssoJwtSecret, { expiresIn: '1d' })
  return ssoToken
}

export const ssoRedirectURL = (ssoToken: string): string => {
  const u = new URL(homePageURL())
  u.pathname = 'api/client/auth'
  u.searchParams.append('tok', ssoToken)   // credential enters URL query param
  return u.toString()
}
```

The `expiresIn: '1d'` only bounds the SSO wrapper; the embedded `authToken` has no expiry.

`packages/api/src/routers/auth/google_auth.ts` — triggered in the `isVercel` branch:

```ts
if (isVercel) {
  const ssoToken = createSsoToken(authToken, redirectURL)
  redirectURL = ssoRedirectURL(ssoToken)
}
```

`packages/web/pages/api/client/auth.ts` — the web endpoint verifies the SSO token and writes the embedded credential as the `auth` cookie:

```ts
const tok = req.query.tok
if (ssoJwtSecret && tok && !Array.isArray(tok)) {
  const payload = jwt.verify(tok, ssoJwtSecret) as AuthPayload
  res.setHeader('Set-Cookie', serialize('auth', payload.authToken, cookieOptions))
  res.writeHead(302, { Location: payload.redirectTo })
}
```

`packages/api/src/utils/auth.ts` — the API accepts that credential unconditionally:

```ts
return jwt.verify(token, env.server.jwtSecret) as Claims   // no session binding, no revocation
```

(used from `packages/api/src/apollo.ts`: `req?.cookies?.auth || req?.headers?.authorization`).

## Why the exposure is recoverable

A JWT signature provides **integrity, not confidentiality**. The `tok` parameter's payload is base64url-encoded JSON, so decoding `tok` yields the full `authToken`, and decoding `authToken` yields `{ "uid": ... }`. No key is needed to read it. Verified with an HS256 equivalent of the exact `jwt.sign`/`jwt.verify` calls above.

## Minimal reproduction (local, no attacker needed to demonstrate the mechanism)

1. Configure a local instance per the repo README with `JWT_SECRET`, `SSO_JWT_SECRET`, `GAUTH_CLIENT_ID`, `GAUTH_SECRET`, `CLIENT_URL`, and `NEXT_PUBLIC_*` vars.
2. Complete a Google sign-in.
3. Capture the API `302 Location`: `https://<CLIENT_URL>/api/client/auth?tok=<ssoToken>`.
4. base64url-decode the middle segment of `tok` → `{"authToken":"<JWT>","redirectTo":"...","exp":...}`.
5. base64url-decode the middle segment of `authToken` → `{"uid":"<userId>"}` (no `exp`).
6. Replay against `POST /api/graphql` with `Cookie: auth=<authToken>` (or `Authorization: <authToken>`) → fully authenticated as the victim.

Redacted example of what `tok` reveals (placeholders, not real tokens):
```
tok payload:     {"authToken":"eyJhbGciOiJIUzI1NiIs...","redirectTo":"https://omnivore.app/home","exp":17…}
authToken payload: {"uid":"9a…"}            // no exp
```

## Negative controls (boundaries)

- If `ssoJwtSecret` is unset, or `tok` is missing/an array, `/api/client/auth` redirects to the default home path and does **not** set the cookie.
- A tampered or invalid `tok` fails `jwt.verify` and sets no cookie (not fail-open).
- The non-Vercel `POST /gauth-redirect` path (`isVercel=false`) does not use `tok`.
- This issue does **not** rely on forging tokens — that would require `JWT_SECRET`/`SSO_JWT_SECRET`.

## Suggested fix

1. Do not pass persistent credentials in URLs: use a one-time, short-lived server-side exchange token (e.g., stored in Redis, single-use), or have `/api/client/auth` issue/refresh the cookie directly.
2. Add a short `expiresIn` to `createWebAuthToken` with rolling refresh so a leaked credential is not valid forever.
3. Set `Referrer-Policy: no-referrer` on `/api/client/auth` and sanitize query strings in access logs.

## Coordination

Reported per `SECURITY.md` (Issues tab). Full call chain, raw HTTP reconstruction, and verification notes are available on request. Happy to provide any additional evidence or a patch draft.

ayam04:
Closes #4676

## The vulnerability

In the default prod (Vercel) Google sign-in flow, the API embedded the permanent web auth JWT (the `auth` cookie value) inside the SSO JWT that gets carried in the redirect URL as `/api/client/auth?tok=<ssoToken>`. JWT payloads are signed, not encrypted, so anyone who observed the URL (access logs, CDN/Vercel logs, browser history, HAR exports, shared links) could decode `tok` and recover a credential that never expires and has no session binding or revocation. That is persistent account takeover.

## The fix

The `tok` parameter now carries only a random, one-time exchange id:

- API (`packages/api/src/utils/sso.ts`): `createSsoToken` stores the real auth token in redis under `sso:exchange:<uuid>` with a 5 minute TTL and never puts it in the JWT. `exchangeSsoToken` redeems it with GETDEL, so replaying the same tok from a log or a shared link fails on the second attempt even inside the expiry window.
- API (`packages/api/src/routers/auth/auth_router.ts`): new `POST /api/auth/exchange` endpoint validates the sso token and returns the stored auth token; 400 on missing, expired, or already-redeemed tokens.
- Web (`packages/web/pages/api/client/auth.ts`): verifies the sso JWT, exchanges it server-side against the API, then sets the `auth` cookie. The permanent credential never appears in a URL. Login/confirm-email redirect flows are unchanged, so existing sessions and mobile flows are untouched.

The API-domain cookie path (`setAuthInCookie`) is unchanged.

## Tests

Three new cases in `packages/api/test/routers/auth.test.ts`:
1. a valid sso token exchanges for an auth token
2. the same token refuses a second exchange (single use)
3. a garbage token is refused

## Verification

- `tsc --noEmit` clean in both packages (typescript via root node_modules)
- The DB-backed suite needs a running Postgres; none is available in this environment, so the auth test suite itself was not executed locally. CI runs it.

## Notes

redisDataSource is initialized in server startup (server.ts) before routes are served, so redisClient is always available. ioredis 6 maps getdel to redis GETDEL (redis >= 6.2).

### DIFF
--- a/packages/api/src/routers/auth/auth_router.ts
+++ b/packages/api/src/routers/auth/auth_router.ts
@@ -39,7 +39,7 @@ import { logger } from '../../utils/logger'
 import { ARCHIVE_ACCOUNT_PATH, DEFAULT_HOME_PATH } from '../../utils/navigation'
 import { hourlyLimiter } from '../../utils/rate_limit'
 import { verifyChallengeRecaptcha } from '../../utils/recaptcha'
-import { createSsoToken, ssoRedirectURL } from '../../utils/sso'
+import { createSsoToken, exchangeSsoToken, ssoRedirectURL } from '../../utils/sso'
 import { handleAppleWebAuth } from './apple_auth'
 import type { AuthProvider } from './auth_types'
 import {
@@ -409,6 +409,21 @@ export function authRouter() {
       return res.redirect(`${env.client.url}/login?errorCodes=AUTH_FAILED`)
     }
   }
+
+  router.post(
+    '/exchange',
+    async (req: express.Request, res: express.Response) => {
+      const token = (req.body?.ssoToken ?? '') as string
+      if (!token) {
+        return res.status(400).json({ errorCode: 'SSO_EXCHANGE_NOT_FOUND' })
+      }
+      const authToken = await exchangeSsoToken(token)
+      if (!authToken) {
+        return res.status(400).json({ errorCode: 'SSO_EXCHANGE_NOT_FOUND' })
+      }
+      return res.status(200).json({ authToken })
+    }
+  )
 
   router.options(
     '/email-login',
--- a/packages/api/src/utils/sso.ts
+++ b/packages/api/src/utils/sso.ts
@@ -1,18 +1,57 @@
 import { env, homePageURL } from '../env'
+import { redisDataSource } from '../redis_data_source'
 import * as jwt from 'jsonwebtoken'
+import { v4 as uuidv4 } from 'uuid'
+
+const SSO_EXPIRY_SECONDS = 300
+const SSO_EXCHANGE_KEY_PREFIX = 'sso:exchange:'
 
 export const createSsoToken = (
   authToken: string,
   redirectTo: string
 ): string => {
+  const exchangeId = uuidv4()
+  // The permanent auth token never leaves the server: it is stored under a
+  // random one-time exchange id instead of being embedded in the redirect
+  // URL. JWT payloads are signed, not encrypted, so any credential carried in
+  // the tok parameter is recoverable by anyone who observes the URL (access
+  // logs, browser history, HAR exports). The URL now carries only an
+  // exchange id that is single-use and short-lived.
+  void redisDataSource.redisClient?.setex(
+    `${SSO_EXCHANGE_KEY_PREFIX}${exchangeId}`,
+    SSO_EXPIRY_SECONDS,
+    authToken
+  )
+
   const ssoToken = jwt.sign(
-    { authToken, redirectTo },
+    { exchangeId, redirectTo },
     env.server.ssoJwtSecret,
     {
-      expiresIn: '1d',
+      expiresIn: `${SSO_EXPIRY_SECONDS}s`,
     }
   )
   return ssoToken
+}
+
+export const exchangeSsoToken = async (
+  ssoToken: string
+): Promise<string | undefined> => {
+  let payload: { exchangeId: string; redirectTo: string } | undefined
+  try {
+    payload = jwt.verify(ssoToken, env.server.ssoJwtSecret) as typeof payload
+  } catch {
+    return undefined
+  }
+  if (!payload?.exchangeId) return undefined
+
+  // getdel gives single-use semantics: replaying the same tok from a log or
+  // shared link fails on the second attempt even inside the 5 minute window.
+  const client = redisDataSource.redisClient
+  if (!client) return undefined
+  const authToken = await client.getdel(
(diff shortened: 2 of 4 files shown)


ayam04:
Hi, gentle bump on this one. The complete fix is in: the SSO redirect no longer carries the permanent auth token; it carries a short-lived single-use exchange id redeemed server-side (redis GETDEL, 400 on replay), with three regression tests covering mint, replay refusal, and garbage tokens. Both packages typecheck clean, and CI covers the DB-backed tests. Happy to adjust anything you want changed.

